PR #11, opened 2023-08-15
Copy and normalize query data when cosine for thread safe

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/26202

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -395,11 +395,7 @@ IvfIndexNode<T>::Search(const DataSet& dataset, const Config& cfg, const BitsetV
     auto data = dataset.GetTensor();
 
     const IvfConfig& ivf_cfg = static_cast<const IvfConfig&>(cfg);
-
-    // do normalize for COSINE metric type
-    if (IsMetricType(ivf_cfg.metric_type.value(), knowhere::metric::COSINE)) {
-        Normalize(dataset);
-    }
+    bool is_cosine = IsMetricType(ivf_cfg.metric_type.value(), knowhere::metric::COSINE);
 
     auto k = ivf_cfg.k.value();
     auto nprobe = ivf_cfg.nprobe.value();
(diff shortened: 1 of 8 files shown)


src/index/ivf/ivf.cc
```diff
@@ -498,14 +509,32 @@ IvfIndexNode<T>::RangeSearch(const DataSet& dataset, const Config& cfg, const Bi
                     auto cur_data = (const uint8_t*)xq + index * dim / 8;
                     index_->range_search_thread_safe(1, cur_data, radius, &res, index_->nlist, bitset);
                 } else if constexpr (std::is_same<T, faiss::IndexIVFFlat>::value) {
-                    auto cur_data = (const float*)xq + index * dim;
-                    index_->range_search_without_codes_thread_safe(1, cur_data, radius, &res, index_->nlist, 0, bitset);
+                    auto cur_query = (const float*)xq + index * dim;
```
liliu-z, line 512: Why we don't do 
```
auto cur_query = (const float*)xq + index * dim;
if (is_cosine) {
  cur_query = CopyAndNormalizeFloatVec(cur_query, dim);
}
```
And then we don't have to do any changes in the if else blocks

thirdparty/hnswlib/hnswlib/hnswalg.h
```diff
@@ -1154,14 +1154,18 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
             return {};
 
+        size_t dim = *(size_t*)dist_func_param_;
+
         // do normalize for COSINE metric type
+        std::unique_ptr<float[]> query_data_norm;
         if (metric_type_ == Metric::COSINE) {
-            knowhere::NormalizeVec((float*)query_data, *((size_t*)dist_func_param_));
+            query_data_norm = knowhere::CopyAndNormalizeFloatVec((const float*)query_data, dim);
+            query_data = query_data_norm.get();
```
chasingegg, line 1168: will this operation still changes the original query_data?

src/index/ivf/ivf.cc
```diff
@@ -423,17 +419,36 @@ IvfIndexNode<T>::Search(const DataSet& dataset, const Config& cfg, const BitsetV
...
+                    } else {
+                        index_->search_thread_safe(1, cur_query, k, distances + offset, ids + offset, nprobe,
+                                                   scann_cfg.reorder_k.value(), bitset);
+                    }
                 } else {
-                    auto cur_data = (const float*)data + index * dim;
-                    index_->search_thread_safe(1, cur_data, k, distances + offset, ids + offset, nprobe, 0, bitset);
+                    auto cur_query = (const float*)data + index * dim;
+                    if (is_cosine) {
+                        auto cur_query_norm = CopyAndNormalizeFloatVec(cur_query, dim);
```
chasingegg, line 445: better to rename all this `cur_query_norm` to `cur_normalized_query`, it is confusing to use query_norm which could be considered as a numeric number means norm.

src/index/ivf/ivf.cc
cydrain: updated

PR #24, opened 2023-08-22
change default stream size

Presburger:
issue: #29 

The execution threads in the gpu thread pool share the stream in the stream pool, and if the user does not set the KNOWHERE_STREAMS_PER_GPU environment variable, the stream pool with a default size of 1 is too small.

### DIFF
--- a/src/common/raft/raft_utils.h
+++ b/src/common/raft/raft_utils.h
@@ -53,13 +53,13 @@ inline auto raft_mutex = std::mutex{};
 struct gpu_resources {
     gpu_resources(std::optional<std::size_t> streams_per_device = std::nullopt)
         : streams_per_device_{streams_per_device.value_or([]() {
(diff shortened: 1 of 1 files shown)


wphicks (approved):
This looks like an excellent change. The default should certainly have been higher than 1. In the near future, we might take advantage of RAFT's new device_resources_manager to simplify device resource generation generally, but in the meantime, this seems like a very worthwhile change.

Presburger:
> This looks like an excellent change. The default should certainly have been higher than 1. In the near future, we might take advantage of RAFT's new device_resources_manager to simplify device resource generation generally, but in the meantime, this seems like a very worthwhile change.

Yes, setting the default value to 1 prevents parallelism for h2d (host-to-device) and exec (execution) of kernels, leading to a significant performance drop on the A100.

PR #26, opened 2023-08-22
Temporary solution to fix ivf_flat search result wrong when cosine

cydrain:


liliu-z:
issue: #27 

We don't have metric type in Deserialization, so this PR do normalization lazily to ensure the recall of IVFFlat. This solution will be deprecated in the next version

PR #36, opened 2023-08-29
Temporary solution to fix ivf_flat search result wrong when cosine

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/26675


liliu-z (approved):
/lgtm

Issue #44, opened 2023-09-01
[bug]: ScaNN add_with_ids bug will cause cosine metric result wrong


chasingegg:
In IndexIVFPQFastScan.cpp, add_with_ids will compute norms when it is cosine metric.
```
void IndexIVFPQFastScan::add_with_ids(
        idx_t n,
        const float* x,
        const idx_t* xids) {
    if (is_cosine_) {
        this->norms = ....
        add_with_ids_impl(n, norm_data, xids);
    } else {
        add_with_ids_impl(n, x, xids);
    }
}
```
And in `add_with_ids_impl` function, it should call itself in function body not `add_with_ids`. Current behaviour will cause norm is wrong, which leads to wrong result in cosine metric.
```
void IndexIVFPQFastScan::add_with_ids_impl(
        idx_t n,
        const float* x,
        const idx_t* xids) {
  ....
  // do some blocking to avoid excessive allocs
  idx_t bs = 65536;
  if (n > bs) {
      for (idx_t i0 = 0; i0 < n; i0 += bs) {
          idx_t i1 = std::min(n, i0 + bs);
          add_with_ids(i1 - i0, x + i0 * d, xids ? xids + i0 : nullptr);
      }
      return;
  }
  ....
}
```

PR #45, opened 2023-09-01
Fix scann add_with_ids logic

chasingegg:
related #44 
Test on glove-100 dataset, under the configuration of nlist=1024, nprobe=50, reorder_k=200, topk=100.
before this change, we get recall=0.3071
after this change, we get recall=0.9351


Presburger (approved):

hhy3 (approved):
/lgtm

Issue #61, opened 2023-09-07
[feat] Add ann iterator support to HNSW index


zhengbuqian:
this is the first index to add iterator support.

PR #62, opened 2023-09-07
minor refactor: extract codes to search for top layers into a method

zhengbuqian:
this is some pre-work for the iterator work #61 

the code to search in top layers is duplicated for searchKnn and searchRange, and will be duplicated again for iterator. thus use a method to wrap the common logic

/kind improvement


liliu-z:
Good catch, plz resolve the conflict before check in

Issue #70, opened 2023-09-11
[bug] growing index search result is failed on valid offset checking 


foxspy:
[2023-09-06 00:00:42,768 - ERROR - fouram]: RPC error: [search], <MilvusException: (code=1, message=attempt #0: fail to Search, QueryNode ID=1, reason=worker(1) query failed: UnexpectedError: Assert "0 <= offsets[index] && offsets[index] < segment->get_row_count()" at /go/src/github.com/milvus-io/milvus/internal/core/src/segcore/Reduce.cpp:97
 => invalid offset 179938, segment 444057371270847350 with rows num 179937, data or index corruption: channel=by-dev-rootcoord-dml_1_444057371270447222v1: fail to search/query shard leader: attempt #1: no available shard delegator found: service unavailable: fail to search on all shard leaders)>, <Time:{'RPC start': '2023-09-06 00:00:42.156459', 'RPC error': '2023-09-06 00:00:42.768611'}> (decorators.py:108)
[2023-09-06 00:00:42,772 - ERROR - fouram]: (api_response) : <MilvusException: (code=1, message=attempt #0: fail to Search, QueryNode ID=1, reason=worker(1) query failed: UnexpectedError: Assert "0 <= offsets[index] && offsets[index] < segment->get_row_count()" at /go/src/github.com/milvus-io/milvus/internal/core/src/segcore/Reduce.cpp:97
 => invalid offset 179938, segment 444057371270847350 with rows num 179937, data or index corruption: channel=by-dev-rootcoord-dml_1_444057371270447222v1: fail to search/query shard leader: attempt #1: no available shard delegator found: service unavailable: fail to search on all shard leaders)> (api_request.py:53)
[2023-09-06 00:00:42,772 - ERROR - fouram]: [CheckFunc] search request check failed, response:<MilvusException: (code=1, message=attempt #0: fail to Search, QueryNode ID=1, reason=worker(1) query failed: UnexpectedError: Assert "0 <= offsets[index] && offsets[index] < segment->get_row_count()" at /go/src/github.com/milvus-io/milvus/internal/core/src/segcore/Reduce.cpp:97

1. growing index updates `active_count` after all data has been inserted into index
2. growing index searching will be applied on the index with actual `count` bigger than  `active_count` 
3. when id >= `active_count` has been selected, it will cause the id validation checking failed.

PR #71, opened 2023-09-11
fix search on out-of-range id

foxspy:
#70 
fix the bug by ignoring all ids beyond the bitset size limit.

### DIFF
--- a/include/knowhere/bitsetview.h
+++ b/include/knowhere/bitsetview.h
@@ -50,7 +50,8 @@ class BitsetView {
 
     bool
     test(int64_t index) const {
-        return bits_[index >> 3] & (0x1 << (index & 0x7));
+        // when index is larger than the max_offset, ignore it
+        return (index >= static_cast<int64_t>(num_bits_)) || (bits_[index >> 3] & (0x1 << (index & 0x7)));
     }
 
     size_t


include/knowhere/bitsetview.h
```diff
@@ -50,7 +50,8 @@ class BitsetView {
 
     bool
     test(int64_t index) const {
-        return bits_[index >> 3] & (0x1 << (index & 0x7));
+        // when index is larger than the max_offset, ignore it
+        return (bits_[index >> 3] & (0x1 << (index & 0x7))) || (index >= static_cast<int64_t>(num_bits_));
```
liliu-z, line 54: Can I have the code point of the validation?

Some questions:
1. If index is out of range, `bits_[index >> 3]` will overflow and the result before `||` is undefined, making the range check get ignore. Is this as expected?
2. Will the `active_count` get updated in the chunk granularity?

foxspy: 1. Good catch !  I set the range check at the right side because it is just a corner case, and once the left checking success, then the right side checking can be ignored, which can reduce the influence of cost at most . But it seems that the range checking is better placed at the left side to avoid overflow.
2. `active_count` is just the count of visible vectors in the segments, which can be determined just by the [`Timestamp`.](https://github.com/milvus-io/milvus/blob/45d9fb5929e400f42f7d35d51046189089b2ad85/internal/core/src/query/visitors/ExecPlanNodeVisitor.cpp#L91) . This value is a determined by the search context. For other cases such as sealed index, it is a const property and never be changed.

Issue #30, opened 2023-08-23
Remove IVF_FLAT_NM index type


cydrain:
IVF_FLAT_NM index type is introduced since Milvus 0.7, to save the index disk size.

But in Milvus 2.x it increases the complexity for following feature's development:
1. IVF_FLAT + COSINE support GetVectorByIds
2. zero copy

it's time to remove IVF_FLAT_NM index type

PR #76, opened 2023-09-12
New solution to handle IVF_FLAT backward compatibility

cydrain:
Issue: #30

### DIFF
--- a/src/common/utils.cc
+++ b/src/common/utils.cc
@@ -83,28 +83,39 @@ ConvertIVFFlatIfNeeded(const BinarySet& binset, const uint8_t* raw_data, const s
 
     MemoryIOReader reader(binary->data.get(), binary->size);
 
-    // there are 2 possibilities for the input index binary:
-    //  1. native IVF_FLAT, do nothing
-    //  2. IVF_FLAT_NM, convert to native IVF_FLAT
     try {
-        // try to parse as native format, if it's actually _NM format,
-        // faiss will raise a "read error" exception for IVF_FLAT_NM format
-        faiss::read_index(&reader);
-    } catch (faiss::FaissException& e) {
-        reader.reset();
+        uint32_t h;
+        reader.read(&h, sizeof(h), 1);
 
-        // convert IVF_FLAT_NM to native IVF_FLAT
-        auto* index = static_cast<faiss::IndexIVFFlat*>(faiss::read_index_nm(&reader));
-        index->restore_codes(raw_data, raw_size);
+        // only read IVF_FLAT index header
+        std::unique_ptr<faiss::IndexIVFFlat> ivfl = std::make_unique<faiss::IndexIVFFlat>(faiss::IndexIVFFlat());
+        faiss::read_ivf_header(ivfl.get(), &reader);
+        ivfl->code_size = ivfl->d * sizeof(float);
 
-        // over-write IVF_FLAT_NM binary with native IVF_FLAT binary
-        MemoryIOWriter writer;
-        faiss::write_index(index, &writer);
-        std::shared_ptr<uint8_t[]> data(writer.data());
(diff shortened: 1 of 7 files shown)


thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -63,9 +63,17 @@ namespace faiss {
 static void read_index_header(Index* idx, IOReader* f) {
     READ1(idx->d);
     READ1(idx->ntotal);
+    READ1(idx->is_cosine);
+
+    uint8_t dummy8;
+    READ1(dummy8);
+    uint16_t dummy16;
```
foxspy, line 70: 8(used for cosine) + 8 + 16 + 32 (rest padding) = 64 ? Pls add some comments here or just use 64 bit to save the `is_cosine`

thirdparty/faiss/faiss/impl/index_write.cpp
```diff
@@ -81,9 +81,17 @@ namespace faiss {
 static void write_index_header(const Index* idx, IOWriter* f) {
     WRITE1(idx->d);
     WRITE1(idx->ntotal);
-    Index::idx_t dummy = 1 << 20;
-    WRITE1(dummy);
+    WRITE1(idx->is_cosine);
```
foxspy, line 84: same as upper

thirdparty/faiss/faiss/impl/index_write.cpp
```diff
@@ -397,6 +319,43 @@ void write_InvertedLists_with_norm(const InvertedLists* ils, IOWriter* f) {
...
+        WRITE1(h);
+        WRITE1(ils->nlist);
+        WRITE1(ils->code_size);
+        // this is a POD object
+        WRITEVECTOR(od->lists);
+
+        {
+            std::vector<OnDiskInvertedLists::Slot> v(
+                      od->slots.begin(), od->slots.end());
+            WRITEVECTOR(v);
```
foxspy, line 352: Has IVFFLAT used these inverted lists? May beyond my knowledge

src/common/utils.cc
```diff
@@ -83,28 +83,39 @@ ConvertIVFFlatIfNeeded(const BinarySet& binset, const uint8_t* raw_data, const s
...
+        ivfl->code_size = ivfl->d * sizeof(float);
+
+        auto remains = binary->size - reader.tellg() - sizeof(uint32_t) - sizeof(ivfl->invlists->nlist) -
+                       sizeof(ivfl->invlists->code_size);
+        auto invlist_size = sizeof(uint32_t) + sizeof(size_t) + ivfl->nlist * sizeof(size_t);
+        auto ids_size = ivfl->ntotal * sizeof(faiss::Index::idx_t);
+        // auto codes_size = ivfl->d * ivfl->ntotal * sizeof(float);
+
+        // IVF_FLAT_NM format, need convert to new format
+        if (remains == invlist_size + ids_size) {
```
foxspy, line 102: little hack to decide the index type by binary size. Is there any other ways?

foxspy:
Pls verify the compatibility with existing index binary. A byte difference may occur load fail. @elstic

src/common/utils.cc
cydrain: I only think out 2 ways to detect IVF_FLAT_NM:
1. use native IVF_FLAT format to parse a binary, if it throw an exception, then this binary should be IVF_FLAT_NM
2. the way as this PR, read ivf_flat index header, then calculate the binary size

before this PR, I use method 1, but it has a little risk, so I change to method 2

thirdparty/faiss/faiss/impl/index_read.cpp
cydrain: yes

Issue #73, opened 2023-09-12
[feature]: ScaNN support with_raw_data build option


chasingegg:
Currently ScaNN index contains raw data inside which can be used in refine stage of search, with more memory consumption. We can provide an option for users to decide if raw data is in index.
For users who want a more compressed index and relatively acceptable recall, can choose with_raw_data=false.
For users who do not care too much about memory usage but require a high recall, can choose with_raw_data=true.

PR #77, opened 2023-09-12
ScaNN support with_raw_data build option

chasingegg:
issue #73 
Test on SIFT1M and GLOVE1M dataset
with `with_raw_data=True`, get recall=**0.9999/0.8613** for sift and glove dataset,
with `with_raw_data=False`, get recall=**0.9692/0.5378** for sift and glove dataset.
with_raw_data=True as the default option


hhy3 (approved):
/lgtm

Issue #79, opened 2023-09-13
Support GetVectorByIds for faiss FLAT COSINE


cydrain:

cydrain:
Search/RangeSearch in #80 is as same as brute force, it's slow.
We'd better save vectors' norms into faiss FLAT index just as faiss IVFFlat index does.

But it's complicated to let knn_cosine() add a new param "y_norms", because APIs in distance.cpp are changed a lot in knowhere's faiss code.
Plan to roll back un-necessary code changes in distance.cpp, then do above enhancement.

PR #80, opened 2023-09-13
Support faiss FLAT index get vector with metric COSINE

cydrain:
Issue: #79

### DIFF
--- a/src/index/flat/flat.cc
+++ b/src/index/flat/flat.cc
@@ -36,17 +36,18 @@ class FlatIndexNode : public IndexNode {
     Train(const DataSet& dataset, const Config& cfg) override {
         const FlatConfig& f_cfg = static_cast<const FlatConfig&>(cfg);
 
-        // do normalize for COSINE metric type
-        if (IsMetricType(f_cfg.metric_type.value(), knowhere::metric::COSINE)) {
-            Normalize(dataset);
-        }
-
         auto metric = Str2FaissMetricType(f_cfg.metric_type.value());
         if (!metric.has_value()) {
             LOG_KNOWHERE_WARNING_ << "please check metric type: " << f_cfg.metric_type.value();
             return metric.error();
         }
-        index_ = std::make_unique<T>(dataset.GetDim(), metric.value());
+        if constexpr (std::is_same<faiss::IndexBinaryFlat, T>::value) {
+            index_ = std::make_unique<faiss::IndexBinaryFlat>(dataset.GetDim(), metric.value());
+        }
+        if constexpr (std::is_same<faiss::IndexFlat, T>::value) {
+            bool is_cosine = IsMetricType(f_cfg.metric_type.value(), knowhere::metric::COSINE);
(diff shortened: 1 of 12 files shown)


thirdparty/faiss/faiss/IndexFlat.cpp
```diff
@@ -36,7 +38,11 @@ void IndexFlat::search(
 
     if (metric_type == METRIC_INNER_PRODUCT) {
         float_minheap_array_t res = {size_t(n), size_t(k), labels, distances};
-        knn_inner_product(x, get_xb(), d, n, ntotal, &res, bitset);
+        if (is_cosine) {
+            knn_cosine(x, get_xb(), d, n, ntotal, &res, bitset);
```
hhy3, line 42: knn_cosine is a slow function which is a compromise for brute force, and will compute the norm each time distance is calculated. It's better to pre-compute norms of all base vectors and use them

cydrain: I know that, this is a middle-solution, I'll roll back some un-necessary faiss code change, then implement the flat with norm data solution

src/index/flat/flat.cc
```diff
@@ -236,7 +237,7 @@ class FlatIndexNode : public IndexNode {
     bool
     HasRawData(const std::string& metric_type) const override {
         if constexpr (std::is_same<T, faiss::IndexFlat>::value) {
-            return !IsMetricType(metric_type, metric::COSINE);
+            return true;
         }
         if constexpr (std::is_same<T, faiss::IndexBinaryFlat>::value) {
             return true;
```
zhengbuqian, line 243: curious: why doesn't this if have an else branch? this doesn't seem legit c++ code to me.. how come it compiles before?

zhengbuqian (commented):
/approve

src/index/flat/flat.cc
cydrain: There are only 2 possibilities to run into here, one is faiss::IndexFlat, the other is faiss::IndexBinaryFlat, so no else

hhy3 (commented):
/lgtm

thirdparty/faiss/faiss/IndexFlat.cpp
```diff
@@ -18,10 +18,28 @@
...
-        : IndexFlatCodes(sizeof(float) * d, d, metric) {}
+IndexFlat::IndexFlat(idx_t d, MetricType metric, bool is_cosine)
+        : IndexFlatCodes(sizeof(float) * d, d, metric) {
+    this->is_cosine = is_cosine;
+}
+
+void IndexFlat::add(idx_t n, const float* x) {
+    FAISS_THROW_IF_NOT(is_trained);
+    codes.resize((ntotal + n) * code_size);
+    sa_encode(n, x, &codes[ntotal * code_size]);
```
liliu-z, line 33: Why don't call `Index::add()` here and handle norm after this calling

liliu-z:
issue: 26960

PR #85, opened 2023-09-16
Add missing header

chasingegg:
https://github.com/milvus-io/milvus/actions/runs/6199723284/job/16832875500?pr=27146
/kind improvement


liliu-z:
Why we don't met any issue before?

chasingegg:
I have met one when implementing scann index, https://github.com/zilliztech/knowhere/pull/8 which is failed at milvus mac code checker rule. I guess we could align the code checker rule of milvus and knowhere.

PR #90, opened 2023-09-18
A candidate that uses Faiss 1.7.4

alexanderguzhva:
This is a beta candidate of Milvus with upgraded Faiss 1.7.4 (based on commit `c8d6f7bb2b09b8928570222131353f75acb4007b` from Sep 6 2023, but with all the most recent merged commits). 

kind/enhancement


thirdparty/faiss/faiss/IndexFlatElkan.cpp
```diff
@@ -0,0 +1,66 @@
...
+    if (distances == nullptr) {
+        dis_inner_deleter = std::make_unique<float[]>(n);
+        dis_inner = dis_inner_deleter.get();
+    }
+
+    switch (metric_type) {
+        case METRIC_INNER_PRODUCT:
+        case METRIC_L2: {
+            // ignore the metric_type, both use L2
+            elkan_L2_sse(x, get_xb(), d, n, ntotal, labels, dis_inner);
```
liliu-z, line 53: Plz CMIIW, in previous Knowhere we put this `elkan_L2_sse` in `assign`. I know `assign` will call `search` eventually. Does it mean previous Knowhere didn't use elkan during the search time? Does it matter?

liliu-z: BTW, should we add avx Elkan calculation support in the future?

src/index/ivf/ivf.cc
```diff
@@ -270,73 +272,153 @@ IvfIndexNode<T>::Train(const DataSet& dataset, const Config& cfg) {
...
+    // if cfg.use_elkan is used, then we'll use a temporary instance of
+    //  IndexFlatElkan for the training.
     try {
         if constexpr (std::is_same<faiss::IndexIVFFlat, T>::value) {
             const IvfFlatConfig& ivf_flat_cfg = static_cast<const IvfFlatConfig&>(cfg);
             auto nlist = MatchNlist(rows, ivf_flat_cfg.nlist.value());
-            qzr = new (std::nothrow) typename QuantizerT<T>::type(dim, metric.value());
-            index = std::make_unique<faiss::IndexIVFFlat>(qzr, dim, nlist, metric.value(), is_cosine);
+
+            const bool use_elkan = ivf_flat_cfg.use_elkan.value_or(true);
```
liliu-z, line 283: Just a discussion. Should we put this `use_elkan` inside Faiss or in Knowhere. 
Now we put it in Knowhere:
 - **Pros:** It makes Faiss cleaner, and easier to upgrade next time.
 - **Cons:** As we talked, if we want to make a pure Faiss support in future, we need to keep Knowhere as lightweight as possible. And anyway this PR introduced an `IndexFlatElkan` into Faiss.

src/index/ivf/ivf.cc
```diff
@@ -270,73 +272,153 @@ IvfIndexNode<T>::Train(const DataSet& dataset, const Config& cfg) {
...
             index->make_direct_map(true, faiss::DirectMap::ConcurrentArray);
         }
         if constexpr (std::is_same<faiss::IndexIVFPQ, T>::value) {
             const IvfPqConfig& ivf_pq_cfg = static_cast<const IvfPqConfig&>(cfg);
             auto nlist = MatchNlist(rows, ivf_pq_cfg.nlist.value());
             auto nbits = MatchNbits(rows, ivf_pq_cfg.nbits.value());
-            qzr = new (std::nothrow) typename QuantizerT<T>::type(dim, metric.value());
-            index = std::make_unique<faiss::IndexIVFPQ>(qzr, dim, nlist, ivf_pq_cfg.m.value(), nbits, metric.value());
+
+            const bool use_elkan = ivf_pq_cfg.use_elkan.value_or(true);
```
liliu-z, line 333: Also, so much dup codes here (I know it is like this before this PR). Any possibility to concentrate them?

src/simd/distances_avx.cc
```diff
@@ -195,5 +151,69 @@ fvec_Linf_avx(const float* x, const float* y, size_t d) {
...
+fvec_madd_avx(size_t n, const float* a, float bf, const float* b, float* c) {
+    FAISS_PRAGMA_IMPRECISE_LOOP
+    for (size_t i = 0; i < n; i++) {
+        c[i] = a[i] + bf * b[i];
+    }
+}
+FAISS_PRAGMA_IMPRECISE_FUNCTION_END
+
+// trust the compiler to unroll this properly
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
```
liliu-z, line 166: Can we make a same macro inside Knowhere can call it Knowhere_xxx_xxx?

src/simd/distances_avx.cc
```diff
@@ -38,80 +40,34 @@ masked_read(int d, const float* x) {
...
-    while (d >= 8) {
-        __m256 mx = _mm256_loadu_ps(x);
-        x += 8;
-        __m256 my = _mm256_loadu_ps(y);
-        y += 8;
-        msum1 = _mm256_add_ps(msum1, _mm256_mul_ps(mx, my));
-        d -= 8;
+    size_t i;
+    float res = 0;
+    FAISS_PRAGMA_IMPRECISE_LOOP
```
Presburger, line 49: Is the 'pragma' here reliable under the GCC compiler?

thirdparty/faiss/faiss/Clustering.cpp
cqy123456, line 1: Early stop happens in (diff = (prev_obj - stats.obj)/ prev_obj, diff < 0.01 * early_stop_threshold).
Update centroids in each iteration can make L2 loss value (obj) become smaller, but not work for IP metric. When metic = IP,  early stop in iter = 1 can easily happen in prev.obj < obj, prev.obj - obj < 0. 
So maybe diff = abs(prev_obj - stats.obj)/ prev_obj is better?

thirdparty/faiss/faiss/Clustering.cpp
```diff
@@ -573,8 +557,12 @@ void Clustering::train_encoded(
             double t0s = getmillisecs();
 
             if (!codec) {
-                index.assign(nx, reinterpret_cast<const float *>(x),
-                             assign.get(), dis.get());
+                index.search(
```
cqy123456, line 560: Early stop happens in (diff = (prev_obj - stats.obj)/ prev_obj, diff < 0.01 * early_stop_threshold).
Update centroids in each iteration can make L2 loss value (obj) become smaller, but not work for IP metric. When metic = IP, early stop in iter = 1 can easily happen in prev.obj < obj, prev.obj - obj < 0.

assign () use L2 distance to avoid this.
Maybe diff = abs(prev_obj - stats.obj)/ prev_obj is better?Early stop happens in (diff = (prev_obj - stats.obj)/ prev_obj, diff < 0.01 * early_stop_threshold).
Update centroids in each iteration can make L2 loss value (obj) become smaller, but not work for IP metric. When metic = IP, early stop in iter = 1 can easily happen in prev.obj < obj, prev.obj - obj < 0.

assign () use L2 distance to avoid this.
Maybe diff = abs(prev_obj - stats.obj)/ prev_obj is better?

src/index/ivf/ivf_config.h
```diff
@@ -31,6 +32,10 @@ class IvfConfig : public BaseConfig {
             .description("number of probes at query time.")
             .for_search()
             .set_range(1, 65536);
+        KNOWHERE_CONFIG_DECLARE_FIELD(use_elkan)
+            .set_default(true)
```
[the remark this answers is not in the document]
alexanderguzhva: No. It is better to introduce a new feature rather that introducing a new behaviour

include/knowhere/bitsetview_idselector.h
```diff
@@ -0,0 +1,33 @@
...
+#include <faiss/impl/IDSelector.h>
+
+#include "knowhere/bitsetview.h"
+
+namespace knowhere {
+
+struct BitsetViewIDSelector : faiss::IDSelector {
+    BitsetView bitset_view;
+
+    inline BitsetViewIDSelector(BitsetView bitset_view) : bitset_view{bitset_view} {
```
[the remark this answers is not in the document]
alexanderguzhva: `BitsetViewIDSelector` is a temporary object that is used during the search only. It is not stored anywhere

thirdparty/faiss/faiss/IndexFlatElkan.cpp
alexanderguzhva: baseline Knowhere does not seem to use `elkan_L2_sse()` in anywhere but the `IndexFlat::assign()` function.

src/index/ivf/ivf.cc
alexanderguzhva: The way it is introduced satisfies both imho. It makes Faiss cleaner, because it is a single Independent file/index. And it keeps Knowhere lightweight.

src/simd/distances_avx.cc
alexanderguzhva: We can, but I don't see a point of reintroducing a macro, because it is already available in Faiss.

src/simd/distances_avx.cc
alexanderguzhva: Yes. Please feel free to double-check via https://godbolt.org/z/7d9EdGx9K if needed

cmake/libs/libfaiss.cmake
```diff
@@ -16,6 +16,10 @@ knowhere_file_glob(GLOB FAISS_AVX2_SRCS
 
 list(REMOVE_ITEM FAISS_SRCS ${FAISS_AVX512_SRCS})
 
+# disable RHNSW
```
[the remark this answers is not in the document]
alexanderguzhva: I thought that it is not used in the codebase, is it?

thirdparty/faiss/faiss/impl/pq4_fast_scan_search_qbs.cpp
```diff
@@ -81,94 +109,78 @@ void kernel_accumulate_block(
...
+        ResultHandler& res,
+        const Scaler& scaler) {
     constexpr int Q1 = QBS & 15;
     constexpr int Q2 = (QBS >> 4) & 15;
     constexpr int Q3 = (QBS >> 8) & 15;
     constexpr int Q4 = (QBS >> 12) & 15;
     constexpr int SQ = Q1 + Q2 + Q3 + Q4;
 
     for (int64_t j0 = 0; j0 < ntotal2; j0 += 32) {
-        res.set_block_origin(0, j0);
```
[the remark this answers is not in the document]
alexanderguzhva: Thanks a lot. My bad, I've missed this change.

thirdparty/faiss/faiss/Clustering.cpp
alexanderguzhva: A good point. Let me see...

src/index/ivf/ivf_config.h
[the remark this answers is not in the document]
alexanderguzhva: Actually, I take it back. It seems that the recall rate improvements happen bcz of L2 distance, not because of elkan itself, hmm.

src/index/flat/flat.cc
```diff
@@ -93,18 +94,31 @@ class FlatIndexNode : public IndexNode {
...
                         auto cur_query = (const float*)x + dim * index;
                         std::unique_ptr<float[]> copied_query = nullptr;
                         if (is_cosine) {
                             copied_query = CopyAndNormalizeVecs(cur_query, 1, dim);
                             cur_query = copied_query.get();
                         }
-                        index_->search(1, cur_query, k, cur_dis, cur_ids, bitset);
+
+                        faiss::SearchParameters search_params;
+                        search_params.sel = id_selector;
```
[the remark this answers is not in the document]
alexanderguzhva: I'd like to keep it as is in order to maintain the consistency between all index styles in all files, including `ivf.cc`.

src/index/ivf/ivf.cc
alexanderguzhva: Refactored `IndexFlatElkan` a bit.

src/index/ivf/ivf_config.h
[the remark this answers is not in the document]
alexanderguzhva: Maybe, I'll drop `IndexFlatElkan` completely and just replace it with a pure `IndexFlatL2`

thirdparty/faiss/faiss/impl/pq4_fast_scan_search_qbs.cpp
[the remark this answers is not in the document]
alexanderguzhva: force-pushed the update.

thirdparty/faiss/faiss/Clustering.cpp
alexanderguzhva: force-pushed the update. Enforced L2 distance for IP / COSINE in order to support early stop criteria.

alexanderguzhva:
let me verify all these ownership things with sanitizers, just in case of unexpected exceptions

alexanderguzhva:
fixed
/unhold

cmake/libs/libfaiss.cmake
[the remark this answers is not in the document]
chasingegg: Not used in knowhere, just curious about the reason to disable it🤔

src/index/ivf/ivf_config.h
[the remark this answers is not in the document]
liliu-z: Quick question, did we estimate elkan for PQ?

include/knowhere/bitsetview_idselector.h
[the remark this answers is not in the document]
liliu-z: I know, just want to complain the implementation of BitsetView generation.

src/index/flat/flat.cc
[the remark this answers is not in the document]
liliu-z: Ahaha, reasonable. I think all these redundant code need to be cleaned, including `ivf.cc`. I am OK we keep it for now, we can create another PR to do so.

src/index/ivf/ivf.cc
liliu-z: Aha, we can discuss this when we really need to implement the pure Faiss support. It need Knowhere to be extremely light weight, which means no Faiss' config should be exposed to Knowhere.
Anyway let's keep it for now.

thirdparty/faiss/faiss/Clustering.cpp
liliu-z: Did we try how it perform with metric = IP and do Kmeans using L2 vs IP?

src/simd/distances_avx.cc
```diff
@@ -38,80 +40,34 @@ masked_read(int d, const float* x) {
     // cannot use AVX2 _mm_mask_set1_epi32
 }
 
+// trust the compiler to unroll this properly
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
```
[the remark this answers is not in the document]
alexanderguzhva: 1. auto-vectorizers exists in almost every compiler. The quality depends on the compiler version & family, but for a simple code like this one every compiler should work
2. depending on the flags that are used for compiling a particular file (`-mavx`, `-mavx2`, `-mavx512f`, `-mfma`)

cmake/libs/libfaiss.cmake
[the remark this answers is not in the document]
alexanderguzhva: well, I can reenable it, I disable one bcz it is not used, as far as I remember

src/index/ivf/ivf_config.h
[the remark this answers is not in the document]
alexanderguzhva: I did. Same conclusions.

thirdparty/faiss/faiss/Clustering.cpp
alexanderguzhva: yes, if metric=IP(cosine) and kmeans=L2, then it provides a higher recall than if kmeans=IP(cosine)

src/simd/distances_avx.cc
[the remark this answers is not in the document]
cydrain: Hi Alex,
Currently when knowhere init, it will select SIMD API sets by config or AUTO, it can guarantee that Milvus can run on all kinds of intel CPUs.
If we use auto-vectorizer, seems we need not src/simd/hook.cc any more, is it correct ?

src/simd/distances_avx.cc
liliu-z: Since this will be apply not only FAISS indexes, but also others. So it is a little bit weird to have FAISS word here.

src/simd/distances_avx.cc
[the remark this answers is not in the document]
alexanderguzhva: Please let me know which compiler versions are going to be used.

alexanderguzhva:
Republished as #181

chasingegg:
Close this pr as #181 be merged.

PR #93, opened 2023-09-19
Support IVF_FLAT backward compatible when cosine

cydrain:
Issue: #30 

In latest IVF_FLAT index, there is a byte to indicate whether the metric type is COSINE in the index header; while the byte indicator not exist in the IVF_FLAT_NM index header.
So when read IVF_FLAT_NM format, knowhere uses the input config json to tell it's COSINE or not, then convert IVF_FLAT_NM format to the new IVF_FLAT format.

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -692,7 +692,8 @@ IvfIndexNode<T>::Deserialize(const BinarySet& binset, const Config& config) {
         if constexpr (std::is_same<T, faiss::IndexIVFFlat>::value) {
             auto raw_binary = binset.GetByName("RAW_DATA");
             if (raw_binary != nullptr) {
-                ConvertIVFFlatIfNeeded(binset, raw_binary->data.get(), raw_binary->size);
+                const BaseConfig& base_cfg = static_cast<const BaseConfig&>(config);
+                ConvertIVFFlatIfNeeded(binset, base_cfg.metric_type.value(), raw_binary->data.get(), raw_binary->size);
                 // after conversion, binary size and data will be updated
                 reader.data_ = binary->data.get();
                 reader.total_ = binary->size;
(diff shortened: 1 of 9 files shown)


thirdparty/faiss/faiss/invlists/InvertedLists.cpp
```diff
@@ -275,19 +276,33 @@ void ArrayInvertedLists::resize(size_t list_no, size_t new_size) {
 
 void ArrayInvertedLists::restore_codes(
```
liliu-z, line 277: Let's add some comments here. This is a temp code for backward compatibility

src/index/ivf/ivf.cc
```diff
@@ -687,12 +687,15 @@ IvfIndexNode<T>::Deserialize(const BinarySet& binset, const Config& config) {
...
+    const BaseConfig& base_cfg = static_cast<const BaseConfig&>(config);
+    bool is_cosine = IsMetricType(base_cfg.metric_type.value(), knowhere::metric::COSINE);
+
     MemoryIOReader reader(binary->data.get(), binary->size);
     try {
         if constexpr (std::is_same<T, faiss::IndexIVFFlat>::value) {
             auto raw_binary = binset.GetByName("RAW_DATA");
             if (raw_binary != nullptr) {
-                ConvertIVFFlatIfNeeded(binset, raw_binary->data.get(), raw_binary->size);
+                ConvertIVFFlatIfNeeded(binset, raw_binary->data.get(), raw_binary->size, is_cosine);
```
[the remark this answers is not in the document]
cydrain: updated

thirdparty/faiss/faiss/invlists/InvertedLists.cpp
cydrain: updated

PR #106, opened 2023-09-21
More fixes for clang-16 compiler

alexanderguzhva:
This diff solves two problems.
1. Compilation of benchmarks with clang in Debug mode.
2. Compilation of unit tests and benchmarks with clang on ubuntu 20.04

/kind improvement


liliu-z:
Looks like some issue with swig @Presburger can you help take a look?

PR #116, opened 2023-09-26
Add a missing virtual destructor FileManager::~FileManager()

alexanderguzhva:
/kind improvement


liliu-z:
Good Catch. Thanks for this fix

PR #121, opened 2023-09-27
lower the OS scheduling priority of knowhere threads

zhengbuqian:
Set the OS scheduling priority of knowhere threads to the lowest possible value(a nice value of 19, which is the lowest possible value for a non real time scheduling policy)

/kind improvement


liliu-z:
Good catch!!!!
/lgtm

Issue #132, opened 2023-10-07
[BUG] IVF `GetVectorByIds` interface has concurrence problem


foxspy:
IVF index constructs the id->offset map during getvectors stage :  https://github.com/zilliztech/knowhere/blob/main/src/index/ivf/ivf.cc#L580. This design will cause write/write concurrence when user calls the interface concurrently because the idmap is shared and not thread-safe.

related milvus issue: #https://github.com/milvus-io/milvus/issues/27390

liliu-z:
@elstic Fixed

elstic:
> @elstic Fixed

Yes, it was fixed.
I'm not getting search failures on the milvus side of the continuous concurrency test

PR #134, opened 2023-10-08
Move make offset map to load stage

foxspy:
issue: #132 
Delete Making map at `getVectors` will cause unexpected error for existing ivf index, because these indexes have no `direct_map` in the binary set. So just move the making operation to load stage to support `getVectors` from existing indexes.


src/index/ivf/ivf.cc
```diff
@@ -320,15 +319,13 @@ IvfIndexNode<T>::Train(const DataSet& dataset, const Config& cfg) {
             index = std::make_unique<faiss::IndexIVFScalarQuantizer>(qzr, dim, nlist, faiss::QuantizerType::QT_8bit,
                                                                      metric.value());
             index->train(rows, (const float*)data);
-            index->make_direct_map(true);
```
liliu-z, line 323: We don't need this for SQ and PQ

foxspy: limited the make operation by having raw data or not

PR #137, opened 2023-10-11
opensource neon code

Presburger:
issue: #141

### DIFF
--- a/src/simd/hook.cc
+++ b/src/simd/hook.cc
@@ -18,6 +18,10 @@
 
 #include "faiss/FaissHook.h"
 
+#if defined(__ARM_NEON)
+#include "distances_neon.h"
+#endif
+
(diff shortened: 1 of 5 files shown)


src/simd/hook.cc
```diff
@@ -124,6 +128,22 @@ fvec_hook(std::string& simd_type) {
         simd_type = "GENERIC";
     }
 #endif
+
+#if defined(__ARM_NEON)
+    fvec_inner_product = fvec_inner_product_neon;
```
liliu-z, line 133: Just curios, is there any context why we are using raw functions with different names instead of a class with template?

Presburger: template cannot work on the code runtime, template need choosed on the compiling time.

Issue #142, opened 2023-10-12
Improve DiskANN Hybrid Search Performance by Introducing Alpha-Strategy


PwzXxm:
We have seen significant improvements in the alpha candidates strategy for HNSW in PR #4.

Experiment on DiskANN.

patelprateek:
@PwzXxm : can you please elaborate on what exactly is the alpha candidates strategy , any pointers or documents ?

PR #143, opened 2023-10-12
Improve DiskANN Hybrid Search via Alpha-Strategy

PwzXxm:
issue: #142 
/kind enhancement

This PR contains:
1. Implement alpha-strategy in DiskANN `cached_beam_search`, which makes at most 5x QPS improvements, 87% recall at filter rate 90%.
2. Only compute PQ distances for unvisited/unfiltered neighbors instead of all neighbors. This yields around 16% and 34% improvements in QPS on Cohere 10M and OpenAI 10M.
3. Remove the dynamic threshold and fix it to 0.93 as HNSW.

image

Note that the recall at the filter rate of 95% does not match with the code, only for experimental purposes.


alwayslove2013 (commented):
/lgtm

PR #148, opened 2023-10-16
Record knowhere index build time

chasingegg:
This pr helps add metric and info log of knowhere index building time.
/kind improvement


hhy3 (commented):
/approve

Issue #149, opened 2023-10-16
[Faiss] some issues in Faiss::simdlib after #144


alexanderguzhva:
#144 allows to compile faiss, faiss_avx2 and faiss_avx512 library codes, depending on the CPU architecture. It seems that faiss (that uses SSE4.2) code leads to use of `simd_emulated.h` file instead of `simd_avx2.h`, and this file triggers run-time SIGSEGV error and incorrect results while running SCANN unittests for Knowhere.

alexanderguzhva:
I've added #150 and #152 that stop SIGSEGV from happening

PR #152, opened 2023-10-16
add proper IndexScaNN instantiation during Faiss deserialization

alexanderguzhva:
related issue: #149 
/kind improvement

### DIFF
--- a/thirdparty/faiss/faiss/impl/index_read.cpp
+++ b/thirdparty/faiss/faiss/impl/index_read.cpp
@@ -902,6 +902,13 @@ Index* read_index(IOReader* f, int io_flags) {
             }
         } else {
             idxrf->refine_index = nullptr;
+            if (dynamic_cast<IndexIVFPQFastScan*>(idxrf->base_index)) {
+                // this is IndexScaNN
+                IndexRefine* idxrf_old = idxrf;
+                idxrf = new IndexScaNN();
+                *idxrf = *idxrf_old;
+                delete idxrf_old;
+            }
         }
         READ1(idxrf->k_factor);
         idxrf->own_fields = true;


thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -902,6 +902,13 @@ Index* read_index(IOReader* f, int io_flags) {
             }
         } else {
             idxrf->refine_index = nullptr;
+            if (dynamic_cast<IndexIVFPQFastScan*>(idxrf->base_index)) {
+                // this is IndexScaNN
+                IndexRefine* idxrf_old = idxrf;
+                idxrf = new IndexScaNN();
+                *idxrf = *idxrf_old;
+                delete idxrf_old;
```
liliu-z, line 910: Can you help elaborate a little bit what is this doing? Thanks

alexanderguzhva: The root of the problem is the following. Serializing / deserializing of the `IndexScaNN` is done wrong. It needs to be a separate code branch and a separate FourCC index in `index_read.cpp` and `index_write.cpp`. But its serialization / deserialization is implemented inside `IndexRefine` branch. So, once `IndexRefine` object is deserialized, it can be converted into `IndexScaNN` object under certain conditions. This is exactly what this piece of code does. Also, this is called 'slicing' in C++ terminology and it extremely error prone and a practice that is recommended to avoid, but it's the meaningful way of converting `IndexRefine` object to `IndexScaNN` object.
So, it is a dirty hack and a technical debug. Again, because `IndexScaNN` ser/de was implemented incorrectly and now we're kinda supposed to maintain it bcz of backward compatibility, if I'm not mistaken.

chasingegg:
we could close this pr since #156 be merged.

PR #155, opened 2023-10-17
Fix a problem with faiss simdlib use v2

### LINKED ISSUE #149: [Faiss] some issues in Faiss::simdlib after #144

chasingegg:
Try fix #149  cc @alexanderguzhva                            
/kind bug

### DIFF
--- a/cmake/libs/libfaiss.cmake
+++ b/cmake/libs/libfaiss.cmake
@@ -8,7 +8,9 @@ knowhere_file_glob(GLOB FAISS_AVX512_SRCS
 
 knowhere_file_glob(GLOB FAISS_AVX2_SRCS
                    thirdparty/faiss/faiss/impl/*avx.cpp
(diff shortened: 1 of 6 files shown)


alexanderguzhva:
/lgtm
I could not crash the unit tests :)

chasingegg:
@Presburger help try this patch to verify if it is ok under your settings.

alexanderguzhva:
Something stopped making sense to me. The performance of SCANN should be the same, no matter whether I use `-mfma -mavx2` for building `faiss`, right? But it is different.
/hold

PR #156, opened 2023-10-18
Add new ser/deser for scann index

chasingegg:
In next releases, scann index will be **officially** supported instead of a **beta** version, we add a new ser/deser for scann index, which will minimize the impact scann index brings if we have any changes to `IndexRefine` in the future.
Revert changes bringed from #77 in `IndexRefine`.
/kind improvement

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -78,7 +78,7 @@ class IvfIndexNode : public IndexNode {
             return false;
         }
         if constexpr (std::is_same<faiss::IndexScaNN, T>::value) {
-            return index_->with_raw_data;
+            return index_->with_raw_data();
         }
         if constexpr (std::is_same<faiss::IndexIVFScalarQuantizer, T>::value) {
             return false;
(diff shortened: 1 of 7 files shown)


alexanderguzhva:
My overall comment would be to restore `IndexRefine` as it was  in the baseline Faiss version (https://github.com/facebookresearch/faiss/blob/main/faiss/IndexRefine.cpp). Same for `IndexRefine`-related code in `index_read.cpp` and `index_write.cpp`

Some more details.

Technically, there is no need to have `bool with_refine;`field at all. So, I'd remove this field completely. It can be expressed as
```
inline bool IndexScaNN::with_refine() const { return (refine_index != nullptr); }
```

Please derive `IndexScaNN` from `IndexRefine`, not from `IndexRefineFlat`. The reason is that `IndexScaNN` has its own custom `IndexScaNN::search()` call and Knowhere code has to patch `IndexRefineFlat::IndexRefineFlat()` constructor for `IndexScaNN` purposes. This can be avoided if `IndexScaNN` is derived directly from `IndexRefine`.

Please override `IndexScaNN::train()` and `IndexScaNN::add()` in the way it is done for `IndexRefine::train()` and `IndexRefine::add()`. The reason is that `IndexRefine` is written in the way that it does not assume that a `refine_index` can be `nullptr` at all. So, this logic needs to go into `IndexScaNN`

alexanderguzhva:
I've also checked the benchmark_float_qps that loads / saves the index files. It seems to work as expected.

PR #157, opened 2023-10-18
Add another file to AVX2 code for PQFastScan

alexanderguzhva:
Baseline
```
[9.631 s] sift-128-euclidean | SCANN | nlist=1024, nprobe=12, reorder_k=256, with_raw_data=1, k=100, R@=0.8000
================================================================================
  thread_num =  1, elapse =  2.228s, VPS = 4488.878
  thread_num =  2, elapse =  1.202s, VPS = 8320.244
  thread_num =  4, elapse =  0.822s, VPS = 12162.166
  thread_num =  8, elapse =  0.785s, VPS = 12739.955
================================================================================
[14.668 s] Test 'sift-128-euclidean/SCANN' done

```

Candidate
```
[7.929 s] sift-128-euclidean | SCANN | nlist=1024, nprobe=12, reorder_k=256, with_raw_data=1, k=100, R@=0.8000
================================================================================
  thread_num =  1, elapse =  1.635s, VPS = 6115.497
  thread_num =  2, elapse =  0.902s, VPS = 11088.023
  thread_num =  4, elapse =  0.636s, VPS = 15733.984
  thread_num =  8, elapse =  0.590s, VPS = 16959.154
================================================================================
[11.691 s] Test 'sift-128-euclidean/SCANN' done

```

/kind improvement

### DIFF
--- a/cmake/libs/libfaiss.cmake
+++ b/cmake/libs/libfaiss.cmake
@@ -11,7 +11,8 @@ knowhere_file_glob(GLOB FAISS_AVX2_SRCS
                    thirdparty/faiss/faiss/impl/pq4_fast_scan_search_1.cpp
                    thirdparty/faiss/faiss/impl/pq4_fast_scan_search_qbs.cpp
                    thirdparty/faiss/faiss/IndexPQFastScan.cpp
-                   thirdparty/faiss/faiss/IndexIVFPQFastScan.cpp)
+                   thirdparty/faiss/faiss/IndexIVFPQFastScan.cpp
+                   thirdparty/faiss/faiss/utils/partitioning.cpp)
 
 list(REMOVE_ITEM FAISS_SRCS ${FAISS_AVX512_SRCS})
 


chasingegg:
This is very tricky.
partitioning.h is used in `ResultHandler.h` and `simd_result_handlers.h`, and simd_result_handlers.h is used in PQFastScan which is compiled with avx2. ResultHandler.h is used in distance.cpp, which can be used by IVF_PQ/IVF_FLAT which is not compiled with avx2, but seems the code snippet will never go to the avx code since the distance type will not be uint16 except PQFastScan.
```
template <class C>
typename C::T partition_fuzzy(
        typename C::T* vals,
        typename C::TI* ids,
        size_t n, size_t q_min,
        size_t q_max, size_t* q_out) {
#ifdef __AVX2__
    constexpr bool is_uint16 = std::is_same<typename C::T, uint16_t>::value;
    if (is_uint16 && is_aligned_pointer(vals)) {
        return simd_partitioning::simd_partition_fuzzy<C>(
                (uint16_t*)vals, ids, n, q_min, q_max, q_out);
    }
#endif
    return partitioning::partition_fuzzy_median3<C>(
            vals, ids, n, q_min, q_max, q_out);
}
```

However, if we use avx2 to compile partitioning_avx2.cpp, it is likely to be auto optimized to avx2 simd under gcc9.4.0 compiler
```
template <class C>
void count_lt_and_eq(
        const typename C::T* vals,
        size_t n,
        typename C::T thresh,
        size_t& n_lt,
        size_t& n_eq) {
    n_lt = n_eq = 0;

    for (size_t i = 0; i < n; i++) {
        typename C::T v = *vals++;
        if (C::cmp(thresh, v)) {
            n_lt++;
        } else if (v == thresh) {
            n_eq++;
        }
    }
}
```
We can reproduce the issue by setting topk > 100, and the code will use `ReservoirResultHandler`, and will hit the code snippet.
Therefore, we need to add *.cpp to avx2 compilation very carefully, if it is widely used. I will vote for dividing the avx2 part code to a new partitioning_avx2.cpp to avoid this problem in #159

chasingegg:
We could close this pr since #159 has beed merged.

PR #167, opened 2023-10-31
Improve IVF index mmap performance by cache policy

yah01:
/kind improvement
As we tested for HSNW index, `MADV_RANDOM` perform the best


PwzXxm:
Please provide some results from benchmarking :)

alexanderguzhva:
two comments
1. It is a OS-specific change and it won't work in MSVC. Is it acceptable?
2. Are there any other locations where the same optimization can be applied? 
Thanks

yah01:
> two comments
> 
> 1. It is an OS-specific change and it won't work in MSVC. Is it acceptable?
> 2. Are there any other locations where the same optimization can be applied?
>    Thanks

1. As I know Milvus stopped windows support, and now Knowhere can't be compiled on Windows (due to HNSW impl), is Knowhere targeting on Winodws support? @liliu-z 
2. For now we supports only IVF and HNSW index with mmap, HNSW has been optimized before this, so this is all

Issue #171, opened 2023-11-02
SCANN range search recall is too low


cydrain:
SCANN range search recall is too low:
```
SCANN {"metric_type": "L2", "radius": 65.0} 0.0088
SCANN {"metric_type": "IP", "radius": 8.699999809265137} 0.003
SCANN {"metric_type": "COSINE", "radius": 0.20000000298023224} 0.0018
```

range search recall for IVF_FLAT with same param:
```
IVF_FLAT {"metric_type": "L2", "radius": 65.0} 0.1613
IVF_FLAT {"metric_type": "IP", "radius": 8.699999809265137} 0.1643
IVF_FLAT {"metric_type": "COSINE", "radius": 0.20000000298023224} 0.1675
```

cydrain:
repo: git@github.com:cydrain/knowhere-test.git
branch: caiyd_refactor

PR #179, opened 2023-11-06
Make scann range search results more accurate

chasingegg:
Related #171 
Range search's recall of ScaNN is poor due to quantization. Enlarge radius to take more candidates into consideration.

Before change
```
SCANN {"metric_type": "L2", "radius": 65.0} 0.0088
SCANN {"metric_type": "IP", "radius": 8.699999809265137} 0.003
SCANN {"metric_type": "COSINE", "radius": 0.20000000298023224} 0.0018
```
After
```
SCANN {"metric_type": "L2", "radius": 65.0} 0.2133
SCANN {"metric_type": "IP", "radius": 8.699999809265137} 0.0339
SCANN {"metric_type": "COSINE", "radius": 0.20000000298023224} 0.0272
```

### DIFF
--- a/thirdparty/faiss/faiss/IndexIVFPQFastScan.cpp
+++ b/thirdparty/faiss/faiss/IndexIVFPQFastScan.cpp
@@ -1332,7 +1332,8 @@ void IndexIVFPQFastScan::range_search_implem_12(
     TIC;
 
     // prepare the result handlers
-    std::unique_ptr<RangeSearchResultHandler<C, true>> handler(new RangeSearchResultHandler<C, true>(result, radius, 0, bitset));
+    float radius_thresh = C::is_max ? radius * 1.2 : radius * 0.6;  // to make range search result more accurate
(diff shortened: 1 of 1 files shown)


thirdparty/faiss/faiss/IndexIVFPQFastScan.cpp
```diff
@@ -1332,7 +1332,8 @@ void IndexIVFPQFastScan::range_search_implem_12(
     TIC;
 
     // prepare the result handlers
-    std::unique_ptr<RangeSearchResultHandler<C, true>> handler(new RangeSearchResultHandler<C, true>(result, radius, 0, bitset));
+    float radius_thresh = C::is_max ? radius * 1.2 : radius * 0.6;  // to make range search result more accurate
```
alexanderguzhva, line 1335: Such a change should never be done on the Faiss side, but on the Knowhere side. Please move it to `ivf.cc`.

alexanderguzhva (changes requested):

chasingegg:
close this, need further investigation

PR #180, opened 2023-11-06
Remove unused stats in PQFastScan

chasingegg:
Remove unused stats in PQFastScan, where milvus could fail on global variable initialization on an sse4_2 machine.

### DIFF
--- a/thirdparty/faiss/faiss/IndexIVFPQFastScan.h
+++ b/thirdparty/faiss/faiss/IndexIVFPQFastScan.h
@@ -230,25 +230,4 @@ struct IndexIVFPQFastScan : IndexIVF {
             const BitsetView bitset = nullptr) const;
 };
 
-struct IVFFastScanStats {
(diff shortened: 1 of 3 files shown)


chasingegg:
> @chasingegg Could you please elaborate why the failure can happen? Thanks

We will get `_GLOBAL__sub_I_IndexPQFastScan.cpp` on sse4_2, which means the initialization code for a global object or function in the file caused the problem, that is FastScan_stats.

alexanderguzhva:
> > @chasingegg Could you please elaborate why the failure can happen? Thanks
> 
> We will get `_GLOBAL__sub_I_IndexPQFastScan.cpp` on sse4_2, which means the initialization code for a global object or function in the file caused the problem, that is FastScan_stats. By removing this, the problem disappears.

Also, as far I as remember, some of the veeeeeery recent commits to Faiss moved this stats from global to local variable, which is passed in a function. We can go this way as well.

PR #182, opened 2023-11-06
Upgrade conan to 1.61.0

jiaoew1991:
/kind improvement


alexanderguzhva:
The only problem will be that the compiling in conda will require some additional workarounds.
/lgtm

Issue #187, opened 2023-11-09
Incorrect thread number when running Knowhere alone


PwzXxm:
Running Knowhere in the K8s pod gets the system thread number to initialize the thread pool.

For example, if the instance has 64 threads, `std::thread::hardware_concurrency()` will return `64` regardless of k8s's CPU requests and limits.

This issue only exists when running Knowhere alone. Milvus + Knowhere does not have this issue because of a hook function Milvus uses.

PR #188, opened 2023-11-09
Fallback to Set Thread Num by Cgroups when Users Did Not Do So

PwzXxm:
issue: #187 
/kind enhancement

Read system cgroups to find out the CPU quota and period to get the correct limited CPU num under the K8s environment when running Knowhere alone, i.e., without Milvus.

For instance, if an instance has 64 CPUs, k8s limits it to 16c,
* Before the change, Knowhere initializes the thread pool to 64 and uses 25% of each CPU.
* After the change, Knowhere initializes the thread pool to 16 and uses 100% of each CPU.

If it fails to read from groups, fallback `hardware_concurrency()`.
For systems other than Linux, fallback as well.

Support cgroups V1 only for now. Maybe support V2 later.

### DIFF
--- a/include/knowhere/comp/thread_pool.h
+++ b/include/knowhere/comp/thread_pool.h
@@ -20,12 +20,12 @@
 #include <thread>
 #include <utility>
 
+#include "cgroup.h"
 #include "folly/executors/CPUThreadPoolExecutor.h"
 #include "folly/futures/Future.h"
 #include "knowhere/log.h"
 
 namespace knowhere {
-
 class ThreadPool {
 #ifdef __linux__
  private:
@@ -138,7 +138,7 @@ class ThreadPool {
     static std::shared_ptr<ThreadPool>
     GetGlobalBuildThreadPool() {
         if (global_build_thread_pool_size_ == 0) {
-            InitThreadPool(std::thread::hardware_concurrency(), global_build_thread_pool_size_);
+            InitThreadPool(CgroupCpuReader::GetCpuNum(), global_build_thread_pool_size_);
             LOG_KNOWHERE_WARNING_ << "Global Build ThreadPool has not been initialized yet, init it with threads num: "
                                   << global_build_thread_pool_size_;
         }
@@ -149,7 +149,7 @@ class ThreadPool {
     static std::shared_ptr<ThreadPool>
     GetGlobalSearchThreadPool() {
         if (global_search_thread_pool_size_ == 0) {
-            InitThreadPool(std::thread::hardware_concurrency(), global_search_thread_pool_size_);
+            InitThreadPool(CgroupCpuReader::GetCpuNum(), global_search_thread_pool_size_);
             LOG_KNOWHERE_WARNING_ << "Global Search ThreadPool has not been initialized yet, init it with threads num: "
                                   << global_search_thread_pool_size_;
         }
@@ -163,7 +163,7 @@ class ThreadPool {
      public:
         explicit ScopedOmpSetter(int num_threads = 0) {
             if (global_build_thread_pool_size_ == 0) {  // this should not happen in prod
-                omp_before = omp_get_max_threads();
+                omp_before = CgroupCpuReader::GetCpuNum();
             } else {
                 omp_before = global_build_thread_pool_size_;
             }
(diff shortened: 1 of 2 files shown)


xiaofan-luan:
The easiest way is simply passed in from golang?

xiaofan-luan:
as a library, I think it make sense to receive a params of concurrency?
If user don't pass in then we fallback to current logic

PwzXxm:
> The easiest way is simply passed in from golang?

Yes, if you run Milvus + Knowhere, Milvus will use a hook to pass it to Knowhere. This PR fixs when Knowhere is running alone, e.g., VectorDBBench now supports benchmark Knowhere/Cardinal alone and we encountered this issue.

PwzXxm:
> as a library, I think it make sense to receive a params of concurrency? If user don't pass in then we fallback to current logic

Right. This PR will not change this logic. Use the Milvus passed params first; if not received anything, fallback to read from cgroups; fails to read from cgroups, fallback to system thread num.

include/knowhere/comp/thread_pool.h
```diff
@@ -14,18 +14,122 @@
...
+        std::vector<std::string> tokens;
+        std::string token;
+        std::istringstream tokenStream(s);
+        while (std::getline(tokenStream, token, delimiter)) {
+            tokens.push_back(token);
+        }
+        return tokens;
+    }
+
+    static auto
```
zhengbuqian, line 51: For better readability can you provide examples of cpu related lines, like the format and meaning of each field?

include/knowhere/comp/thread_pool.h
```diff
@@ -14,18 +14,122 @@
...
+                    std::find(sub_sys.cbegin(), sub_sys.cend(), "cpu") != sub_sys.cend()) {
+                    return fields[2];
+                }
+            }
+        }
+        throw std::runtime_error("Unable to get cgroup file");
+    }
+
+    static auto
+    getCgroupMountPath() -> std::pair<fs::path, fs::path> {
```
zhengbuqian, line 68: ditto

include/knowhere/comp/thread_pool.h
```diff
@@ -14,18 +14,122 @@
...
+                int ret;
+                if (fin >> ret) {
+                    return ret;
+                }
+                throw std::runtime_error("Failed to get int value from " + path.generic_string());
+            };
+            auto [root, mount_path] = getCgroupMountPath();
+            auto cgroup_path = (mount_path / fs::relative(getCgroupCpuPath(), root));
+            auto quota_file_path = cgroup_path / "cpu.cfs_quota_us";
+            auto period_file_path = cgroup_path / "cpu.cfs_period_us";
```
zhengbuqian, line 111: Add more comments on what files are read, how they are named, what content format is the file.

zhengbuqian (commented):
LGTM overall but need more comments to explain how it works.

include/knowhere/comp/thread_pool.h
PwzXxm: will do

liliu-z:
Why we cannot simply call the init method to set the number from lib caller side?

PwzXxm:
> Why we cannot simply call the init method to set the number from lib caller side?

This requires an additional pyknowhere interface and a manual setup every time. Then the question is, should Knowhere use ALL of the CPUs even if it was limited to a smaller number of them BY DEFAULT?

include/knowhere/comp/thread_pool.h
PwzXxm: done

alexanderguzhva:
Just to confirm. What is going to happen if I run under `taskset` or `numactl`?

zhengbuqian:
> Just to confirm. What is going to happen if I run under `taskset` or `numactl`?

`taskset` or `numactl` doesn't rely on cgroups on controlling the cpu usage of processes, they do that by setting the process's CPU affinity mask using `sched_setaffinity` syscall. So this won't detect the limitation set by `taskset` or `numactl`.

PwzXxm:
> Just to confirm. What is going to happen if I run under `taskset` or `numactl`?

Good question. I just tested both. 
image
image

Buqian is correct that they do not interact with each other. So for this PR, you can view it as an additional shield protection under k8s env. If you run `taskset` or `numactl` and bind to a single CPU, under an 8c K8s pod, on an instance with 64 CPU:
* Before the change, you run the process on a single CPU with 64 threads.
* After this change, you run the process on a single CPU with 8 threads.
In this case, it might be better to call the setters or we might create a separate PR for this.
This is an enchantment for the Knowhere default behavior and the initiative is to be less error-prone on testing.

PR #189, opened 2023-11-13
Add iterator support to pyknowhere interface.

zhengbuqian:
Add GetAnnIterator method to the swigknowhere IndexWrap class to wrap the AnnIterator method, so iterator can also be used in pyknowhere.

/kind improvement

### DIFF
--- a/python/knowhere/knowhere.i
+++ b/python/knowhere/knowhere.i
@@ -51,11 +51,13 @@ import_array();
 %include <std_pair.i>
 %include <std_map.i>
 %include <std_shared_ptr.i>
+%include <std_vector.i>
 %include <exception.i>
 %shared_ptr(knowhere::DataSet)
 %shared_ptr(knowhere::BinarySet)
 %template(DataSetPtr) std::shared_ptr<knowhere::DataSet>;
 %template(BinarySetPtr) std::shared_ptr<knowhere::BinarySet>;
+%template(int64_float_pair) std::pair<long long int, float>;
 %include <knowhere/expected.h>
 %include <knowhere/dataset.h>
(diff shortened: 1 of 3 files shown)


python/knowhere/knowhere.i
```diff
@@ -113,6 +115,25 @@ public:
     PyThreadState* save;
 };
 
+class AnnIterator {
```
cydrain, line 118: Suggest to rename to "IteratorWrap"

python/knowhere/knowhere.i
```diff
@@ -157,6 +178,22 @@ class IndexWrap {
         }
     }
 
+    std::vector<AnnIterator>
+    GetAnnIterator(knowhere::DataSetPtr dataset, const std::string& json, const knowhere::BitsetView& bitset, knowhere::Status& status) {
```
cydrain, line 182: Rename to "GetIterator"

python/knowhere/knowhere.i
zhengbuqian: Done

python/knowhere/knowhere.i
zhengbuqian: Keeping keyword `Ann` as discussed.

python/knowhere/knowhere.i
```diff
@@ -113,6 +115,25 @@ public:
     PyThreadState* save;
 };
 
+class AnnIteratorWrap {
+ public:
+    AnnIteratorWrap(std::shared_ptr<IndexNode::iterator> it = nullptr) : it_(it) {
```
cydrain, line 120: suggest assert here if it is NULL

PR #191, opened 2023-11-14
Generate diskann cache asynchronously.

cqy123456:
This PR will remove the previous sync cache generate function. 
The prepare phase takes a thread from the thread pool to do the following tasks in the background:
1. init search counter and node_visit_counter[nb];
2. if search counter < min(base row number, 100000), update these variables with samlping queries or input queries;
3.  load cached nodes into mem, insert cached vector and neighbors into diskann cache map;

If diskann calls the destructor, this background thread will : 
- search counter < min(base row number, 100000)，return immediately；
- return until all cached nodes are loaded.

I used sift1m(nb = 1M, nq = 10000, dim = 128) for experiments. 
Diskann parameters：
```
  "diskANN_build_config": {
      "data_path": data_path,
      "max_degree": 56,
      "search_list_size": 100,
      "pq_code_budget_gb":0.125 * nb * dim *4 / (1024*1024 *1024),
      "build_dram_budget_gb": 32.0,
      "num_threads": 16,
      "disk_pq_dims": 0,
      "accelerate_build": False
  },
  "diskANN_prepare_config": {
      "num_threads": 16,
      "search_cache_budget_gb": cache_gb,
      "warm_up": False,
      "use_bfs_cache": False
  },
  "diskANN_query_config": {
      "k": 10,
      "search_list_size": 36,
      "beamwidth": 4
  }
```
The experimental plan as follows:
1. Run 10 time nq = 10000, topk = 10 to generate cache;
2. Run random queries(nb = 10000) to eliminate the impact of lru cache；
3. Run last time search (nq = 10000, topk = 10)
previous version result : 
**sync cache generate: 22.35s
search VPS: 5119.53**
lastest version result : 
**aync cache generate: 33.23s
search VPS without cache: 3059
search VPS with cache:  6000.12**

The worst case is that there is no input query vector, and the background thread completes the task using the sampling queries alone. **And the worst case would take 108s to generate cache.**

### DIFF
--- a/knowhere/index/vector_index/IndexDiskANN.cpp
+++ b/knowhere/index/vector_index/IndexDiskANN.cpp
@@ -290,9 +290,9 @@ IndexDiskANN<T>::Prepare(const Config& config) {
         KNOWHERE_THROW_MSG("Failed to generate cache, num_nodes_to_cache is larger than 1/3 of the total data number.");
     }
     if (num_nodes_to_cache > 0) {
-        std::vector<uint32_t> node_list;
         LOG_KNOWHERE_INFO_ << "Caching " << num_nodes_to_cache << " sample nodes around medoid(s).";
         if (prep_conf.use_bfs_cache) {
+            std::vector<uint32_t> node_list;
             auto gen_cache_successful = TryDiskANNCall<bool>([&]() -> bool {
                 pq_flash_index_->cache_bfs_levels(num_nodes_to_cache, node_list);
                 return true;
@@ -302,29 +302,28 @@ IndexDiskANN<T>::Prepare(const Config& config) {
                 LOG_KNOWHERE_ERROR_ << "Failed to generate bfs cache for DiskANN.";
                 return false;
             }
-        } else {
-            auto gen_cache_successful = TryDiskANNCall<bool>([&]() -> bool {
-                pq_flash_index_->generate_cache_list_from_sample_queries(warmup_query_file, 15, 6, num_nodes_to_cache,
-                                                                         prep_conf.num_threads, node_list);
+            auto load_cache_successful = TryDiskANNCall<bool>([&]() -> bool {
+                pq_flash_index_->load_cache_list(node_list);
                 return true;
             });
 
-            if (!gen_cache_successful.has_value()) {
-                LOG_KNOWHERE_ERROR_ << "Failed to generate cache from sample queries for DiskANN.";
+            if (!load_cache_successful.has_value()) {
+                LOG_KNOWHERE_ERROR_ << "Failed to load cache for DiskANN.";
                 return false;
             }
-        }
-        auto load_cache_successful = TryDiskANNCall<bool>([&]() -> bool {
-            pq_flash_index_->load_cache_list(node_list);
-            return true;
-        });
-
-        if (!load_cache_successful.has_value()) {
-            LOG_KNOWHERE_ERROR_ << "Failed to load cache for DiskANN.";
-            return false;
+        } else {
+            pq_flash_index_->set_async_cache_flag(true);
+            pool_->push([&, cache_num = num_nodes_to_cache,
+                        sample_nodes_file = warmup_query_file]() {
+                try {
+                    pq_flash_index_->generate_cache_list_from_sample_queries(
+                        sample_nodes_file, 15, 6, cache_num);
+                } catch (const std::exception& e) {
+                    LOG_KNOWHERE_ERROR_ << "DiskANN Exception: " << e.what();
+                }
+            });
         }
     }
-
     // warmup
     if (prep_conf.warm_up) {
         LOG_KNOWHERE_DEBUG_ << "Warming up.";
--- a/thirdparty/DiskANN/include/pq_flash_index.h
+++ b/thirdparty/DiskANN/include/pq_flash_index.h
@@ -22,6 +22,7 @@
 #include "percentile_stats.h"
 #include "pq_table.h"
 #include "utils.h"
+#include "semaphore.h"
 #include "windows_customizations.h"
 
 #define MAX_GRAPH_DEGREE 512
@@ -80,16 +81,15 @@ namespace diskann {
 
     DISKANN_DLLEXPORT void load_cache_list(std::vector<uint32_t> &node_list);
 
+    // asynchronously collect the access frequency of each node in the graph
 #ifdef EXEC_ENV_OLS
     DISKANN_DLLEXPORT void generate_cache_list_from_sample_queries(
-        MemoryMappedFiles &files, std::string sample_bin, _u64 l_search,
-        _u64 beamwidth, _u64 num_nodes_to_cache, uint32_t nthreads,
-        std::vector<uint32_t> &node_list);
+        MemoryMappedFiles files, std::string sample_bin, _u64 l_search,
+        _u64 beamwidth, _u64 num_nodes_to_cache);
 #else
     DISKANN_DLLEXPORT void generate_cache_list_from_sample_queries(
         std::string sample_bin, _u64 l_search, _u64 beamwidth,
-        _u64 num_nodes_to_cache, uint32_t num_threads,
-        std::vector<uint32_t> &node_list);
+        _u64 num_nodes_to_cache);
 #endif
 
     DISKANN_DLLEXPORT void cache_bfs_levels(_u64 num_nodes_to_cache,
@@ -127,6 +127,8 @@ namespace diskann {
     DISKANN_DLLEXPORT size_t get_num_medoids() const noexcept;
 
     DISKANN_DLLEXPORT diskann::Metric get_metric() const noexcept;
+
+    DISKANN_DLLEXPORT void set_async_cache_flag(const bool flag);
 
    protected:
     DISKANN_DLLEXPORT void use_medoids_data_as_centroids();
@@ -195,6 +197,7 @@ namespace diskann {
 
     std::string                        disk_index_file;
     std::vector<std::pair<_u32, _u32>> node_visit_counter;
+    std::atomic<_u32>                  search_counter = 0;
 
     // PQ data
     // n_chunks = # of chunks ndims is split into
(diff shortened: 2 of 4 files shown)


thirdparty/DiskANN/include/pq_flash_index.h
```diff
@@ -195,6 +196,7 @@ namespace diskann {
 
     std::string                        disk_index_file;
     std::vector<std::pair<_u32, _u32>> node_visit_counter;
+    _u32                               search_counter = 0;
```
alexanderguzhva, line 199: std::atomic<_u32>

thirdparty/DiskANN/include/utils.h
```diff
@@ -885,7 +885,7 @@ namespace diskann {
     return disk_index_filename + "_centroids.bin";
   }
 
-  inline std::string get_sample_data_filename(const std::string& prefix) {
+  inline std::string get_sample_data_filename(std::string prefix) {
```
alexanderguzhva, line 888: no, please keep it as `const std::string&` :)

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -332,32 +338,48 @@ namespace diskann {
...
+    std::vector<uint32_t> node_list;
     node_list.clear();
     node_list.shrink_to_fit();
     node_list.reserve(num_nodes_to_cache);
     for (_u64 i = 0; i < num_nodes_to_cache; i++) {
       node_list.push_back(this->node_visit_counter[i].first);
     }
-    this->count_visited_nodes = false;
+
+    reinterpret_cast<std::atomic<bool> &>(this->count_visited_nodes)
```
alexanderguzhva, line 370: this is SUPER dirty, I'd turn `count_visited_nodes` into `std::atomic<bool>`

thirdparty/DiskANN/include/pq_flash_index.h
```diff
@@ -233,6 +235,8 @@ namespace diskann {
     // coord_cache
     T *                       coord_cache_buf = nullptr;
     tsl::robin_map<_u32, T *> coord_cache;
+    Semaphore                 semaph;
+    bool                      asyn_generate_cache = false;
```
alexanderguzhva, line 239: `std::atomic<bool> asyn_generate_cache`

thirdparty/DiskANN/include/semaphore.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+#include <mutex>
+#include <condition_variable>
+
+namespace diskann {
+class Semaphore {
```
alexanderguzhva, line 6: I'd use the syntax from https://en.cppreference.com/w/cpp/thread/counting_semaphore, so that we could transition the code to C++20 more easily

thirdparty/DiskANN/include/pq_flash_index.h
zhengbuqian, line 199: CMIIW: the warm up method performs`sample_num` searches and cache the most visited `num_nodes_to_cache` nodes. If user requested searches come in during the time, those are also counted towards `sample_num` and will affect what node to add to `nodes_to_cache`. **The searches performed by the warm up method are all in a single thread**, but to protect from concurrent access by user searches from other threads, we used several atomic primitives to protect `search_counter, node_visit_counter` and others.

I wonder do we have to count user searches during the warm up? If not we can greatly simplify the code. Since the warm up searches are serial we could just count them without using any atomics. 

```cpp
template <warm_up = false>
cached_beam_search(q) {
  if constexpr (warm_up) {
    // no need to protect node_visit_counter
    count_node_visit();
  }
}

void generate_cache_list_from_sample_queries() {
  // single thread operation:
  // perform all `sample_num` searches, not counting user searches
  for (q : warm_up_qs) {
    cached_beam_search<true>(q);
  }
  top_nodes = sorted(node_visit_counter)[:num_nodes_to_cache];
  load_into_cache(top_nodes);
}

void normal_search(q) {
  // multiple threads may enter here at the same time
  // user searches don't affect `node_visit_counter`
  cached_beam_search(q);
}
```

The current code is not completely thread safe anyway, when we have finished warm up, before we flip the bit `reinterpret_cast<std::atomic<bool> &>(this->count_visited_nodes).exchange(false)`, other threads may continue updating `search_counter` and `node_visit_counter` when sorting `node_visit_counter`.

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -302,29 +302,31 @@ IndexDiskANN<T>::Prepare(const Config& config) {
...
+                    cache_num = num_nodes_to_cache,
+                    sample_nodes_file = warmup_query_file
+            ]() {
+                try {
+                pq_flash_index_->generate_cache_list_from_sample_queries(
+                    sample_nodes_file, 15, 6, cache_num);
+                } catch (const std::exception &e) {
+                LOG_KNOWHERE_ERROR_ << "DiskANN Exception: " << e.what();
+                }
+            });
```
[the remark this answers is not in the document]
cqy123456: updated

thirdparty/DiskANN/include/pq_flash_index.h
```diff
@@ -128,6 +127,8 @@ namespace diskann {
 
     DISKANN_DLLEXPORT diskann::Metric get_metric() const noexcept;
 
+    DISKANN_DLLEXPORT void set_asyn_cache_flag(const bool flag);
```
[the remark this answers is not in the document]
cqy123456: updated

thirdparty/DiskANN/include/pq_flash_index.h
cqy123456: updated

thirdparty/DiskANN/src/pq_flash_index.cpp
cqy123456: updated

thirdparty/DiskANN/include/utils.h
cqy123456: updated

thirdparty/DiskANN/include/pq_flash_index.h
cqy123456: updated

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -293,20 +299,20 @@ namespace diskann {
...
-      _u64 num_nodes_to_cache, uint32_t nthreads,
-      std::vector<uint32_t> &node_list) {
+      _u64 num_nodes_to_cache) {
 #endif
     auto s = std::chrono::high_resolution_clock::now();
-    this->count_visited_nodes = true;
+    this->search_counter = 0;
     this->node_visit_counter.clear();
     this->node_visit_counter.resize(this->num_points);
+    reinterpret_cast<std::atomic<bool> &>(this->count_visited_nodes).exchange(true);
```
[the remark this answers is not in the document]
zhengbuqian: here we don't use the value returned by `exchange`, and since both `exchange(read-modify-write)` and `store(write)` are atomic, there is no difference in terms of effect. 

Why is atomic read-modify-write more suitable?

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -332,32 +338,47 @@ namespace diskann {
...
+                        &tmp_result_ids_64, &tmp_result_dists, beamwidth);
+      id++;
     }
 
+    if (this->semaph.IsWaitting()) {
+      this->semaph.Signal();
+      return;
+    }
+
+    this->count_visited_nodes.exchange(false);
```
zhengbuqian, line 357: Simply moving this line doesn't help guarantee "stop updating `this->search_counter` and `this->node_visit_counter` once we have reached `sample_num` searches".

But since the stop condition `this->search_counter.load() == sample_num` is not strict(it's ok to search more than `sample_num` times to building the cache), we can keep the code as is. But please add a comment like "it is intentional and ok that search_counter/node_visit_counter may continue being updated after search_counter has reached sample_num before we flip count_visited_nodes".

cqy123456: if count_visited_nodes == true： diskann will update node_visit_counter in each iteration, and update search_counter in each query. So when generate cache task done,  i want to release node_visit_counter memory and reduce some search time from updating node_visit_counter and search_counter.

thirdparty/DiskANN/src/pq_flash_index.cpp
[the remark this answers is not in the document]
cqy123456: I think exchange is ： lock->read value to register -> write to mem -> unlock; store : lock -> wirte to mem -> unlock. The update of the counter is not strict, and store can be faster, so i update the code.

thirdparty/DiskANN/src/pq_flash_index.cpp
zhengbuqian: My point is:

In general sync mechanisms are used to guarantee strong correctness, but here we use locks/atomics while tolerating a degree of inaccuracy(as you mentioned in another comment that the update of the counter is not strict). So we should leave a comment explicitly stating such.

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -332,32 +338,47 @@ namespace diskann {
       return;
     }
 
-    std::vector<int64_t> tmp_result_ids_64(sample_num, 0);
-    std::vector<float>    tmp_result_dists(sample_num, 0);
+    int64_t tmp_result_ids_64;
```
liliu-z, line 341: Any race condition risks?

liliu-z (commented):
/lgtm

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -332,32 +338,47 @@ namespace diskann {
...
+    std::vector<uint32_t> node_list;
     node_list.clear();
     node_list.shrink_to_fit();
     node_list.reserve(num_nodes_to_cache);
     for (_u64 i = 0; i < num_nodes_to_cache; i++) {
       node_list.push_back(this->node_visit_counter[i].first);
     }
-    this->count_visited_nodes = false;
+
+    this->node_visit_counter.clear();
```
liliu-z, line 371: resize to 0

Issue #193, opened 2023-11-15
[feat] Add sparse vector support to knowhere


zhengbuqian:
Work Items:

* [x] Search w/ & w/o filter
  * [x] columnar brute force sparse vector index
  * [x] wand based sparse vector index
* [x] other feature
  * [x] range search
  * [x] iterator
  * [x] index mmap
  * [x] get vector by id
  * [ ] etc...
* [x] Unit Test
  * [x] Add brute force search method to provide ground truth
  * [x] Add tests for search/filtered-search/serialize/deserialize/
  * [x] Add tests for growing segment(add-search and add-search again)
* [x] E2E test

zhengbuqian:
all features in comments are now implemented, closing

PR #199, opened 2023-11-17
[feat] Add sparse index support to knowhere

zhengbuqian:
/kind feature
#193 

This PR adds the basic framework of supporting sparse index in knowhere, with 2 index types: inverted index and wand. 

In this PR only Search is supported. 

Support for growing segment is sub-optimal, we naively double the underlying data structure each time we used up the current space.

Added basic unit tests:
* Search without filtering, with different dataset sizes/recall/sparisity, etc.
* Search with filtering, with different filter rate.

### DIFF
--- a/include/knowhere/utils.h
+++ b/include/knowhere/utils.h
@@ -100,7 +100,7 @@ namespace sparse {
 // type used to represent the id of a vector in the index interface.
 // this is same as other index types.
 using label_t = int64_t;
-// type used to represent the id of a vector inside the index.
+// type used to represent the id, indices and indptr of a vector inside the index.
 using table_t = uint32_t;
 
 /**
--- a/include/knowhere/utils.h
+++ b/include/knowhere/utils.h
@@ -179,25 +179,27 @@ struct Neighbor {
     }
 };
 
-// when pushing new elements into a MinMaxHeap, only the `capacity` smallest elements are kept.
-// pop()/top() returns the largest element out of those `capacity` smallest elements.
+// When pushing new elements into a MaxMinHeap, only the `capacity` largest elements are kept.
+// pop()/top() returns the smallest element out of those `capacity` largest elements.
 template <typename T = float>
-class MinMaxHeap {
+class MaxMinHeap {
  public:
-    explicit MinMaxHeap(int capacity) : capacity_(capacity), pool_(capacity) {
+    explicit MaxMinHeap(int capacity) : capacity_(capacity), pool_(capacity) {
     }
     void
     push(table_t id, T dist) {
         if (size_ < capacity_) {
             pool_[size_] = {id, dist};
-            std::push_heap(pool_.begin(), pool_.begin() + ++size_);
-        } else if (dist < pool_[0].distance) {
+            size_ += 1;
+            std::push_heap(pool_.begin(), pool_.begin() + size_, std::greater<Neighbor<T>>());
+        } else if (dist > pool_[0].distance) {
             sift_down(id, dist);
         }
     }
     table_t
     pop() {
-        std::pop_heap(pool_.begin(), pool_.begin() + size_--);
+        std::pop_heap(pool_.begin(), pool_.begin() + size_, std::greater<Neighbor<T>>());
+        size_ -= 1;
         return pool_[size_].id;
     }
     [[nodiscard]] size_t
@@ -224,10 +226,10 @@ class MinMaxHeap {
         for (; 2 * i + 1 < size_;) {
             size_t j = i;
             size_t l = 2 * i + 1, r = 2 * i + 2;
-            if (pool_[l].distance > dist) {
+            if (pool_[l].distance < dist) {
                 j = l;
             }
-            if (r < size_ && pool_[r].distance > std::max(pool_[l].distance, dist)) {
+            if (r < size_ && pool_[r].distance < std::min(pool_[l].distance, dist)) {
                 j = r;
             }
             if (i == j) {
@@ -241,7 +243,7 @@ class MinMaxHeap {
 
     size_t size_ = 0, capacity_;
     std::vector<Neighbor<T>> pool_;
-};  // class MinMaxHeap
+};  // class MaxMinHeap
 
 }  // namespace sparse
 
--- a/include/knowhere/utils.h
+++ b/include/knowhere/utils.h
@@ -14,6 +14,8 @@
 #include <strings.h>
 
 #include <algorithm>
+#include <cstddef>
+#include <functional>
 #include <vector>
 
 #include "knowhere/binaryset.h"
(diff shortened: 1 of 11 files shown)


include/knowhere/utils.h
```diff
@@ -92,4 +94,155 @@ readBinaryPOD(R& in, T& podRef) {
...
+        if (size_ < capacity_) {
+            pool_[size_] = {id, dist};
+            std::push_heap(pool_.begin(), pool_.begin() + ++size_);
+        } else if (dist < pool_[0].distance) {
+            sift_down(id, dist);
+        }
+    }
+    table_t
+    pop() {
+        std::pop_heap(pool_.begin(), pool_.begin() + size_--);
```
alexanderguzhva, line 200: Please
```
std::pop_heap(pool_.begin(), pool_.begin() + size_);
size_ -= 1;
```

src/common/comp/brute_force.cc
```diff
@@ -343,4 +343,96 @@ BruteForce::RangeSearch(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+                    continue;
+                }
+                float dist = 0.0f;
+                for (int64_t k = indptr[j]; k < indptr[j + 1]; ++k) {
+                    auto it = query.find(indices[k]);
+                    if (it != query.end()) {
+                        dist += it->second * data[k];
+                    }
+                }
+                if (dist > 0) {
```
alexanderguzhva, line 415: the code does not distinguish two very different situations:
1) `dist=0` because no item was found in `query`
2) `dist=0` because the match is exact

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+            refine_factor = 1;
+        }
+
+        std::vector<std::pair<IndicesT, T>> q_vec(len);
+        for (size_t i = 0; i < len; ++i) {
+            q_vec[i] = std::make_pair(indices[i], data[i]);
+        }
+        std::sort(q_vec.begin(), q_vec.end(),
+                  [](const auto& lhs, const auto& rhs) { return std::abs(lhs.second) > std::abs(rhs.second); });
+        while (q_vec.size() && q_vec[0].second * drop_ratio_search > q_vec.back().second) {
```
alexanderguzhva, line 177: `while (q_vec.size() && ...)` -> `while (!q_vec.empty() && ...)`

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    [[nodiscard]] size_t
+    n_cols() const {
+        return n_cols_;
+    }
+
+ private:
+    [[nodiscard]] float
+    dot_product(std::unordered_map<IndicesT, T> q_map, table_t u) const {
```
alexanderguzhva, line 229: `dot_product(const std::unordered_map<IndicesT, T>& q_map, table_t u) const {`

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+                        break;
+                    }
+                    std::swap(cursors[i], cursors[i - 1]);
+                }
+            }
+        }
+    }
+
+    void
+    refine_and_collect(std::unordered_map<IndicesT, T>& q_map, MinMaxHeap<T> inaccurate, size_t k, float* distances,
```
alexanderguzhva, line 402: `const std::unordered_map`

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+            search_brute_force(q_vec, heap, bitset);
+        } else {
+            search_wand(q_vec, heap, bitset);
+        }
+
+        // no refinement needed
+        if (refine_factor == 1) {
+            collect_result(heap, distances, labels);
+        } else {
+            std::unordered_map<IndicesT, T> q_map;
```
alexanderguzhva, line 192: the size of the `q_map` is known beforehand (it is `len`). So, please construct `q_map` with the provided number of buckets (I'd try `4 * len` as a start), so 
```
std::unordered_map<IndicesT, T> q_map(4 * len);
```
The number of buckets impacts the performance of the following `dot_product()`

alexanderguzhva: Or I can tweak the performance as soon as this commit lands, there are many things to play with there :)

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+                heap.pop();
+                heap.emplace(u, -dist_acc);
+            }
+        }
+        collect_result(heap, distances, labels);
+    }
+
+    template <typename HeapType>
+    void
+    collect_result(HeapType heap, float* distances, label_t* labels) const {
```
alexanderguzhva, line 423: `HeapType& heap`

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+        Cursor() = default;
+        Cursor(const std::vector<Neighbor<T>>* lut, size_t num_vec, float max_score, float q_value,
+               const BitsetView* bitset)
+            : lut_(lut), num_vec_(num_vec), max_score_(max_score), q_value_(q_value), bitset_(bitset) {
+            while (loc_ < lut_->size() && !bitset_->empty() && bitset_->test(cur_vec_id())) {
+                loc_++;
+            }
+        }
+        Cursor(const Cursor& rhs) = delete;
+        Cursor(Cursor&& rhs) noexcept {
```
alexanderguzhva, line 271: Won't `Cursor(Cursor&& rhs) noexcept = default;` work?
Same question for `Cursor& operator=(Cursor&& rhs) noexcept`

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,449 @@
...
+            swap(lhs.max_score_, rhs.max_score_);
+            swap(lhs.q_value_, rhs.q_value_);
+            // all cursors share the same bitset
+        }
+        const std::vector<Neighbor<T>>* lut_;
+        size_t loc_ = 0;
+        size_t num_vec_ = 0;
+        float max_score_ = 0.0f;
+        float q_value_ = 0.0f;
+        const BitsetView* bitset_;
```
alexanderguzhva, line 340: BitsetView is a lightweight structure (a wrapper around the pointer, basically) that should be passed by reference

alexanderguzhva:
the functionality seems to be fine, but what about the unit tests running time, is it expected?

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    [[nodiscard]] size_t
+    n_cols() const {
+        return n_cols_;
+    }
+
+ private:
+    [[nodiscard]] float
+    dot_product(const std::unordered_map<IndicesT, T>& q_map, table_t u) const {
```
hhy3, line 231: std::unordered_map seems with a bad performance as I tested. It's better to use two arrays `indices` and `data`

zhengbuqian:
> the functionality seems to be fine, but what about the unit tests running time, is it expected?

I added some large unit tsets so ut took quite long, I have removed them. Will revert the change in ut yaml before merge.

zhengbuqian:
run-e2e

zhengbuqian:
Use https://github.com/zilliztech/knowhere/pull/341 instead

PR #211, opened 2023-11-23
Update gen_fbin_file.cpp to support binary data type

cydrain:


benchmark/hdf5/gen_fbin_file.cpp
```diff
@@ -44,20 +44,30 @@ class Create_FBIN : public Benchmark_hdf5, public ::testing::Test {
     TearDown() override {
     }
 
+    template <bool is_binary>
```
alexanderguzhva, line 47: This overcomplicates the code.
Please put the new functionality into new functions, such as `fbin_write_binary()` or `fbin_read_binary()`.

alexanderguzhva (changes requested):

benchmark/hdf5/gen_fbin_file.cpp
cydrain: updated

PR #219, opened 2023-11-27
Switch to brute force for hnsw when topk is super large

### LINKED ISSUE #218: [Enhancement] switch to BF for hnsw when topk is super large
When topk is super large, there are two problems:
- some isolated points could not be reached by index, related milvus issue https://github.com/milvus-io/milvus/issues/28603
- performance degradation

So we can turn to BF for this situation.

chasingegg:
Fix #218, Set threshold to 0.5 for safety.
image
image

### DIFF
--- a/thirdparty/hnswlib/hnswlib/hnswalg.h
+++ b/thirdparty/hnswlib/hnswlib/hnswalg.h
@@ -40,8 +40,9 @@
 namespace hnswlib {
 typedef unsigned int tableint;
 typedef unsigned int linklistsizeint;
-constexpr float kHnswSearchKnnBFThreshold = 0.93f;
-constexpr float kHnswSearchRangeBFThreshold = 0.97f;
+constexpr float kHnswSearchKnnBFFilterThreshold = 0.93f;
+constexpr float kHnswSearchRangeBFFilterThreshold = 0.97f;
+constexpr float kHnswSearchBFTopkThreshold = 0.5f;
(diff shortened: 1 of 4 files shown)


thirdparty/hnswlib/hnswlib/hnswalg.h
```diff
@@ -1240,11 +1241,16 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
         // do bruteforce search when delete rate high
         if (!bitset.empty()) {
-            if (bitset.count() >= (cur_element_count * kHnswSearchKnnBFThreshold)) {
+            if (bitset.count() >= (cur_element_count * kHnswSearchKnnBFFilterThreshold)) {
                 return searchKnnBF(query_data, k, bitset);
             }
         }
 
+        // do bruteforce search when topk is super large
+        if (k >= (cur_element_count * kHnswSearchBFTopkThreshold)) {
```
foxspy, line 1250: The definition of `big top_k` scenario  seems also need take the `filter_ratio` into consideration. For example, if filter_ratio is 60%, then `top_k = cur_element_count  * 40%` is big enough because all unfiltered data will be selected,  and graph searching may return missing result for nodes not fully connected. However, more tests should be added. Hotfix now?

chasingegg: Good point! Updated

PR #223, opened 2023-11-27
Merge faiss changes from baseline

alexanderguzhva:
Propagate changes from the Faiss master branch.
Most of the changes are tiny or related to some small-sized new features.

I'm pushing it in the way of multiple commits in the same PR, because doing it one by one is an overkill due to the nature of changes.

/kind improvement


liliu-z:
@alexanderguzhva Do you think we need to keep all commits msg in this PR? This means we cannot squash, and the default/only allowed merge method is squash right now

alexanderguzhva:
Superseded by #230

PR #226, opened 2023-11-28
Add async thread pool for generating diskann cache and catch unexpected return.

cqy123456:


knowhere/common/ThreadPool.cpp
```diff
@@ -60,4 +60,18 @@ ThreadPool::GetGlobalThreadPool() {
...
+    if (global_thread_pool_size_ == 0) {
+        std::lock_guard<std::mutex> lock(global_thread_pool_mutex_);
+        if (global_thread_pool_size_ == 0) {
+            global_thread_pool_size_ = std::thread::hardware_concurrency();
+        }
+    }
+    uint32_t async_thread_pool_size = int(std::ceil(global_thread_pool_size_ / 2.0));
+    LOG_KNOWHERE_WARNING_ << "async thread pool size with thread number:" << async_thread_pool_size;
+    static auto async_pool = std::make_shared<ThreadPool>(async_thread_pool_size);
+    return async_pool;
```
liliu-z, line 75: This is not for general purpose, we need to make a static instance in DiskANN instead of here.

liliu-z:
Note: This PR is to solve slow loading time for DiskANN. It lets segments use a shared global threadpool to make cache.

PR #240, opened 2023-11-30
Add I/O tailer for index meta check

cqy123456:


src/index/ivf/ivf.cc
```diff
@@ -846,6 +847,7 @@ IvfIndexNode<faiss::IndexIVFFlat>::Serialize(BinarySet& binset) const {
             faiss::write_index(index_.get(), &writer);
             LOG_KNOWHERE_INFO_ << "write IVF_FLAT, file size " << writer.tellg();
         }
+        writer.addtailer(Type(), this->version_);
```
foxspy, line 850: Can move the `addtailer` to index.cc?  Just a suggestion, maybe different to index ser/des with file.

src/io/tailer.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace knowhere {
+union Tailer {
+    uint8_t bytes[KNOWHERE_TAILER_SIZE];
+    struct Meta {
+        uint32_t flag;
+        uint64_t bin_size;
+        IndexVersion version;
+        uint32_t checksum;
+        char index_name[MAX_INDEX_NAME_SIZE + 1];
```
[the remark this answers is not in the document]
cqy123456: Tailer has 512bytes，it can add more meta info within 512bytes. Maybe we use larger size ?

src/io/memory_io.h
```diff
@@ -100,6 +102,9 @@ struct MemoryIOWriter : public faiss::IOWriter {
     tellg() const {
         return rp_;
     }
+
+    void
```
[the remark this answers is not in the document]
cqy123456: I will remove addtailer and valid check in IOReader/Writer, and add AddTailerForMultiFiles/AddTailerForMemoryIO functions.

src/io/trailer.h
```diff
@@ -0,0 +1,86 @@
...
+        return meta.bin_size;
+    }
+    uint32_t
+    GetCheckSum() {
+        return meta.checksum;
+    }
+    void
+    SetIndexName(std::string index_name) {
+        if (index_name.size() > MAX_INDEX_NAME_SIZE) {
+            LOG_KNOWHERE_ERROR_ << "the size of index name larger than " << MAX_INDEX_NAME_SIZE;
```
zhengbuqian, line 57: we should not just silently ignore the error. we should throw exception or return false so the caller of this method is aware of the error and can handle it.

src/io/trailer.h
```diff
@@ -0,0 +1,86 @@
...
+namespace knowhere {
+union Trailer {
+    uint8_t bytes[KNOWHERE_TRAILER_SIZE];
+    struct Meta {
+        uint32_t flag;
+        uint64_t bin_size;
+        IndexVersion version;
+        uint32_t checksum;
+        char index_name[MAX_INDEX_NAME_SIZE + 1];
+    } meta;
```
zhengbuqian, line 30: consider making members private.

src/io/trailer.cc
```diff
@@ -0,0 +1,156 @@
...
+    for (auto i = 0; i < size; i++) {
+        checksum ^= data[i];  // xor
+    }
+    return checksum;
+}
+
+uint32_t
+GetFilesCheckSum(std::vector<std::string> files) {
+    uint32_t checksum = 0;
+    auto buffer = std::shared_ptr<uint8_t[]>(new uint8_t[kBlockSize]);
```
zhengbuqian, line 31: use unique ptr as it is never shared.

src/io/trailer.h
```diff
@@ -0,0 +1,86 @@
...
+    };
+    bool
+    TrailerValidCheck() {
+        return meta.flag == TRAILER_VALID_FLAG;
+    }
+    std::string
+    GetIndexName() {
+        return std::string(meta.index_name);
+    }
+    int32_t
```
zhengbuqian, line 42: why not use type alias IndexVersion?

src/io/trailer.cc
```diff
@@ -0,0 +1,156 @@
...
+        return true;
+    }
+
+    auto trailer_ptr = std::make_unique<Trailer>();
+    auto pre_rp = reader.tellg();
+    reader.seekg(bin_size);
+    reader.read(trailer_ptr.get(), KNOWHERE_TRAILER_SIZE);
+    reader.seekg(pre_rp);
+
+    if (!trailer_ptr->TrailerValidCheck()) {
```
zhengbuqian, line 73: what if this file has no trailer but these 4 bytes happens to be the same as TRAILER_VALID_FLAG?

is it possible to use version number to decide whether a index file should include trailer?

src/io/trailer.cc
```diff
@@ -0,0 +1,156 @@
...
+    auto fsize = reader.tellg();
+
+    reader.seekg(0, std::ios::beg);
+    if (fsize != KNOWHERE_TRAILER_SIZE) {
+        LOG_KNOWHERE_ERROR_ << "Trailer size (" << fsize << ")not correct.";
+        return false;
+    }
+    auto trailer_ptr = std::make_unique<Trailer>();
+    reader.read((char*)trailer_ptr.get()->bytes, KNOWHERE_TRAILER_SIZE);
+    if (!trailer_ptr->TrailerValidCheck()) {
```
zhengbuqian, line 132: avoid duplicating those checks in CheckTrailerForFiles and CheckTrailerForMemoryIO. trailer class can have method `validate(name, checksum)` to do those checks.

src/io/trailer.h
```diff
@@ -0,0 +1,86 @@
...
+    }
+};
+using TrailerPtr = std::unique_ptr<Trailer>;
+void
+AddTrailerForMemoryIO(MemoryIOWriter& writer, const std::string& name, const Version& version);
+void
+AddTrailerForFiles(const std::vector<std::string>& files, const std::string& trailer_file, const std::string& name,
+                   const Version& version);
+bool
+CheckTrailerForMemoryIO(MemoryIOReader& reader, const std::string& name);
```
zhengbuqian, line 83: I see no usage of CheckTrailer in deserialize methods of memory indexes?

src/io/trailer.cc
```diff
@@ -0,0 +1,156 @@
...
+    }
+    LOG_KNOWHERE_INFO_ << "Index Trailer check succeed.";
+    return true;
+}
+
+void
+AddTrailerForFiles(const std::vector<std::string>& files, const std::string& trailer_file, const std::string& name,
+                   const Version& version) {
+    auto trailer_ptr = std::make_unique<Trailer>();
+    trailer_ptr->SetCheckSum(GetFilesCheckSum(files));
```
zhengbuqian, line 106: trailer for files is missing binary size. I prefer to make them consistent:

* either remove binary size from single binary index and rely solely on checksum for correctness, which should be sufficient?
* or add binary size sum of all files for multiple files index

src/io/memory_io.h
```diff
@@ -71,6 +70,7 @@ getSwappedBytes(char C) {
 
 #endif
 
+// MemoryIOwriter and MemoryIOreader it not thread safe
```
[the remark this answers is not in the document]
cqy123456: updated.

src/io/trailer.h
cqy123456: updated

src/io/trailer.h
cqy123456: updated.

src/io/trailer.h
cqy123456: updated

src/io/trailer.cc
cqy123456: I added flags at the head and tail of the trailer for checking.

src/io/trailer.cc
cqy123456: next pr. This pr verifies old knowhere code also can deserialize with the new binaryset.

src/io/trailer.cc
cqy123456: updated

src/io/trailer.cc
cqy123456: binary size removed.

src/io/trailer.h
cqy123456: next pr. This pr verifies old knowhere code also can deserialize with the new binaryset.

src/io/trailer.h
```diff
@@ -0,0 +1,111 @@
...
+ public:
+    Trailer() {
+        FLAG_TYPE* buf = (FLAG_TYPE*)bytes;
+        buf[BEG_FLAG_OFFEST] = FLAG_MAGIC_NUMBER;
+        buf[END_FLAG_OFFEST] = FLAG_MAGIC_NUMBER;
+    };
+    bool
+    TrailerValidCheck() {
+        FLAG_TYPE* buf = (FLAG_TYPE*)bytes;
+        return buf[BEG_FLAG_OFFEST] == FLAG_MAGIC_NUMBER && buf[END_FLAG_OFFEST] == FLAG_MAGIC_NUMBER;
```
zhengbuqian, line 57: you can check on meta members directly

meta.placeholder == FLAG_MAGIC_NUMBER

src/io/trailer.cc
```diff
@@ -0,0 +1,150 @@
...
+    if (!reader) {
+        LOG_KNOWHERE_WARNING_ << "Trailer file not exist.";
+        return Status::success;
+    }
+    reader.seekg(0, std::ios::end);
+    auto fsize = reader.tellg();
+
+    reader.seekg(0, std::ios::beg);
+    if (fsize != KNOWHERE_TRAILER_SIZE) {
+        LOG_KNOWHERE_ERROR_ << "Trailer size (" << fsize << ")not correct.";
```
zhengbuqian, line 136: should "trailer file exists but has incorrect size" indicate error instead of success?

zhengbuqian: chatted offline, we should consider add a constant `max_version_without_trailer` to the knowhere versioning system, all indexes with a greater version should include trailer, or else error.

src/io/trailer.h
```diff
@@ -0,0 +1,111 @@
...
+#define KNOWHERE_TRAILER_SIZE 512  // bytes
+#define MAX_INDEX_NAME_SIZE 63     // bytes
+#define FLAG_TYPE uint32_t
+#define FLAG_MAGIC_NUMBER UINT32_C(0x1234ABCD)
+#define BEG_FLAG_OFFEST 0
+#define END_FLAG_OFFEST KNOWHERE_TRAILER_SIZE / sizeof(FLAG_TYPE) - 1
+#define TRAILER_OFFSET(size) size - KNOWHERE_TRAILER_SIZE
+
+namespace knowhere {
+struct Meta {
```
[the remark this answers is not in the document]
zhengbuqian: I was wrong, my proposed solution could result in struct member alignment error if we add 8 bytes members

src/io/trailer.cc
```diff
@@ -0,0 +1,150 @@
...
+        while (reader.read((char*)buffer.get(), kBlockSize)) {
+            std::streamsize read_size = reader.gcount();
+            checksum ^= CalculateCheckSum(buffer.get(), read_size);
+        }
+    }
+    return checksum;
+}
+
+knowhere::Status
+TrailerCheck(const knowhere::TrailerPtr& trailer_ptr, const std::string& name, const uint32_t checksum) {
```
zhengbuqian, line 47: consider making this a method of the Trailer class

PR #241, opened 2023-12-01
fix: diskann hang out in release function

cqy123456:


thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -307,6 +307,7 @@ namespace diskann {
       std::string sample_bin, _u64 l_search, _u64 beamwidth,
       _u64 num_nodes_to_cache) {
 #endif
+    this->async_generate_cache.store(true);
```
liliu-z, line 310: If it is `false` during the destruction check, and then do cache gen, crash will occur.

Issue #140, opened 2023-10-11
[Bug] Knowhere can not run on old cpus that only support sse4.2 instruction set


chasingegg:
milvus issue https://github.com/milvus-io/milvus/issues/27552

PR #243, opened 2023-12-01
[2.2] Update folly to resolve simd issue

chasingegg:
issue #140
related milvus issue https://github.com/milvus-io/milvus/issues/27552
milvus fix https://github.com/milvus-io/milvus/pull/28879
/kind bug


chasingegg:
duplicated with #247

Issue #232, opened 2023-11-29
ARM64 compile knowhere library got a compiler error.


Presburger:
use hamming function, but the funtion not find on the struct define, change to compute.

PR #245, opened 2023-12-04
Fix faiss arm compile error in partitioning.cpp

chasingegg:
issue #232 
/kind bug

### DIFF
--- a/thirdparty/faiss/faiss/utils/partitioning.cpp
+++ b/thirdparty/faiss/faiss/utils/partitioning.cpp
@@ -819,7 +819,7 @@ template uint16_t partition_fuzzy<CMax<uint16_t, int>>(
  * Histogram subroutines
  ******************************************************************/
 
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@chasingegg Tbh I'm not sure what exactly this PR fixes, at least from the description and the ref to an issue. Could you please provide more details?

chasingegg:
> @chasingegg Tbh I'm not sure what exactly this PR fixes, at least from the description and the ref to an issue. Could you please provide more details?

We have move simd related code to new file partitioning_avx2.cpp(maybe a better name..), non-simd related code is only used for x86 sse, so the header is proteced by a marco `ifdef __AVX2__`. There are many ways to fix the issue, I try to make the code stay the same with its original version before upgraing faiss, just exclude arm marco and use partitioning_avx2.cpp's implementation for arm arch.

PR #249, opened 2023-12-04
chore(e2e-test): bump build image to amd64-ubuntu20.04-20231117-7a169d0

samhuang-z:


liliu-z:
Should we do it both for the master and 2.2?

PR #264, opened 2023-12-05
Minimize rebuilds of RAFT code due to knowhere changes

### LINKED ISSUE #263: Minimize recompilation of RAFT code on changes to knowhere-only code
Currently because of the way that the `.cu` files are split for RAFT index support, small unrelated changes in code strictly related to knowhere can trigger a full rebuild of RAFT. To help improve development velocity, it would be useful to restructure the code in such a way that RAFT index rebuilds only happen on changes that are strictly RAFT-related.

wphicks:
The primary purpose of this PR is to minimize long rebuilds during development (or in CI if previous builds are cached). We do this by strictly separating the compilation of RAFT components from the compilation of components requiring logic from knowhere headers.

In addition to this separation, this PR updates the RAFT version to a fork of 23.12 containing a critical thread safety fix for `device_resources_manager`. This branch will also be used to bring in changes necessary for integration of a RAFT brute force index prior to the 2.4 release of Milvus.

During development, I noticed that linting had been inconsistently applied to cuda code, resulting in confusingly long diffs for relatively small changes. I have updated the linting here to more consistently apply to cuda code in order to make review of this and future PRs more straightforward.

More detailed comments inline.

Resolve #263


.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@
 #
 
 default_language_version:
-    python: python3.8
```
wphicks, line 19: Made this change because I was developing in an environment where python3.8 was not available. Any problem with having this generic for python3?

.pre-commit-config.yaml
```diff
@@ -32,4 +32,5 @@ repos:
       rev: v1.3.5
       hooks:
         - id: clang-format
+          types_or: [c, c++, cuda]
```
wphicks, line 35: Apply clang-format to all CUDA code.

CMakeLists.txt
```diff
@@ -22,7 +22,9 @@ include(cmake/utils/utils.cmake)
 
 knowhere_option(WITH_RAFT "Build with RAFT indexes" OFF)
 if (WITH_RAFT)
-  set(CMAKE_CUDA_ARCHITECTURES RAPIDS)
+  if("${CMAKE_CUDA_ARCHITECTURES}" STREQUAL "")
```
wphicks, line 25: User should be able to override this if desired. This is particularly useful for setting a value of `NATIVE`, which builds strictly for the local architecture.

CMakeLists.txt
```diff
@@ -139,7 +141,7 @@ list(APPEND KNOWHERE_LINKER_LIBS Folly::folly)
 add_library(knowhere SHARED ${KNOWHERE_SRCS})
 add_dependencies(knowhere ${KNOWHERE_LINKER_LIBS})
 if(WITH_RAFT)
-  list(APPEND KNOWHERE_LINKER_LIBS raft::raft CUDA::cublas CUDA::cusparse CUDA::cusolver)
+  list(APPEND KNOWHERE_LINKER_LIBS raft::raft raft::compiled_static CUDA::cublas CUDA::cusparse CUDA::cusolver)
```
wphicks, line 144: This helps eliminate unintended compilation of RAFT algorithms for types that RAFT does not optimize for. By building the static libraft ourselves, we can be confident that we are only referencing types that RAFT algorithms handle optimally.

cmake/libs/libraft.cmake
```diff
@@ -14,9 +14,10 @@
 # the License.
 
 add_definitions(-DKNOWHERE_WITH_RAFT)
+add_definitions(-DRAFT_EXPLICIT_INSTANTIATE_ONLY)
```
wphicks, line 17: See above. This enforces that we only use instantiations that are specifically optimized for in RAFT.

src/common/raft/integration/cagra_index.cu
```diff
@@ -14,8 +14,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "common/raft/proto/raft_index_kind.hpp"
+#undef RAFT_EXPLICIT_INSTANTIATE_ONLY
 #include "common/raft/integration/raft_knowhere_index.cuh"
+#include "common/raft/proto/filtered_search_instantiation.cuh"
+#include "common/raft/proto/raft_index_kind.hpp"
+
+RAFT_FILTERED_SEARCH_EXTERN(cagra, raft_knowhere::raft_data_t<raft_proto::raft_index_kind::cagra>,
```
wphicks, line 22: Unfortunately due to an unresolved issue in RAFT, filtered searches trigger an extensive rebuild of RAFT components. Here, we declare extern templates for all of the instantiations we wish to use, which allows us to avoid triggering rebuilds unnecessarily.

src/common/raft/integration/cagra_instantiations.cu
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#undef RAFT_EXPLICIT_INSTANTIATE_ONLY
+#include "common/raft/integration/type_mappers.hpp"
+#include "common/raft/proto/filtered_search_instantiation.cuh"
+#include "common/raft/proto/raft_index_kind.hpp"
+
+RAFT_FILTERED_SEARCH_INSTANTIATION(cagra, raft_knowhere::raft_data_t<raft_proto::raft_index_kind::cagra>,
```
wphicks, line 22: Here we actually instantiate the extern templates we use in `cagra_index.cu`. Note that we only include a very limited set of headers here, so changes in knowhere should only rarely trigger a rebuild of the underlying algorithm.

src/common/raft/integration/raft_knowhere_index.hpp
```diff
@@ -18,52 +18,11 @@
 #include <cstdint>
 
 #include "common/raft/integration/raft_knowhere_config.hpp"
+#include "common/raft/integration/type_mappers.hpp"
 #include "common/raft/proto/raft_index_kind.hpp"
 
 namespace raft_knowhere {
 
-using knowhere_data_type = float;
```
wphicks, line 25: We move these to their own header because we need them in the extern template instantiations, and we do not want to have to recompile the extern templates every time we change the connector code between RAFT and knowhere.

src/common/raft/proto/raft_index.cuh
```diff
@@ -20,377 +20,230 @@
 #include <istream>
 #include <optional>
 #include <ostream>
-#include <type_traits>
```
wphicks, line 23: I believe all changes in this file are the result of applying the linters to CUDA code.

PR #280, opened 2023-12-08
Move memory allocation out of elkan_L2_sse

### LINKED ISSUE #279: enhance: Reduce memory malloc for elkan data train
elkan_L2_sse is called by the loop in train_encoded, that means it will repeatly do memory malloc/free. For memory allocation will trigger multiple behaviors in kernel side, such as vma allocation, physical page allocation, page mapping setup and memory cgroup statistic update, it's unecessary to repeately take the overhead.

bjzhjing:
elkan_L2_sse() does malloc/free each time when it's called by the loop in train_encoded() during data train for most of IVF index building. Memory allocation triggers the process of vma allocation, physical page allocation, page mapping setup and memory cgroup statistic update in kernel side, so more frequent allocation means more overhead. Move it out of the loop and do it only once when IndexFlatElkan is initialized to save CPU cycles.

issue #279

### DIFF
--- a/thirdparty/faiss/faiss/IndexFlatElkan.cpp
+++ b/thirdparty/faiss/faiss/IndexFlatElkan.cpp
@@ -21,6 +21,8 @@ namespace faiss {
 IndexFlatElkan::IndexFlatElkan(idx_t d, MetricType metric, bool is_cosine, bool use_elkan)
         : IndexFlat(d, metric, is_cosine) {
     this->use_elkan = use_elkan;
+    if (this->use_elkan)
+        this->tmp_buffer_for_elkan = std::make_unique<float[]>(1024 * (1024 - 1) / 2);
 }
 
 void IndexFlatElkan::search(
@@ -54,7 +56,7 @@ void IndexFlatElkan::search(
             // ignore the metric_type, both use L2
             if (use_elkan) {
                 // use elkan
-                elkan_L2_sse(x, get_xb(), d, n, ntotal, labels, dis_inner);
+                elkan_L2_sse(x, get_xb(), d, n, ntotal, labels, dis_inner, tmp_buffer_for_elkan.get());
             }
             else {
                 // use L2 search. The same code as in IndexFlat::search() for L2.
--- a/thirdparty/faiss/faiss/IndexFlatElkan.cpp
+++ b/thirdparty/faiss/faiss/IndexFlatElkan.cpp
@@ -21,8 +21,10 @@ namespace faiss {
 IndexFlatElkan::IndexFlatElkan(idx_t d, MetricType metric, bool is_cosine, bool use_elkan)
         : IndexFlat(d, metric, is_cosine) {
     this->use_elkan = use_elkan;
-    if (this->use_elkan)
-        this->tmp_buffer_for_elkan = std::make_unique<float[]>(1024 * (1024 - 1) / 2);
+    if (this->use_elkan) {
+        this->sym_dim = 1024;
+        this->tmp_buffer_for_elkan = std::make_unique<float[]>(this->sym_dim * (this->sym_dim - 1) / 2);
+    }
 }
 
 void IndexFlatElkan::search(
@@ -56,7 +58,7 @@ void IndexFlatElkan::search(
             // ignore the metric_type, both use L2
             if (use_elkan) {
                 // use elkan
-                elkan_L2_sse(x, get_xb(), d, n, ntotal, labels, dis_inner, tmp_buffer_for_elkan.get());
+                elkan_L2_sse(x, get_xb(), d, n, ntotal, labels, dis_inner, tmp_buffer_for_elkan.get(), sym_dim);
             }
             else {
                 // use L2 search. The same code as in IndexFlat::search() for L2.
--- a/thirdparty/faiss/faiss/IndexFlatElkan.h
+++ b/thirdparty/faiss/faiss/IndexFlatElkan.h
@@ -36,8 +36,13 @@ namespace faiss {
 // 'IndexFlatElkan::search()' on the same instance should be avoided
 // to prevent data overwite corruption in the temporary buffer.
 //
+// The temporary buffer is used to store half size of symmetric matrix
+// data, excluding the diagnoal. IndexFlatElkan specifies the dimension
+// of the matrix.
+//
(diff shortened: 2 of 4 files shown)


thirdparty/faiss/faiss/utils/distances.h
```diff
@@ -422,6 +422,7 @@ void compute_PQ_dis_tables_dsub2(
  * @param y    database vectors, size ny * d
  * @param ids  result array ids
  * @param val  result array value
+ * @param data the pointer of memory for symmetric matrix data
```
alexanderguzhva, line 425: Please call this variable `tmp_buffer` in order to stress out its meaning. It is not a data, it is a temporary storage.

thirdparty/faiss/faiss/IndexFlatElkan.cpp
```diff
@@ -15,12 +15,15 @@
 
 #include <faiss/impl/FaissAssert.h>
 #include <faiss/utils/distances.h>
+#include <faiss/utils/utils.h>
 
 namespace faiss {
 
-IndexFlatElkan::IndexFlatElkan(idx_t d, MetricType metric, bool is_cosine, bool use_elkan)
+IndexFlatElkan::IndexFlatElkan(idx_t d, MetricType metric, bool is_cosine, bool use_elkan, float* data)
```
alexanderguzhva, line 22: ?! `data` which is passed as a parameters is not used anywhere

thirdparty/faiss/faiss/IndexFlatElkan.h
```diff
@@ -29,9 +29,17 @@ namespace faiss {
 // Elkan algo was introduced into Knowhere in #2178, #2180 and #2258. 
 struct IndexFlatElkan : IndexFlat {
     bool use_elkan = true;
+    float* data = nullptr;
```
alexanderguzhva, line 32: `std::vector<float> tmp_buffer_for_elkan;`

thirdparty/faiss/faiss/IndexFlatElkan.h
```diff
@@ -29,9 +29,17 @@ namespace faiss {
...
 
     explicit IndexFlatElkan(idx_t d, MetricType metric = METRIC_L2,
-                       bool is_cosine = false, bool use_elkan = true);
+                       bool is_cosine = false, bool use_elkan = true, float* data = nullptr);
```
alexanderguzhva, line 35: I see no point in passing `float* data` to constructor. By design, it is a temporary buffer

thirdparty/faiss/faiss/utils/distances.cpp
```diff
@@ -1155,13 +1155,18 @@ void elkan_L2_sse(
...
-        float* val) {
+        float* val,
+        float* data) {
     if (nx == 0 || ny == 0) {
         return;
     }
 
     const size_t bs_y = 1024;
-    float* data = (float*)malloc((bs_y * (bs_y - 1) / 2) * sizeof(float));
+    bool allocate = false;
```
alexanderguzhva, line 1165: Please use `std::unique_ptr<float[]>` + `new` for allocating a local temporary buffer

thirdparty/faiss/faiss/utils/distances.h
bjzhjing: Thanks for helping review, Alex! The change is done.

thirdparty/faiss/faiss/IndexFlatElkan.cpp
bjzhjing: Yeah, it's not required to be set by callers. Remove it from the interface.

thirdparty/faiss/faiss/IndexFlatElkan.h
bjzhjing: Changed it.

thirdparty/faiss/faiss/IndexFlatElkan.h
bjzhjing: Yes, remove it.

thirdparty/faiss/faiss/utils/distances.cpp
bjzhjing: Call resize() to allocate the memory in constructor. Remove the checking here for use_elkan can help ensure the memory is allocated when come here.

bjzhjing:
@alexanderguzhva Hi Alex, I use smart pointer instead of std::vector to implement the tmp_buffer finally, for I IndexFlatElkan::search is defined with 'const', which means it's not allowed to modify the class variables. Thus, it's required to use const_cast<std::vector<float>&> for convert while passing tmp_buffer to elkan_L2_sse, there is address change, which make things a bit complex. Smart pointer can help avoid the issue and make the code more clean and safe.

Please help review. Thanks very much!

alexanderguzhva:
@bjzhjing Two more notes
1. Please add a comment about the situation to the top of `IndexFlatElkan` . Basically, in the provided form `IndexFlatElkan` is no longer thread-safe for the search. So, two threads cannot use 'IndexFlatElkan::search()' simultaneosly on the same instance. This is fine, bcz `IndexFlatElkan` is used only during the index construction. But the note about the implementation detail and why this was done should be included.
2. The needed size of the temporary buffer depends on the `const size_t bs_y = 1024;` magic constant inside  `elkan_L2_sse()`. So, notes on the sizes need to be added to the description of the `elkan_L2_sse()` implementation. Maybe, `bs_y` size needs to be exposed as a function parameter as well.

bjzhjing:
> @bjzhjing Two more notes
> 
> 1. Please add a comment about the situation to the top of `IndexFlatElkan` . Basically, in the provided form `IndexFlatElkan` is no longer thread-safe for the search. So, two threads cannot use 'IndexFlatElkan::search()' simultaneosly on the same instance. This is fine, bcz `IndexFlatElkan` is used only during the index construction. But the note about the implementation detail and why this was done should be included.
> 2. The needed size of the temporary buffer depends on the `const size_t bs_y = 1024;` magic constant inside  `elkan_L2_sse()`. So, notes on the sizes need to be added to the description of the `elkan_L2_sse()` implementation. Maybe, `bs_y` size needs to be exposed as a function parameter as well.

Greetings @alexanderguzhva! I'm just back from a bad cold, and sorry for the late response. I've addressed your above comments. Please help review, thanks!

Issue #227, opened 2023-11-28
RAFT should fill result with id = -1


cydrain:
When doing search and there are no enough results data, knowhere should fill in dummy results with id = -1.
But in RAFT, dummy results are filled with id = 0.

![Screenshot from 2023-11-28 16-38-02](https://github.com/zilliztech/knowhere/assets/3992404/72b67028-2134-476a-9c45-ca0c8ac7c0b7)

test case: test_func.TestFunc.test_gpu_index

cydrain:
@Presburger 
please use case "test_func.TestFunc.test_gpu_index" to verify, still see result ids filled with 0

![Screenshot from 2023-12-07 18-40-57](https://github.com/zilliztech/knowhere/assets/3992404/d4238a00-5292-42fc-b3f5-be56601cda08)

cydrain:
test with knowhere (commit_id = "edcc6c718"), GPU_IVF_FLAT 's id is filled with -1, but GPU_IVF_PQ 's id is filled with some strange values:

![Screenshot from 2023-12-11 15-10-23](https://github.com/zilliztech/knowhere/assets/3992404/cd5817f7-2377-43b1-b987-46dc966600b7)

PR #286, opened 2023-12-12
Add support for GPU-accelerated brute force indexes

wphicks:
This PR provides GPU-accelerated exact search using RAFT's brute force index. As part of the necessary work to implement post-filtering for brute force, I have also updated the logic for handling missing values. Rather than trying to handle the various missing values resulting from https://github.com/rapidsai/raft/issues/1822, we check both the distance and ID, replacing invalid IDs or IDs with associated infinite distance with `-1`. This may address #227, but if reviewers could double-check my logic, it would be much appreciated.

/issue: #288


wphicks (commented):
Just added comments explaining changes likely to raise questions.

Presburger (approved):

Presburger (approved):

PR #292, opened 2023-12-12
Allow RAFT indexes to perform refinement after serialization

### LINKED ISSUE #289: [BUG] Deserialized RAFT indexes cannot perform refinement
After being serialized and deserialized, RAFT indexes switch to a different representation of their dataset, causing it to be unavailable for use in refinement when `refine_ratio` is set greater than 1. To fix this, we should serialize out the dataset separately and restore it separately.

I will work on this as a follow-up to #286.

wphicks:
Because RAFT requires access to the dataset itself for refinement requests (and for the upcoming fallback to brute force at large k), serialization of RAFT-based indexes should include the dataset if it is available. This PR ensures that RAFT indexes can still perform refinement after a serialization-deserialization round-trip.

/issue #289

### DIFF
--- a/src/common/raft/integration/raft_knowhere_index.cuh
+++ b/src/common/raft/integration/raft_knowhere_index.cuh
@@ -26,6 +26,7 @@
 #include <raft/core/device_setter.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/resource/thrust_policy.hpp>
+#include <raft/core/serialize.hpp>
 #include <raft/distance/distance_types.hpp>
 #include <raft/neighbors/ivf_pq_types.hpp>
 #include <raft/neighbors/sample_filter.cuh>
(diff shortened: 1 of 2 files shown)


src/common/raft/integration/raft_knowhere_index.cuh
```diff
@@ -454,40 +486,42 @@ struct raft_knowhere_index<IndexKind>::impl {
...
         } else {
             raft::copy(res, host_ids, device_ids);
+            raft::copy(res, host_distances, device_distances);
+        }
+        raft::resource::sync_stream(res);
+        auto static max_distance = std::nextafter(std::numeric_limits<data_type>::max(), 0.0f);
+        for (auto i = knowhere_indexing_type{}; i < row_count; ++i) {
+            for (auto j = knowhere_indexing_type{}; j < k; ++j) {
+                auto distance = host_distances(i, j);
+                if (distance >= max_distance || distance < 0 ||
```
Presburger, line 519: Will handling this on the host side result in poorer performance compared to GPU, especially in large-scale K?

src/common/raft/integration/raft_knowhere_index.cuh
```diff
@@ -423,12 +445,22 @@ struct raft_knowhere_index<IndexKind>::impl {
...
             std::optional<raft::core::bitset<knowhere_bitset_data_type, knowhere_bitset_indexing_type>>{};
+        auto k_tmp = k;
 
         if (bitset_data != nullptr && bitset_byte_size != 0) {
             device_bitset =
                 raft::core::bitset<knowhere_bitset_data_type, knowhere_bitset_indexing_type>(res, bitset_size);
             raft::copy(res, device_bitset->to_mdspan(), raft::make_host_vector_view(bitset_data, bitset_byte_size));
-            device_bitset->flip(res);
+            if constexpr (index_kind == raft_proto::raft_index_kind::brute_force) {
+                k_tmp += device_bitset->count(res);
```
Presburger, line 455: It's truly a brute force approach.🤣

src/common/raft/integration/raft_knowhere_index.cuh
wphicks: That's a very good question and something I mentioned in [this comment](https://github.com/zilliztech/knowhere/pull/286#discussion_r1424309842) on the PR that is really responsible for this change. The short answer is that it really depends for both batch size and k. Sometimes, the device allocation and copy is more expensive than the host-side compute; sometimes the host-side compute is more expensive.

I'm actually fine with doing it either way in the initial version. I don't think we'll have really good guidance until we do more thorough benchmarking. Happy to change it in #286 if you'd rather be conservative there.

src/common/raft/integration/raft_knowhere_index.cuh
wphicks: Haha, yeah. If this is overkill, we can cap the maximum k at some reasonable value. This was just a way to make very, very sure that brute force returns an exact result via postfiltering until we get prefiltering in.

src/common/raft/integration/raft_knowhere_index.cuh
wphicks: I just took a closer look at my initial perf testing and realizing that this has more of an impact than I thought. I've just updated #286 with [this change](https://github.com/zilliztech/knowhere/pull/286/commits/d3f459f432d1cf80b96248e3c6ef588e409b8968) that allows us to avoid the extra allocation when it's not necessary, and that gets us much better performance all the way around.

@Presburger would you mind taking a look at that change, and if you like it, I'll cherry-pick it onto #292 and #296?

Presburger: @wphicks I prefer substitution on the device side. Comparatively, there are more performance factors on the host side.

wphicks: Excellent! I'll rebase to pull in those changes. Thanks very much!

Presburger (approved):

cqy123456 (approved):

PR #294, opened 2023-12-13
Remove useless prometheus metrics

cydrain:


chasingegg (commented):
why remove search_topk.. it is useful to see if knowhere'search topk vs milvus search topk are the same

Issue #287, opened 2023-12-12
Knowhere support multi datatype


cqy123456:
Tasks：
- [x] knowhere interface support mutli data type
- [x] pyknowhere support mutli data type
- [x] Diskann
- [ ] HNSW (doing)
- [ ] Add distance function for fp16/bf16

PR #306, opened 2023-12-14
Remove args that xargs can not recognized

cqy123456:
issue: https://github.com/zilliztech/knowhere/issues/287
/kind improvement


chasingegg (approved):
/lgtm

PR #319, opened 2023-12-26
Optimize thread pool initialization

cydrain:


liliu-z:
Any reason we need to make this change?

PR #322, opened 2024-01-02
Shuffle ids for building graph index

chasingegg:
This pr is to shuffle ids for building graph index since user's data may be clustered by its sequence, this could cause some islands in graph to affect recall significantly, shuffle could alleviate this.
Divide base **vector ids**(not raw data itself) into different blocks(diskann already did this), and shuffle these blocks to build index, and adding points in each block is still memory friendly, result shows on a dataset with clustered data. Try to use 8192 as default block size in hnsw, which is currently used in diskann, to achieve almost the same building time and high recall in these cases.

hnsw | Build time(s) | recall(top100) | diskann | Build time(s) | recall(top100)
-- | -- | -- | -- | -- | --
Full shuffle | 246 | 0.926 |   | 479 | 0.933
Block size=1024 | 147 | 0.9079 |   | 389 | 0.9203
Block size=8192 | 126 | 0.8988 |   | 376 | 0.8667
No shuffle(now) | 117 | 0.4975 |   | 371 | 0.4752

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -519,6 +519,7 @@ class BaseConfig : public Config {
     CFG_BOOL trace_visit;
     CFG_BOOL enable_mmap;
     CFG_BOOL for_tuning;
+    CFG_BOOL shuffle_build;
     KNOHWERE_DECLARE_CONFIG(BaseConfig) {
         KNOWHERE_CONFIG_DECLARE_FIELD(metric_type)
             .set_default("L2")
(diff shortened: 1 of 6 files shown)


alexanderguzhva:
@chasingegg 
1) what problem does this commit solve exactly and why the problem is actually solved? 
2) is it worth making such a shuffle configurable in order to have 'more reproducible' results, for example, during some kind of debugging? 

Thanks

liliu-z:
Still need to know why order plus OOD data can damage the recall.

foxspy (approved):

foxspy (approved):
/lgtm

chasingegg:
Fix #326

PR #330, opened 2024-01-05
Make storage of dataset on device for refinement optional

### LINKED ISSUE #329: Reduce device memory consumption for RAFT when refinement is not required
Currently, a copy of the dataset is held on device to facilitate low-latency refinement of results returned by a base index for RAFT index types. This should be optional to allow users who will not be performing refinement to fit larger datasets on GPUs with less device memory.

wphicks:
This can significantly reduce device memory consumption in cases where refinement is not desired.

Also ensure that RAFT indexes compile and work correctly for compute capability 8.9 devices.

Close #329


wphicks:
Noticed that the current changes are not correct because of how vector-adding is inconsistently handled upstream. Working on a more complete fix, but it will not be quite as simple as this.

Presburger (approved):

PR #341, opened 2024-01-10
[feat] Add sparse index support to knowhere

zhengbuqian:
/kind feature
https://github.com/zilliztech/knowhere/issues/193

This PR adds the basic framework of supporting sparse index in knowhere, with 2 index types: inverted index and wand.

In this PR only Search and GetVectorByIds is supported.

Added basic unit tests:

Search without filtering, with different dataset sizes/recall/sparisity, etc.
Search with filtering, with different filter rate.

### DIFF
--- a/include/knowhere/sparse_utils.h
+++ b/include/knowhere/sparse_utils.h
@@ -1,0 +1,261 @@
+// Copyright (C) 2019-2023 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// valributed under the License is valributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <type_traits>
+#include <vector>
+
+#include "knowhere/operands.h"
+
+namespace knowhere::sparse {
+
+// integer type in SparseRow
+using table_t = uint32_t;
+// type used to represent the id of a vector in the index interface.
+// this is same as other index types.
+using label_t = int64_t;
+
+template <typename T>
+struct IdVal {
+    table_t id;
+    T val;
+
+    IdVal() = default;
+    IdVal(table_t id, T val) : id(id), val(val) {
+    }
+
+    inline friend bool
+    operator<(const IdVal& lhs, const IdVal& rhs) {
+        return lhs.val < rhs.val || (lhs.val == rhs.val && lhs.id < rhs.id);
+    }
+    inline friend bool
+    operator>(const IdVal& lhs, const IdVal& rhs) {
+        return !(lhs < rhs);
+    }
+
+    inline friend bool
+    operator==(const IdVal& lhs, const IdVal& rhs) {
+        return lhs.id == rhs.id && lhs.val == rhs.val;
+    }
+};
+
+template <typename T>
+class SparseRow {
+    static_assert(std::is_same_v<T, fp32>, "SparseRow supports float only");
+
+ public:
+    // construct an SparseRow with memory allocated to hold `count` elements.
+    SparseRow(size_t count = 0)
+        : data_(count ? new uint8_t[count * element_size()] : nullptr), count_(count), own_data_(true) {
+    }
+
+    SparseRow(size_t count, uint8_t* data, bool own_data) : data_(data), count_(count), own_data_(own_data) {
+    }
+
+    // copy constructor and copy assignment operator perform deep copy
+    SparseRow(const SparseRow<T>& other) : SparseRow(other.count_) {
+        std::copy(other.data_, other.data_ + count_ * element_size(), data_);
+    }
+
+    SparseRow(SparseRow<T>&& other) noexcept : SparseRow() {
+        swap(*this, other);
+    }
+
+    SparseRow&
+    operator=(const SparseRow<T>& other) {
+        if (this != &other) {
+            SparseRow<T> tmp(other);
+            swap(*this, tmp);
+        }
+        return *this;
+    }
+
+    SparseRow&
+    operator=(SparseRow<T>&& other) noexcept {
+        swap(*this, other);
+        return *this;
+    }
+
+    ~SparseRow() {
+        if (own_data_ && data_ != nullptr) {
+            delete[] data_;
+            data_ = nullptr;
+        }
+    }
+
+    size_t
+    size() const {
+        return count_;
+    }
+
+    size_t
+    memory_usage() const {
+        return count_ * element_size() + sizeof(*this);
+    }
+
+    void*
+    data() {
+        return data_;
+    }
+
+    const void*
+    data() const {
+        return data_;
+    }
+
+    // dim of a sparse vector is the max index + 1, or 0 for an empty vector.
+    int64_t
+    dim() const {
+        if (count_ == 0) {
+            return 0;
+        }
+        auto* elem = reinterpret_cast<const ElementProxy*>(data_) + count_ - 1;
+        return elem->index + 1;
+    }
+
+    IdVal<T>
+    operator[](size_t i) const {
+        auto* elem = reinterpret_cast<const ElementProxy*>(data_) + i;
+        return {elem->index, elem->value};
+    }
+
+    void
+    set_at(size_t i, table_t index, T value) {
+        auto* elem = reinterpret_cast<ElementProxy*>(data_) + i;
+        elem->index = index;
+        elem->value = value;
+    }
+
+    float
+    dot(const SparseRow<T>& other) const {
+        float product_sum = 0.0f;
+        size_t i = 0, j = 0;
+        while (i < count_ && j < other.count_) {
+            auto* left = reinterpret_cast<const ElementProxy*>(data_) + i;
+            auto* right = reinterpret_cast<const ElementProxy*>(other.data_) + j;
+
+            if (left->index < right->index) {
+                ++i;
+            } else if (left->index > right->index) {
+                ++j;
+            } else {
+                product_sum += left->value * right->value;
+                ++i;
+                ++j;
+            }
+        }
+        return product_sum;
+    }
+
+    friend void
+    swap(SparseRow<T>& left, SparseRow<T>& right) {
+        using std::swap;
+        swap(left.count_, right.count_);
+        swap(left.data_, right.data_);
+        swap(left.own_data_, right.own_data_);
+    }
+
+    static inline size_t
+    element_size() {
+        return sizeof(table_t) + sizeof(T);
+    }
+
+ private:
+    // ElementProxy is used to access elements in the data_ array and should
+    // never be actually constructed.
+    struct __attribute__((packed)) ElementProxy {
+        table_t index;
+        T value;
+        ElementProxy() = delete;
+        ElementProxy(const ElementProxy&) = delete;
+    };
+    // data_ must be sorted by column id. use raw pointer for easy mmap and zero
+    // copy.
+    uint8_t* data_;
+    size_t count_;
+    bool own_data_;
+};
+
+// When pushing new elements into a MaxMinHeap, only `capacity` elements with the
+// largest val are kept. pop()/top() returns the smallest element out of them.
+template <typename T>
+class MaxMinHeap {
+ public:
+    explicit MaxMinHeap(int capacity) : capacity_(capacity), pool_(capacity) {
+    }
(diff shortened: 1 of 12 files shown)


src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,444 @@
...
+            if (use_wand_) {
+                max_in_dim_.resize(dim);
+            }
+        }
+
+        raw_data_.resize(current_rows + rows);
+        for (size_t i = 0; i < rows; ++i) {
+            raw_data_[current_rows + i] = data[i];
+            for (auto [idx, val] : data[i].data) {
+                if (val < val_threshold) {
```
alexanderguzhva, line 147: if `drop_ratio_build_ == 0`, then `val_threshold` is `0`, so negative `val` values will be skipped. Is it by design? This implies that sparse values are expected to be positive only

zhengbuqian: Thanks Alex! I have changed it so that now we ignore only values that are close enough to zero.

Also the drop strategy has been modified from "drop any value that is below ratio * global max" to "drop the smallest ratio% abs values".

zhengbuqian:
@alexanderguzhva @hhy3 @liliu-z PTAL

include/knowhere/sparse_utils.h
```diff
@@ -0,0 +1,226 @@
...
+ public:
+    SparseRow() = default;
+
+    SparseRow(const SparseRow& other) {
+        copy_from(other);
+    }
+
+    SparseRow&
+    operator=(const SparseRow& other) {
+        copy_from(other);
```
alexanderguzhva, line 68: Please do all these nasty C++ things like `if (this == &other)` here

include/knowhere/sparse_utils.h
```diff
@@ -0,0 +1,226 @@
...
+        return *this;
+    }
+
+    ~SparseRow() {
+        if (own_data && data != nullptr) {
+            delete[] data;
+        }
+    }
+
+    size_t
```
alexanderguzhva, line 78: move constructor and move assignment operator please :)

include/knowhere/sparse_utils.h
```diff
@@ -0,0 +1,226 @@
...
+    int64_t
+    dim() const {
+        if (count == 0)
+            return 0;
+        return (*this)[count - 1].id + 1;
+    }
+
+    IdVal<T>
+    operator[](size_t i) const {
+        table_t index = *(table_t*)(data + i * (sizeof(table_t) + sizeof(T)));
```
alexanderguzhva, line 103: all these raw pointers operations are a nightmare. Please introduce a helper struct, such as 
```
template<typename T>
struct SparseRowElement {
   table_t idx;
   T value;
};
```
make sure that it is packed and just use `reinterpret_cast<SparseRowElement<T>*>(data)` or `reinterpret_cast<const ...*>(data)` for working .

Alternatively, just use `IdVal<T>` but make sure that it is packed.

include/knowhere/sparse_utils.h
```diff
@@ -0,0 +1,226 @@
...
+    // data must be sorted by column id. use raw pointer for easy mmap and zero
+    // copy.
+    uint8_t* data = nullptr;
+    size_t count = 0;
+    bool own_data = false;
+};
+
+// When pushing new elements into a MaxMinHeap, only `capacity` elements with the
+// largest val are kept. pop()/top() returns the smallest element out of them.
+template <typename T>
```
alexanderguzhva, line 162: Don't we have any similar structures in our code already? I'm just curious, the implementation is fine

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,457 @@
...
+
+    // Non zero drop ratio is only supported for static index, i.e. data should
+    // include all rows that'll be added to the index.
+    Status
+    Train(const SparseRow<T>* data, size_t rows, float drop_ratio_build) {
+        if (drop_ratio_build == 0.0f) {
+            return Status::success;
+        }
+
+        std::vector<T> vals;
```
alexanderguzhva, line 113: just compute the size that needs to be preallocated, such as
```
size_t amount = 0;
for (size_t i = 0; i < rows; i++) { amount += data[i].size(); }

std::vector<T> vals(amount);
```

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,457 @@
...
+    // any value in q_vec that is smaller than q_threshold and any value with dimension >= n_cols() will be ignored.
+    void
+    search_brute_force(const SparseRow<T>& q_vec, T q_threshold, MaxMinHeap<T>& heap, const BitsetView& bitset) const {
+        std::vector<float> scores(n_rows_internal(), 0.0f);
+        for (size_t idx = 0; idx < q_vec.size(); ++idx) {
+            auto [i, v] = q_vec[idx];
+            if (v < q_threshold || i >= n_cols_internal()) {
+                continue;
+            }
+            for (size_t j = 0; j < inverted_lut_[i].size(); j++) {
```
alexanderguzhva, line 249: this loop can be improved via SIMD. Maybe, later

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,457 @@
...
+        for (size_t i = 0; i < n_rows_internal(); ++i) {
+            if ((bitset.empty() || !bitset.test(i)) && scores[i] != 0) {
+                heap.push(i, scores[i]);
+            }
+        }
+    }
+
+    // LUT supports size() and operator[] which returns an IdVal.
+    template <typename LUT>
+    class Cursor {
```
alexanderguzhva, line 263: the design of the iterator is somewhat unusual, but it is fine

src/index/sparse/sparse_inverted_index.h
```diff
@@ -0,0 +1,457 @@
...
+            raw_data_[current_rows + i].copy_from(data[i]);
+            add_row_to_index(data[i], current_rows + i);
+        }
+        return Status::success;
+    }
+
+    void
+    Search(const SparseRow<T>& query, size_t k, float drop_ratio_search, float* distances, label_t* labels,
+           size_t refine_factor, const BitsetView& bitset) const {
+        std::shared_lock<std::shared_mutex> lock(mu_);
```
alexanderguzhva, line 157: nit: you could move the lock of `mu_` a little bit down further, because in the following lines you do nothing but `std::fill` and not using any of the fields of this struct

include/knowhere/sparse_utils.h
zhengbuqian: Done

include/knowhere/sparse_utils.h
zhengbuqian: Done

include/knowhere/sparse_utils.h
zhengbuqian: Done, added a private structure.

src/index/sparse/sparse_inverted_index.h
zhengbuqian: Done

src/index/sparse/sparse_inverted_index.h
zhengbuqian: great suggestion! I'll leave that as is for now but I left a TODO.

src/index/sparse/sparse_inverted_index.h
zhengbuqian: Done

include/knowhere/sparse_utils.h
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+ private:
+    // ElementProxy is used to access elements in the data_ array and should
+    // never be actually constructed.
+    struct __attribute__((packed)) ElementProxy {
+        table_t index;
+        T value;
+        ElementProxy() = delete;
+        ElementProxy(const ElementProxy&) = delete;
```
alexanderguzhva, line 186: I would add `ElementProxy& operator=(const ElementProxy&) = delete;` as well

src/common/comp/brute_force.cc
```diff
@@ -362,6 +363,94 @@ BruteForce::RangeSearch(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+                cur_distances[j] = heap.top().val;
+                heap.pop();
+            }
+            return Status::success;
+        }));
+    }
+    for (auto& fut : futs) {
+        fut.wait();
+        RETURN_IF_ERROR(fut.result().value());
+    }
```
alexanderguzhva, line 433: I'd use the following to make sure that no futs get unfinished
```
for (auto& fut : futs) fut.wait();
for (auto& fut : futs) RETURN_IF_ERROR(fut.result().value());
```

src/common/comp/brute_force.cc
```diff
@@ -362,6 +363,94 @@ BruteForce::RangeSearch(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+                         const BitsetView& bitset) {
+    auto nq = query_dataset->GetRows();
+    BruteForceConfig cfg;
+    std::string msg;
+    auto status = Config::Load(cfg, config, knowhere::SEARCH, &msg);
+    if (status != Status::success) {
+        return expected<DataSetPtr>::Err(status, msg);
+    }
+    int topk = cfg.k.value();
+    auto labels = new sparse::label_t[nq * topk];
```
alexanderguzhva, line 448: A potential memory leak.  I'd use something like
```
auto labels = std::make_unique<sparse::label_t>(nq * topk);
auto distances = std::make_unique<float>(nq * topk);

SearchSparseWithBuf(base_dataset, query_dataset, labels.get(), distances.get(), config, bitset);
return GenResultDataSet(nq, topk, labels.release(), distances.release());
```

If we're using C++ 17, of course

src/index/sparse/sparse_index_node.cc
```diff
@@ -0,0 +1,219 @@
...
+            return expected<DataSetPtr>::Err(Status::empty_index, "index not loaded");
+        }
+        auto cfg = static_cast<const SparseInvertedIndexConfig&>(config);
+        auto nq = dataset.GetRows();
+        auto queries = static_cast<const sparse::SparseRow<T>*>(dataset.GetTensor());
+        auto k = cfg.k.value();
+        auto refine_factor = cfg.refine_factor.value_or(10);
+        auto drop_ratio_search = cfg.drop_ratio_search.value_or(0.0f);
+
+        auto p_id = new sparse::label_t[nq * k];
```
alexanderguzhva, line 87: same for using `std::make_unique<>` as above

src/index/sparse/sparse_index_node.cc
```diff
@@ -0,0 +1,219 @@
...
+        for (int64_t idx = 0; idx < nq; ++idx) {
+            futs.emplace_back(search_pool_->push([&, idx = idx]() {
+                index_->Search(queries[idx], k, drop_ratio_search, p_dist + idx * k, p_id + idx * k, refine_factor,
+                               bitset);
+            }));
+        }
+        for (auto& fut : futs) {
+            fut.wait();
+        }
+        return GenResultDataSet(nq, k, p_id, p_dist);
```
alexanderguzhva, line 101: `for (auto& fut : futs) RETURN_IF_ERROR(fut.result().value());` is missing

src/common/comp/brute_force.cc
zhengbuqian: that is correct. Considering the lambda does not return a Status other than success, I removed the return value and use `fut.get()` instead.

src/index/sparse/sparse_index_node.cc
zhengbuqian: changed to `fut.get()`

include/knowhere/sparse_utils.h
zhengbuqian: deleting the default and the copy constructor(so no constructor will be generated) should be sufficient. if no instance can ever be made there is no way to call copy assignment operator.

the reason is this element is used solely to help access what's in the raw `data_` pointer, we don't actually have the need to create such objects.

src/common/comp/brute_force.cc
zhengbuqian: good catch!

src/index/sparse/sparse_index_node.cc
zhengbuqian: Done

src/common/comp/brute_force.cc
zhengbuqian: there are actually other places that made the same mistake, creating another PR to fix

zhengbuqian: see https://github.com/zilliztech/knowhere/pull/355

src/index/sparse/sparse_index_node.cc
alexanderguzhva: I've added comments to #355 . Basically, it does not work as desired.

alexanderguzhva:
/hold
#355-related comments need to be addressed first
but looks good overall

alexanderguzhva:
/unhold
folly's `collectAll()` works

alexanderguzhva:
Please change for using `collectAll()`  and I'll lgtm the diff
Thanks

PR #355, opened 2024-01-19
make sure we rethrow exceptions in async tasks

zhengbuqian:
in many places where folly::Future is used, we only wait but not try to get the value. if the async task thrown an exception wait() will unblock but not rethrow it, causing the async thread to crash. 

now using folly::collect().get() to make sure we always wait for all futures and rethrow exceptions so callers may catch and handle them.

/kind improvement

### DIFF
--- a/src/index/diskann/diskann.cc
+++ b/src/index/diskann/diskann.cc
@@ -538,17 +538,17 @@ DiskANNIndexNode<DataType>::Search(const DataSet& dataset, const Config& cfg, co
                                                  search_conf.search_list_size.value());
     }
 
-    auto p_id = new int64_t[k * nq];
-    auto p_dist = new DistType[k * nq];
+    auto p_id = std::make_unique<int64_t[]>(k * nq);
+    auto p_dist = std::make_unique<DistType[]>(k * nq);
 
     std::vector<folly::Future<folly::Unit>> futures;
     futures.reserve(nq);
     for (int64_t row = 0; row < nq; ++row) {
-        futures.emplace_back(search_pool_->push([&, index = row]() {
+        futures.emplace_back(search_pool_->push([&, index = row, p_id_ptr = p_id.get(), p_dist_ptr = p_dist.get()]() {
             diskann::QueryStats stats;
-            pq_flash_index_->cached_beam_search(xq + (index * dim), k, lsearch, p_id + (index * k),
-                                                p_dist + (index * k), beamwidth, false, &stats, feder_result, bitset,
-                                                filter_ratio, for_tuning);
+            pq_flash_index_->cached_beam_search(xq + (index * dim), k, lsearch, p_id_ptr + (index * k),
+                                                p_dist_ptr + (index * k), beamwidth, false, &stats, feder_result,
+                                                bitset, filter_ratio, for_tuning);
 #ifdef NOT_COMPILE_FOR_SWIG
             knowhere_diskann_search_hops.Observe(stats.n_hops);
 #endif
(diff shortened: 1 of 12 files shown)


alexanderguzhva:
No.

Basically, we want to wait for all the futures to finish and only then to process the exception. Alternatively, we could have something that stops further futures from getting spawned if one of them experienced a problem, but this is not really needed.

I've tried the following snippet:
```
int main() {
  auto pool = ThreadPool::GetGlobalSearchThreadPool(); // say, it can run 4 threads concurrently
  std::vector<folly::Future<folly::Unit>> futures;

  for (size_t i = 0; i < 10; ++i) {
    futures.emplace_back(pool->push([&, id = i]() {
      if (id == 0)
        throw std::runtime_error("foo");

      using namespace std::chrono_literals;
      std::this_thread::sleep_for(2000ms);

      printf("%zd ", id);
    }));
  }

  try {
    folly::collect(futures).get();
  } catch (std::exception &e) {
    std::cout << " caught " << e.what() << std::endl;
  }
  printf(" done\n");
}
```

And this code may produce the following:
```
1 2 3  caught foo
 done
4 5 6 7 8 9
````
or
```
1 2 3 4 5  caught foo
6  done
7 8 9 
```

This is wrong, because it means that some futures continue working after the synchronization code `folly::collect(futures).get();` finishes.

What we'd like ideally to have is
```
1 2 3 4 5 6 7 8 9 
caught foo
done
```
 
The original snippet
```
  try {
    for (auto &f : futures)
      f.wait();
    for (auto &f : futures)
      f.result().value();
  } catch (std::exception &e) {
    std::cout << " caught " << e.what() << std::endl;
  }
```
produces
```
1 2 3 4 5 6 7 8 9  caught foo
 done
```
which is a desired behavior.

**Comment by zhengbuqian** [comment]
thanks @alexanderguzhva for the clarification! I should have used `collectAll` instead of `collect`, I thought `collect` waits for all futures.

File: src/index/diskann/diskann.cc [hunk 1]
```diff
@@ -490,16 +488,14 @@ DiskANNIndexNode<DataType>::Deserialize(const BinarySet& binset, const Config& c
...
-        for (auto& future : futures) {
-            if (TryDiskANNCall([&]() { future.wait(); }) != Status::success) {
-                all_searches_are_good = false;
-            }
-        }
+
+        bool failed = TryDiskANNCall([&]() { WaitAllSuccess(futures); }) != Status::success;
+
         if (warmup != nullptr) {
             diskann::aligned_free(warmup);
```
alexanderguzhva, line 495: a memory leak in case of exception

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -745,9 +746,7 @@ namespace diskann {
       }));
     }
 
-    for (auto &future : futures) {
-      future.wait();
-    }
+    knowhere::WaitAllSuccess(futures);
```
alexanderguzhva, line 749: memory leaks in case of exception
```
    if (samples != nullptr)
      delete[] samples;
    if (pq_code != nullptr)
      delete[] pq_code;
```

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -800,9 +799,7 @@ namespace diskann {
               stats + index);
         }));
       }
-      for (auto &future : futures) {
-        future.wait();
-      }
+      knowhere::WaitAllSuccess(futures);
```
alexanderguzhva, line 802: a memory leak in case of exception
```
delete[] stats;
```

thirdparty/DiskANN/src/index.cpp
```diff
@@ -816,9 +817,7 @@ namespace diskann {
             }
           }));
     }
-    for (auto &future : futures) {
-      future.wait();
-    }
+    knowhere::WaitAllSuccess(futures);
```
alexanderguzhva, line 820: memory leaks in case of exception
```
    delete[] distances;
    delete[] center;
```

alexanderguzhva (changes requested):
`collectAll` works, but possible memory leaks need to be addressed. Please change `T* x = new T[N];` into `std::vector<T> x(N);` or `std::unique_ptr<T[]> x(N);` whenever possible.

zhengbuqian (commented):
thanks Alex!

PR #357, opened 2024-01-22
fix:miss wait thread tasks finish in diskann.

cqy123456:
issue: #128
/kind improvement


Presburger (approved):

thirdparty/DiskANN/src/partition_and_pq.cpp
```diff
@@ -530,6 +530,9 @@ int generate_pq_data_from_pivots(const std::string data_file,
             }
           }));
     }
+    for (auto &future : futures) {
+      future.wait();
```
zhengbuqian, line 534: use knowhere::WaitAllSuccess introduced in #355 to make sure we catch and rethrow possible exceptions in the thread pool task.

PR #359, opened 2024-01-23
wrap IVF index train/build calls in knowhere build pool

zhengbuqian:
so OMP threads spawned will have low nice values

/kind improvement

Context in https://github.com/milvus-io/milvus/issues/30097


src/index/ivf/ivf.cc
```diff
@@ -224,9 +225,17 @@ class IvfIndexNode : public IndexNode {
     Status
     SerializeImpl(BinarySet& binset, IVFFlatTag) const;
 
+    Status
+    TrainInternal(const DataSet& dataset, const Config& cfg);
```
alexanderguzhva, line 229: I'd move this function into `private` section, because it is Internal, right? :)

Issue #364, opened 2024-01-26
Support integrating vector search engines by dlopen


foxspy:
Knowhere has integrated many opensource vector search engines such as faiss/hnsw/diskann etc. Moreover, it will support intergrating external search engines by dlopen. Index search engines can be used as long as it obeys the index registration interface.

PR #365, opened 2024-01-26
Support build cardinal as a plugin

foxspy:
issue: #364 

/kind enhancement

### DIFF
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -11,6 +11,9 @@
 
 #include "knowhere/comp/knowhere_config.h"
 
+#include <dlfcn.h>
+
+#include <filesystem>
 #include <string>
 
 #ifdef KNOWHERE_WITH_DISKANN
@@ -26,6 +29,8 @@
 #ifdef KNOWHERE_WITH_RAFT
 #include "common/raft/integration/raft_initialization.hpp"
 #endif
+#include "knowhere/factory.h"
+#include "knowhere/index.h"
 #include "simd/hook.h"
 
 namespace knowhere {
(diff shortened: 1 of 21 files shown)


alexanderguzhva:
Is it a general-purpose plugin system or just a mechanism for optionally attaching Cardinal? If the first one, then how many plugins are we supposed to have? If it is for Cardinal only, then why a plain 
```
#ifdef USE_CARDINAL
#include "cardinal/cardinal.h"
REGISTER_CARDINAL(whatever)
#endif
```
is not sufficient?

foxspy:
> Is it a general-purpose plugin system or just a mechanism for optionally attaching Cardinal? If the first one, then how many plugins are we supposed to have? If it is for Cardinal only, then why a plain
> 
> ```
> #ifdef USE_CARDINAL
> #include "cardinal/cardinal.h"
> REGISTER_CARDINAL(whatever)
> #endif
> ```
> 
> is not sufficient?

We will support not only cardinal. Someone can integrate their own indexes into knowhere without additional development work once their indexes obey the interface rule.

src/common/comp/knowhere_config.cc
```diff
@@ -125,6 +130,79 @@ KnowhereConfig::SetClusteringType(const ClusteringType clustering_type) {
     }
 }
 
+bool
+KnowhereConfig::InitPlugin(std::string path) {
+    static std::once_flag init_plugin_once_;
+    static bool init_result = false;
+
+    auto init = [&]() {
+        auto splitPaths = [&](std::string input, char delimiter) {
```
alexanderguzhva, line 139: Please use `std::filesystem` facilities for this

src/common/comp/knowhere_config.cc
```diff
@@ -125,6 +130,79 @@ KnowhereConfig::SetClusteringType(const ClusteringType clustering_type) {
...
+            LOG_KNOWHERE_INFO_ << "Plugin index list is empty";
+        }
+
+        for (auto& lib : libs) {
+            if (!std::filesystem::exists(lib)) {
+                LOG_KNOWHERE_ERROR_ << "Failed to find lib " << lib << " in filesystem";
+                init_result = false;
+                return init_result;
+            }
+            void* plugin = dlopen(lib.c_str(), RTLD_NOW | RTLD_GLOBAL);
```
alexanderguzhva, line 168: I could not find the corresponding `dlclose()`

alexanderguzhva (changes requested):
A couple of things to fix. Otherwise, looks good :)

PR #373, opened 2024-02-01
fix diskann async cache generation

foxspy:
issue: #372

### DIFF
--- a/include/knowhere/comp/thread_pool.h
+++ b/include/knowhere/comp/thread_pool.h
@@ -28,6 +28,7 @@
 #include <utility>
 
 #include "folly/executors/CPUThreadPoolExecutor.h"
+#include "folly/executors/task_queue/UnboundedBlockingQueue.h"
 #include "folly/futures/Future.h"
 #include "knowhere/expected.h"
 #include "knowhere/log.h"
@@ -35,6 +36,8 @@
 namespace knowhere {
 
 class ThreadPool {
+ public:
+    enum class QueueType { LIFO, FIFO };
 #ifdef __linux__
  private:
     class LowPriorityThreadFactory : public folly::NamedThreadFactory {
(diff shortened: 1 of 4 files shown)


include/knowhere/comp/thread_pool.h
```diff
@@ -48,23 +51,33 @@ class ThreadPool {
...
 
  public:
-    explicit ThreadPool(uint32_t num_threads, const std::string& thread_name_prefix)
-        : pool_(folly::CPUThreadPoolExecutor(
-              num_threads,
-              std::make_unique<
-                  folly::LifoSemMPMCQueue<folly::CPUThreadPoolExecutor::CPUTask, folly::QueueBehaviorIfFull::BLOCK>>(
-                  num_threads * kTaskQueueFactor),
-              std::make_shared<LowPriorityThreadFactory>(thread_name_prefix))) {
+    explicit ThreadPool(uint32_t num_threads, const std::string& thread_name_prefix, QueueType queueT = QueueType::LIFO)
```
alexanderguzhva, line 54: Personally, I would introduce `static ThreadPool CreateFIFO()` and `static ThreadPool CreateLIFO()`. These are two very different behaviors which one does not control via a boolean flag, and it make the code more readable.

liliu-z: Sounds good

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -268,6 +269,19 @@ namespace diskann {
     }
     this->count_visited_nodes.store(true);
 
+    // sync allocate memory
+    if (nhood_cache_buf == nullptr) {
+        nhood_cache_buf = new unsigned[num_nodes_to_cache * (max_degree + 1)];
+        memset(nhood_cache_buf, 0, num_nodes_to_cache * (max_degree + 1));
```
alexanderguzhva, line 275: `sizeof(unsigned)` is 4 on my machine, so memset should include `sizeof(unsigned)` as well.
But instead I would just  convert `nhood_cache_buf` from a pointer to a `std::vector<>`.

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -190,18 +189,20 @@ namespace diskann {
...
-    assert(this->nhood_cache_buf == nullptr && "nhoodc_cache_buf is not null");
-    assert(this->coord_cache_buf == nullptr && "coord_cache_buf is not null");
 
     auto ctx = this->reader->get_ctx();
 
-    nhood_cache_buf = new unsigned[num_cached_nodes * (max_degree + 1)];
-    memset(nhood_cache_buf, 0, num_cached_nodes * (max_degree + 1));
+    if (nhood_cache_buf == nullptr) {
+        nhood_cache_buf = new unsigned[num_cached_nodes * (max_degree + 1)];
+        memset(nhood_cache_buf, 0, num_cached_nodes * (max_degree + 1));
```
alexanderguzhva, line 197: `sizeof(unsigned)` is 4 on my machine, so memset should include `sizeof(unsigned)` as well.
But instead I would just  convert `nhood_cache_buf` from a pointer to a `std::vector<>`.

alexanderguzhva (changes requested):

PR #395, opened 2024-02-20
Update ci docker image and optimize gpu build

cydrain:


scripts/prepare_gpu_build.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# This script is to modify CMakeLists.txt to build knowhere for certain GPU serials
+#   GTX 1060 - sm_61
+#   GTX 1660 - sm_75
+sed 's/set(CMAKE_CUDA_ARCHITECTURES \${supported_archs})/set(CMAKE_CUDA_ARCHITECTURES \"61-real;75-real\")/' CMakeLists.txt > tmp
```
alexanderguzhva, line 6: sm_61 and sm_75 are for e2e tests, not production, correct?

cydrain: yes, this script is only used to speed up e2e test

Issue #394, opened 2024-02-19
gettid() compile failed because of not declare 


zhagnlu:
![image](https://github.com/zilliztech/knowhere/assets/11935707/0dbf97b5-1557-4924-b723-92a69e109438)
gettid is not standard glibc interface and may be not implement in gcc env. 
can wrapper it use 
pid_t gettid(){
    return static_cast<pid_t>(syscall(SYS_gettid));
}

PR #396, opened 2024-02-20
Replace gettid() with syscall(SYS_gettid)

zhengbuqian:
in some older glibc version gettid() wrapper does not exist.

issue: #394

### DIFF
--- a/include/knowhere/comp/thread_pool.h
+++ b/include/knowhere/comp/thread_pool.h
@@ -12,7 +12,11 @@
 #pragma once
 
 #include <omp.h>
+
+#ifdef __linux__
 #include <sys/resource.h>
+#include <sys/syscall.h>
+#endif
 
 #include <cerrno>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


zhengbuqian:
issue: #394

include/knowhere/comp/thread_pool.h
```diff
@@ -36,7 +40,7 @@ class ThreadPool {
         std::thread
         newThread(folly::Func&& func) override {
             return folly::NamedThreadFactory::newThread([&, func = std::move(func)]() mutable {
-                if (setpriority(PRIO_PROCESS, gettid(), 19) != 0) {
+                if (setpriority(PRIO_PROCESS, static_cast<pid_t>(syscall(SYS_gettid)), 19) != 0) {
```
alexanderguzhva, line 43: I'd rather go with the following solution
```
#if __GLIBC__ == 2 && __GLIBC_MINOR__ < 30
#include <sys/syscall.h>
#define gettid() syscall(SYS_gettid)
#endif
```
as suggested https://stackoverflow.com/questions/30680550/c-gettid-was-not-declared-in-this-scope/

alexanderguzhva (changes requested):

PR #404, opened 2024-02-23
fix: delete wrong used container in building pod

samhuang-z:


zhengbuqian:
issue: #405

ci/E2E2.groovy
```diff
@@ -19,7 +19,7 @@ pipeline {
     stages {
         stage("Build"){
             steps {
-                container("build"){
+                container("main"){
```
cydrain, line 22: can remove this line, since the default container is "main"

ci/E2E_GPU.groovy
```diff
@@ -20,7 +20,7 @@ pipeline {
         stage("Build"){
 
             steps {
-                container("build"){
+                container("main"){
```
cydrain, line 23: can remove this line, since the default container is "main"

PR #406, opened 2024-02-23
fix: diskann memory leak after failure

PwzXxm:
issue: #407 

Fixes some memory leaks in DiskANN caused by allocating memory without ownership. It could happen when
1. Memory allocate with `new` keyword
2. Exception throws during execution
3. Instructions freeing memory do not have a chance to execute because of the exception

This PR fixes most of them related to `new`. There are still some `new` and `alloc_aligned`, but they are enclosed in the destructors so rely on RAII they should be fine.

### DIFF
--- a/thirdparty/DiskANN/src/math_utils.cpp
+++ b/thirdparty/DiskANN/src/math_utils.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 #include <limits>
+#include <memory>
 #include <unordered_set>
 #include <malloc.h>
 #include <diskann/math_utils.h>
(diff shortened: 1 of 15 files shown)


thirdparty/DiskANN/src/math_utils.cpp
```diff
@@ -221,9 +220,12 @@ namespace math_utils {
 
     bool is_norm_given_for_pts = (pts_norms_squared != NULL);
 
-    float* pivs_norms_squared = new float[num_centers];
-    if (!is_norm_given_for_pts)
-      pts_norms_squared = new float[num_points];
+    auto pivs_norms_squared = std::make_unique<float>(num_centers);
+    std::unique_ptr<float[]> pts_norms_squared_deleter = nullptr;
```
PwzXxm, line 224: Use deleter approach as `pts_norms_squared` has default value in the parameter list

liliu-z (approved):
/lgtm

However, a lighter weight change can be:
1. Use smart point to manage data (already done in this PR)
2. In the usage, simply assign the address to a raw pointer to avoid too many changes. Since anyway we cannot get rid of all them in one shot.

I am fine about keeping this. Depends on your taste.

thirdparty/DiskANN/src/math_utils.cpp
```diff
@@ -221,9 +220,12 @@ namespace math_utils {
...
+    if (!is_norm_given_for_pts) {
+      pts_norms_squared_deleter = std::make_unique<float[]>(num_points);
+      pts_norms_squared = pts_norms_squared_deleter.get();
```
liliu-z, line 227: Yes, just like this. Simpler but safer.

PR #422, opened 2024-02-28
reduce ut run time

cqy123456:
issue:https://github.com/zilliztech/knowhere/issues/423


PwzXxm:
Reduced ~20 min for this PR.
The compilation and conan build takes ~40 min on the GitHub action runners. Maybe we need to introduce cache?

PwzXxm:
FP16/BF16 is covered in E2E as stated.

PR #428, opened 2024-02-29
[skip ci] Update Dockerfile to install knowhere dependancies

cydrain:
Issue: #429


zhengbuqian:
/lgtm

this PR is to create a docker image with all the knowhere dependencies and pre-loaded conan caches, so that during ci runs, our ci machine doesn't need to repeat those steps over and over again.

PR #439, opened 2024-03-05
Add customized plugin vector index support

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/364

/kind enhancement

### DIFF
--- a/include/knowhere/comp/knowhere_config.h
+++ b/include/knowhere/comp/knowhere_config.h
@@ -12,6 +12,7 @@
 #ifndef COMP_KNOWHERE_CONFIG_H
 #define COMP_KNOWHERE_CONFIG_H
 
+#include <mutex>
 #include <string>
 #include <vector>
 
@@ -113,6 +114,24 @@ class KnowhereConfig {
      */
     static void
     SetRaftMemPool();
+};
+
+class KnowherePluginManager {
+ public:
+    static KnowherePluginManager&
+    Instance();
+
+    bool
+    InitPlugin(const std::string& path);
+
+    virtual ~KnowherePluginManager() = default;
+
+ private:
+    KnowherePluginManager() = default;
+
+    bool inited_ = false;
+    bool init_results_ = false;
+    std::mutex init_mutex_;
 };
 
 }  // namespace knowhere
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -11,6 +11,9 @@
 
 #include "knowhere/comp/knowhere_config.h"
 
+#include <dlfcn.h>
+
+#include <filesystem>
 #include <string>
 
 #ifdef KNOWHERE_WITH_DISKANN
@@ -26,6 +29,8 @@
 #ifdef KNOWHERE_WITH_RAFT
 #include "common/raft/integration/raft_initialization.hpp"
 #endif
+#include "knowhere/factory.h"
+#include "knowhere/index.h"
 #include "simd/hook.h"
 
 namespace knowhere {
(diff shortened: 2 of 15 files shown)


src/common/factory.cc
```diff
@@ -19,6 +19,9 @@ IndexFactory::Create(const std::string& name, const int32_t& version, const Obje
     static_assert(KnowhereDataTypeCheck<DataType>::value == true);
     auto& func_mapping_ = MapInstance();
     auto key = GetIndexKey<DataType>(name);
+    if (func_mapping_.find(key) == func_mapping_.end()) {
+        LOG_KNOWHERE_ERROR_ << "failed to find index " << key << " in factory";
+    }
```
alexanderguzhva, line 24: This function returns an `Index` object, and if no mapping is found, then this function has to throw. Alternatively, this function may get changed to return `std::shared_ptr<Index>` or `std::optional<Index>` and return `{}`

tests/ut/plugin_demo/customIdx/customIdx.cc
```diff
@@ -0,0 +1,503 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#include <queue>
+
+#include "customIdx_config.h"
+#include "knowhere/index.h"
+#include "knowhere/index_node.h"
+#include "knowhere/log.h"
+
+namespace demo {
```
alexanderguzhva, line 19: this file has to contain a loooooooooooooooooooooooooooooooooot of comments, which describes what every piece does

tests/ut/test_plugin_index.cc
```diff
@@ -0,0 +1,133 @@
...
+#include "catch2/catch_test_macros.hpp"
+#include "catch2/generators/catch_generators.hpp"
+#include "knowhere/comp/brute_force.h"
+#include "knowhere/comp/index_param.h"
+#include "knowhere/comp/knowhere_config.h"
+#include "knowhere/factory.h"
+#include "knowhere/utils.h"
+#include "utils.h"
+
+TEST_CASE("Test Plugin Index", "[Correctness]") {
```
alexanderguzhva, line 24: this file has to contain a loooooooooooooooooooooooooooooooooooooooooot of comments as well

tests/ut/plugin_demo/customIdx/customIdx.cc
```diff
@@ -0,0 +1,503 @@
...
+        return "CustomIdx";
+    }
+
+ private:
+    int64_t row_;
+    int64_t dim_;
+    std::unique_ptr<VecT[]> data_;
+};
+}  // namespace demo
+extern "C" {
```
alexanderguzhva, line 480: why is `extern "C"` used here?

src/common/comp/knowhere_config.cc
```diff
@@ -182,4 +187,90 @@ KnowhereConfig::SetRaftMemPool() {
...
+        GetPluginIndexes getPluginIndexes = reinterpret_cast<GetPluginIndexes>(getImpList);
+        if (!getImpList) {
+            LOG_KNOWHERE_ERROR_ << "Failed to get the index list function in library with path = " << lib
+                                << ", error = " << dlerror();
+            dlclose(plugin);
+            init_results_ = false;
+            inited_ = true;
+            return init_results_;
+        }
+        auto indexList = getPluginIndexes();
```
alexanderguzhva, line 252: you're exposing C++ interface in a problematic way. It is a PITA, according to my past experience!!!
Please read https://stackoverflow.com/questions/17885060/passing-reference-to-stl-vector-over-dll-boundary , which is about MSVC, but the problems apply for linux compilers as well. Basically, when and which code allocates the `indexList` (calls `new`) and when and which code  destructs it (calls `delete`)?

I'd stick with a C interface coupled with wrappers.

tests/ut/plugin_demo/customIdx/customIdx_config.h
```diff
@@ -0,0 +1,30 @@
...
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#ifndef FLAT_CONFIG_H
+#define FLAT_CONFIG_H
+
+#include "knowhere/config.h"
+
+namespace demo {
+class CustomIdxConfig : public knowhere::BaseConfig {
```
alexanderguzhva, line 18: comments here are needed as well
What if I need to return strings, ints, doubles, arrays?
A demo needs to demonstrate it all

alexanderguzhva (changes requested):

tests/ut/plugin_demo/customIdx/customIdx.cc
```diff
@@ -0,0 +1,617 @@
...
+    } else {
+        LOG_KNOWHERE_ERROR_ << "Only support {L2, IP, COSINE} metric. " << metricType << " does not support";
+        throw knowhere::KnowhereException("Unsupported Metric");
+    }
+}
+
+}  // namespace calculator
+
+template <typename VecT>
+class CustomIdx : public knowhere::IndexNode {
```
alexanderguzhva, line 249: Please add a note on how the index is used. Basically, a user sees a constructor, `Train`, `Add`, `Search`. A typical user would think "let me construct, Add points, then Train, and then I will Search", right? 
So, the sequence needs to be mentioned and the reasoning behind it as well. Something like that `Train()` is like building a house, and once it is built, we may `Add()` people in it and then `Search()`.
Also, can 'Add()' be called multiple times, such as adding people one by one into the house, something that Faiss allows, or not?

include/knowhere/comp/knowhere_config.h
```diff
@@ -115,6 +116,24 @@ class KnowhereConfig {
...
 
+class KnowherePluginManager {
+ public:
+    static KnowherePluginManager&
+    Instance();
+
+    bool
+    InitPlugin(std::string path);
+
+    virtual ~KnowherePluginManager() = default;
```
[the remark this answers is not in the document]
foxspy: dlcose may lead to unpredictable effects because index generated from plugin may be destructed later. The order is unpredictable

tests/ut/plugin_demo/customIdx/customIdx.cc
foxspy: @liliu-z is about to reshap the interface in the next few days. We will just remain `Build` for immutable index, while growing index will use `Build` and `Add` for first initialization and subsequent insertions.

tests/ut/plugin_demo/customIdx/customIdx.cc
foxspy: To avoid compiler modify the function name in c++ style.

include/knowhere/comp/knowhere_config.h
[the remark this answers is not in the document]
alexanderguzhva: this means that we have to operate `Index` objects as `std::shared_ptr<Index>` and `KnowherePluginManager` has to maintain a `std::vector<std::shared_ptr<Index>>` in order to ensure the correct order of destructors.

src/common/comp/knowhere_config.cc
alexanderguzhva: No. 
The interface should not be a C++ one, but C one.

For example,
```
// dataset C API
knowhere::CAPI_Status CreateDataset(void** dataset_handle);
knowhere::CAPI_Status GetDatasetVar(const void* dataset_handle, const char* name, void** var_data);
knowhere::CAPI_Status SetDatasetVar(const void* dataset_handle, const char* name, const void* var_data);
knowhere::CAPI_Status GetDatasetTensor(const void* dataset_handle, void** data);
knowhere::CAPI_Status GetDatasetRows(const void** dataset_handle, int64_t* rows);
knowhere::CAPI_Status DestroyDataset(const void* dataset_handle); 
...

// config C API
knowhere::CAPI_Status GetConfigInt(const void** config_handle, const char* name, int* value);
knowhere::CAPI_Status GetConfigBool(const void** config_handle, const char* name, bool* value);
...

// index C API
knowhere::Status CreateIndex(void** index_handle, const int32_t version, const knowhere::Object* object);
knowhere::Status Train(const void* index_handle, const void* dataset_handle, const void* config_handle);
knowhere::Status Search(const void* index_handle, const void* dataset_handle, const void* config_handle, void* bitset_view_data, const size_t bitset_view_size, void** result_dataset_handle);
...

```
plus a C-to-C++ wrapper that a plugin would use.

alexanderguzhva: This C interface allows to write a plugin in any language, even Python or Rust.

alexanderguzhva: Let me write a full C interface...

alexanderguzhva: #449

PR #442, opened 2024-03-06
Decode traceID and spanID as hex string

cydrain:


cydrain:
/hold 
wait https://github.com/milvus-io/milvus/pull/31071 merged first

src/common/tracer.cc
```diff
@@ -140,4 +143,26 @@ EmptySpanID(const TraceContext* ctx) {
...
+}
+
+std::string
+HexToString(const std::string& input) {
+    std::string output;
+    std::stringstream ss;
+    for (size_t i = 0; i < input.length(); i += 2) {
+        std::string byteString = input.substr(i, 2);
+        char byte = static_cast<char>(std::stoi(byteString, nullptr, 16));
+        output += byte;
```
alexanderguzhva, line 163: `ss` is not used, so it seems that you originally wanted to do something like `ss << byte;`
So, please either remove `ss` or use it instead of `output += byte;`

alexanderguzhva (changes requested):

src/common/tracer.cc
cydrain: updated

PR #447, opened 2024-03-07
Allow brute force fallback for large k on GPU

### LINKED ISSUE #290: [FEA] Allow RAFT indexes to fall back to brute force search for large k
As a follow-up to #286, RAFT indexes should be able to fall back to a brute force search when k exceeds their maximum value or when it exceeds a value beyond which approximate methods no longer realistically offer any benefit

Presburger:
issue:  #290
This PR allows RAFT indexes to fall back to brute force if the requested k exceeds the value currently supported by the ANN implementation.

### DIFF
--- a/src/common/raft/integration/raft_knowhere_index.cuh
+++ b/src/common/raft/integration/raft_knowhere_index.cuh
@@ -15,10 +15,12 @@
  * limitations under the License.
  */
 #pragma once
+#include <atomic>
 #include <cmath>
 #include <cstdint>
 #include <istream>
 #include <limits>
+#include <mutex>
 #include <ostream>
 #include <raft/core/bitset.cuh>
 #include <raft/core/copy.cuh>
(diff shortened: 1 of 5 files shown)


src/common/raft/integration/raft_knowhere_index.cuh
```diff
@@ -464,18 +478,63 @@ struct raft_knowhere_index<IndexKind>::impl {
...
+                if (!exact_index_) {
+                    RAFT_EXPECTS(device_dataset_storage, "Exact search requires a dataset");
+                    auto const& res = raft::device_resources_manager::get_device_resources();
+                    auto index_params = raft_index_params_t<exact_index_kind>{};
+                    index_params.metric = raft_index_type::get_metric(*index_);
+                    exact_index_ = exact_raft_index_type::template build<data_type, exact_indexing_type,
+                                                                         exact_input_indexing_type>(
+                        res, index_params, *get_dataset_view<exact_input_indexing_type>());
+                }
+                std::atomic_thread_fence(std::memory_order_release);
```
alexanderguzhva, line 512: a problem if raft throws during the `build<>` call

Issue #444, opened 2024-03-06
Support cluster framework


chasingegg:
Milvus issue: https://github.com/milvus-io/milvus/issues/30633
Milvus major compaction will do clustering on different data segments, knowhere needs to support cluster framework for doing this.

PR #448, opened 2024-03-07
Support cluster framework

chasingegg:
issue #444 
/kind feature
/hold


include/knowhere/kmeans.h
```diff
@@ -0,0 +1,157 @@
...
+#include "knowhere/file_manager.h"
+
+namespace knowhere::kmeans {
+namespace {
+
+static constexpr int64_t MAX_TRAIN_SIZE = 10000000L * 700 * 4;
+
+template <typename T>
+inline bool
+load_bin_file(const std::string& fname, std::unique_ptr<T[]>& data, uint64_t& offset) {
```
alexanderguzhva, line 31: do I get it correct that this function loads the whole file into a given preallocated buffer into a given offset? If so, then could you please add a comment on what this function does exactly, otherwise it is somewhat confusing. Thanks.

src/common/kmeans.cc
```diff
@@ -0,0 +1,422 @@
...
+#include "knowhere/log.h"
+#include "knowhere/operands.h"
+#include "knowhere/utils.h"
+#include "simd/hook.h"
+
+namespace knowhere::kmeans {
+
+template <typename VecT>
+void
+KMeans<VecT>::elkan_L2(const VecT* x, const VecT* y, size_t d, size_t nx, size_t ny, uint32_t* ids, float* val) {
```
alexanderguzhva, line 39: my concert with elkan that it was MUCH slower that a regular implementatoin in my past experiments, up to 10x slower if I am not mistaken. Faiss uses the following BLAS call: https://github.com/facebookresearch/faiss/blob/dafdff110489db7587b169a0afee8470f220d295/faiss/utils/distances.cpp#L263

Would you consider providing a plain BLAS-based implementation as well?

src/common/kmeans.cc
```diff
@@ -0,0 +1,422 @@
...
+#include "knowhere/comp/time_recorder.h"
+#include "knowhere/dataset.h"
+#include "knowhere/log.h"
+#include "knowhere/operands.h"
+#include "knowhere/utils.h"
+#include "simd/hook.h"
+
+namespace knowhere::kmeans {
+
+template <typename VecT>
```
alexanderguzhva, line 37: faiss has `faiss::Clustering` class, just in case :)

src/common/kmeans.cc
```diff
@@ -0,0 +1,422 @@
...
+
+    // sample data for kmeans train, so we need to go through tail data to get overall cluster_id_mapping
+    if (sample_num < total_num) {
+        auto cluster_id_mapping_all = std::make_unique<uint32_t[]>(total_num);
+        std::memcpy(cluster_id_mapping_all.get(), cluster_id_mapping.get(), sample_num * sizeof(uint32_t));
+        auto cluster_dis = std::make_unique<float[]>(total_num - sample_num);
+
+        offset = 0;
+        for (int i = sample_files; i < file_paths.size(); i++) {
+            uint64_t dumb = 0;
```
alexanderguzhva, line 396: what a variable name :))) I'd use `dummy` in this case

include/knowhere/kmeans.h
```diff
@@ -0,0 +1,157 @@
...
+        actual_file_size = reader.tellg();
+    } else {
+        LOG_KNOWHERE_ERROR_ << "Could not open file: " << fname;
+        return false;
+    }
+    reader.seekg(0, std::ios::beg);
+    uint64_t npts, dim;
+    uint32_t npts_32, dim_32;
+    reader.read((char*)&npts_32, sizeof(uint32_t));
+    reader.read((char*)&dim_32, sizeof(uint32_t));
```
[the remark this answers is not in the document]
chasingegg: you are right, at least the number of points could be very large, change them to uint64_t

include/knowhere/kmeans.h
chasingegg: updated

cmake/libs/libfaiss.cmake
```diff
@@ -110,7 +110,7 @@ if(__X86_64)
             -Wno-unused-function
             -Wno-strict-aliasing>)
   target_link_libraries(
-    faiss PUBLIC OpenMP::OpenMP_CXX ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES}
+    faiss PUBLIC OpenMP::OpenMP_CXX openblas ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES}
```
alexanderguzhva, line 113: I'm not sure why we need `openblas` here, because there is already `${BLAS_LIBRARIES}` O_o

src/common/kmeans.cc
```diff
@@ -0,0 +1,483 @@
...
+openblas_set_num_threads(int num_threads);
+}
+
+namespace knowhere::kmeans {
+
+template <typename VecT>
+void
+KMeans<VecT>::exhaustive_L2sqr_blas(const VecT* x, const VecT* y, size_t d, size_t nx, size_t ny, uint32_t* ids,
+                                    float* val) {
+    static_assert(std::is_same_v<VecT, float>, "sgemm only support float now");
```
alexanderguzhva, line 57: What is the point of not just calling `faiss::exhaustive_L2sqr_blas()` here?

src/common/kmeans.cc
```diff
@@ -0,0 +1,453 @@
...
+            }
+            if (verbose_ &&
+                ((loss < std::numeric_limits<float>::epsilon()) || ((iter != 1) && ((old_loss - loss) / loss) < 0))) {
+                LOG_KNOWHERE_INFO_ << "Residuals unchanged: " << old_loss << " becomes " << loss
+                                   << ". Early termination.";
+                break;
+            }
+            old_loss = loss;
+        } else {
+            throw std::runtime_error(std::string("Algorithm: ") + std::string(algorithm) + " not supported yet.");
```
liliu-z, line 219: We don't throw exceptions. Use errorcode instead

chasingegg:
Make clustering (currently only one kmeans implmentation)the same level as index, so refactor some index-related .h and .cc to the index folder, the same as clustering.

src/clustering/kmeans/kmeans_config.h
```diff
@@ -0,0 +1,44 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#ifndef KMEANS_CONFIG_H
+#define KMEANS_CONFIG_H
+
+#include "knowhere/config.h"
+
+namespace knowhere {
+
+class KmeansConfig : public BaseConfig {
```
liliu-z, line 19: Too many KNN related param in the BaseConfig

include/knowhere/dataset.h
```diff
@@ -275,6 +292,15 @@ GenResultDataSet(const int64_t rows, const int64_t dim, const void* tensor) {
...
 
+inline DataSetPtr
+GenResultDataSet(const int64_t rows, const void* centroid_id_mapping) {
+    auto ret_ds = std::make_shared<DataSet>();
+    ret_ds->SetRows(rows);
+    ret_ds->SetCentroidIdMapping(centroid_id_mapping);
+    ret_ds->SetIsOwner(true);
+    return ret_ds;
+}
+
```
liliu-z, line 303: Let's put this Genxxx function to the user side.

include/knowhere/dataset.h
```diff
@@ -162,6 +168,17 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
         return nullptr;
     }
 
+    const void*
+    GetCentroidIdMapping() const {
```
liliu-z, line 172: And these Getter. We can have a index_dataset_util & another cluster_dataset_util to gather them.

include/knowhere/clustering/clustering.h
```diff
@@ -0,0 +1,149 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#ifndef CLUSTERING_H
+#define CLUSTERING_H
+
+#include "knowhere/binaryset.h"
+#include "knowhere/clustering/clustering_node.h"
```
liliu-z, line 16: It is a little bit weird to make `clustering` = `index`. A better name is needed, some thing like `cluster` and `cluster_operator`. These just an example for reference.

include/knowhere/clustering/clustering_node.h
```diff
@@ -0,0 +1,49 @@
...
+    // kmeans train, return id_mapping
+    virtual expected<DataSetPtr>
+    Train(const DataSet& dataset, const Config& cfg) = 0;
+
+    // clustering assign, return id_mapping
+    virtual expected<DataSetPtr>
+    Assign(const DataSet& dataset, const Config& cfg) = 0;
+
+    // return centroids, must be called after trained
+    virtual expected<DataSetPtr>
```
liliu-z, line 35: If we use a general object like `DataSet` as input/output. I will suggest to add more comments to declare what we need in side this Object. Or we can directly avoid using this.

src/clustering/kmeans/kmeans.cc
```diff
@@ -0,0 +1,462 @@
...
+        }
+        old_loss = loss;
+    }
+    build_time.RecordSection("");
+    return GenResultDataSet(rows, centroid_id_mapping.release());
+}
+
+template <typename DataType>
+expected<DataSetPtr>
+KmeansClusteringNode<DataType>::Assign(const DataSet& dataset, const Config& cfg) {
```
liliu-z, line 162: `cfg` is not used. Let's either mark it as `/* unused */` in the signature or get rid of this param

src/clustering/kmeans/kmeans.cc
```diff
@@ -0,0 +1,462 @@
...
+    if (!centroids_) {
+        LOG_KNOWHERE_ERROR_ << "clustering not trained";
+        return expected<DataSetPtr>::Err(Status::clustering_error, "clustering not trained");
+    }
+    auto rows = dataset.GetRows();
+    auto vecs = dataset.GetTensor();
+    knowhere::TimeRecorder build_time("Kmeans assign cost", 2);
```
liliu-z, line 169: Let's do a dim check here

src/clustering/kmeans/kmeans.cc
```diff
@@ -0,0 +1,462 @@
...
+        size_t start = i * block_size;
+        size_t end = std::min(n, (i + 1) * block_size);
+        futures.emplace_back(pool_->push([&, start, end]() {
+            if (std::is_same_v<DataType, float>) {
+                exhaustive_L2sqr_blas(vecs + start * dim_, centroids, dim_, end - start, num_clusters_,
+                                      closest_centroid + start, closest_centroid_distance + start);
+            } else {
+                elkan_L2(vecs + start * dim_, centroids, dim_, end - start, num_clusters_, closest_centroid + start,
+                         closest_centroid_distance + start);
+            }
```
liliu-z, line 451: Looks like this is for float. If we don't support other data types, let's simply return error.

src/clustering/kmeans/kmeans.cc
```diff
@@ -0,0 +1,462 @@
...
+        }));
+    }
+    for (auto& future : futures) {
+        future.wait();
+    }
```
liliu-z, line 456: Async call need to handle the error

src/clustering/kmeans/kmeans.cc
```diff
@@ -0,0 +1,462 @@
...
+                }
+            }
+            for (size_t j = 0; j < dim_; ++j) {
+                centroids_[c * dim_ + j] = DataType(centroids_tmp[j] / closest_docs[c].size());
+            }
+        }));
+    }
+    for (auto& future : futures) {
+        future.wait();
+    }
```
liliu-z, line 423: Same here, let's handle the error

liliu-z (changes requested):

include/knowhere/clustering/clustering.h
chasingegg: Lets go with cluster

include/knowhere/dataset.h
chasingegg: we could reuse current dataset actually

src/clustering/kmeans/kmeans.cc
chasingegg: remove elkan

chasingegg: cluster factory has limited data type to be float, already have a static assert

include/knowhere/comp/index_param.h
```diff
@@ -61,6 +65,7 @@ constexpr const char* INDEX_ENGINE_VERSION = "index_engine_version";
 constexpr const char* RETRIEVE_FRIENDLY = "retrieve_friendly";
 constexpr const char* DIM = "dim";
 constexpr const char* TENSOR = "tensor";
+constexpr const char* CENTROID_ID_MAPPING = "centroid_id_mapping";
```
liliu-z, line 68: Remove redundant

include/knowhere/comp/index_param.h
```diff
@@ -82,6 +87,7 @@ constexpr const char* TRACE_FLAGS = "trace_flags";
 constexpr const char* MATERIALIZED_VIEW_SEARCH_INFO = "materialized_view_search_info";
 constexpr const char* MATERIALIZED_VIEW_OPT_FIELDS_PATH = "opt_fields_path";
 constexpr const char* MAX_EMPTY_RESULT_BUCKETS = "max_empty_result_buckets";
+constexpr const char* NUM_CLUSTERS = "num_clusters";
```
liliu-z, line 90: Only used by UT

liliu-z:
/lgtm
Thanks for addressing all comments!

PR #450, opened 2024-03-08
[feature]add a new growing index: ivf sq cc

cqy123456:


include/knowhere/comp/index_param.h
```diff
@@ -89,6 +90,7 @@ constexpr const char* SSIZE = "ssize";
 constexpr const char* REORDER_K = "reorder_k";
 constexpr const char* WITH_RAW_DATA = "with_raw_data";
 constexpr const char* ENSURE_TOPK_FULL = "ensure_topk_full";
```
chasingegg, line 92: One thing to note: growing index does not provide interface for users to tune nprobe(although we have config in milvus.yaml), index itself should make sure topk result is full, otherwise we will enlarge search space to get more results. Could take ivf_flat_cc as an example, we have a ensure_topk_full option for faiss indexes.

PR #453, opened 2024-03-11
Upgrade faiss to 1.8.0

alexanderguzhva:
Upgrade Faiss to 1.8.0
/kind improvement


thirdparty/faiss/faiss/impl/ScalarQuantizerCodec_avx.h
```diff
@@ -284,8 +284,8 @@ struct SimilarityL2_avx<8> {
         accu8 = _mm256_fmadd_ps(tmp, tmp, accu8);
     }
 
-    FAISS_ALWAYS_INLINE void add_8_components_2(__m256 x, __m256 y) {
-        __m256 tmp = _mm256_sub_ps(y, x);
+    FAISS_ALWAYS_INLINE void add_8_components_2(__m256 x, __m256 y_2) {
```
cydrain, line 287: why need rename to "y_2" ?

cydrain:
/hold
will merge to main branch after Milvus 2.4 released

thirdparty/faiss/faiss/impl/ScalarQuantizerCodec_avx.h
alexanderguzhva: these were semi-automatic changes across the whole Faiss codebase. Let me cite it:

> Shadowed variables are variables in an inner scope with the same name as another variable in an outer scope. Having the same name for both variables might be semantically correct, but it can make the code confusing to read! It can also hide subtle bugs.

alexanderguzhva:
rebase on top of master

alexanderguzhva:
rebased to master

PR #464, opened 2024-03-14
Add new config type CFG_BYTES, and use it for trace_id and span_id

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/30961

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -23,6 +23,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <variant>
+#include <vector>
 
 #include "knowhere/comp/materialized_view.h"
 #include "knowhere/expected.h"
@@ -55,6 +56,10 @@ typedef nlohmann::json Json;
 
 #ifndef CFG_MATERIALIZED_VIEW_SEARCH_INFO_TYPE
 #define CFG_MATERIALIZED_VIEW_SEARCH_INFO_TYPE std::optional<MaterializedViewSearchInfo>
+#endif
+
+#ifndef CFG_BYTES
+#define CFG_BYTES std::optional<std::vector<uint8_t>>
 #endif
 
 template <typename T>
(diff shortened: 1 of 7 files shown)


alexanderguzhva:
@cydrain Why CFG_STRING is insufficient for this purpose?

cydrain:
> @cydrain Why CFG_STRING is insufficient for this purpose?

you can see https://github.com/milvus-io/milvus/issues/30961, convert an arbitrary binary into a string and save it into search config, there may be problem when call config.dump().
Saving a binary as a uint8_t vector, can avoid such kind of issue.

include/knowhere/config.h
```diff
@@ -57,6 +58,10 @@ typedef nlohmann::json Json;
 #define CFG_MATERIALIZED_VIEW_SEARCH_INFO_TYPE std::optional<MaterializedViewSearchInfo>
 #endif
 
+#ifndef CFG_VECTOR
+#define CFG_VECTOR std::optional<std::vector<uint8_t>>
```
zhengbuqian, line 62: use `CFG_BINARY` or `CFG_BYTES`

src/common/comp/brute_force.cc
```diff
@@ -63,10 +63,9 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
 #ifdef NOT_COMPILE_FOR_SWIG
     std::shared_ptr<tracer::trace::Span> span = nullptr;
     if (cfg.trace_id.has_value()) {
-        auto trace_id_str = tracer::HexToString(cfg.trace_id.value());
-        auto span_id_str = tracer::HexToString(cfg.span_id.value());
-        auto ctx = tracer::TraceContext{(uint8_t*)trace_id_str.c_str(), (uint8_t*)span_id_str.c_str(),
-                                        (uint8_t)cfg.trace_flags.value()};
+        auto trace_id = cfg.trace_id.value();
+        auto span_id = cfg.span_id.value();
+        auto ctx = tracer::TraceContext{trace_id.data(), span_id.data(), (uint8_t)cfg.trace_flags.value()};
```
zhengbuqian, line 68: can we avoid repeating these code?

cydrain: done

include/knowhere/config.h
cydrain: updated

PR #467, opened 2024-03-14
Add max_empty_result_buckerts for ScaNN range search

### LINKED ISSUE #454: Add param to control IVF range search
By now, IVF-serial index range search has only 2 params: radius and range_filter.
And the strategy for IVF range search is: 
search each bucket from close to far, till one of following conditions is met:
1). meet one bucket with no valid results
2). all buckets are visited

With above strategy, user has no ways to control IVF range search behavior.
In some cases, especially when bitset is used, there may be one bucket with no valid range search results,
this bucket will cause IVF range search terminate, and user has no way to improve their recall.

So in this PR, we introduce one new param named "empty_result_nbuckets", it means only when a certain num of continuous buckets
with empty results are met, range search flow will be terminated.
When "empty_result_nbuckets" is set to 1, its behavior is as same as the old strategy.

chasingegg:
Fix #454

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -769,14 +769,11 @@ IvfIndexNode<DataType, IndexType>::RangeSearch(const DataSet& dataset, const Con
                     }
 
                     // todo aguzhva: this is somewhat alogical. Refactor?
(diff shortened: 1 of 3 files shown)


liliu-z (approved):
/lgtm

PR #477, opened 2024-03-22
[skip ci] Update dockerfile

cydrain:
Issue: #429


ci/docker/builder/cpu/ubuntu20.04/Dockerfile
```diff
@@ -43,5 +43,8 @@ RUN git clone https://github.com/zilliztech/knowhere.git \
     && mkdir build && cd build \
     && conan install .. --build=missing -o with_ut=True -o with_diskann=True -s compiler.libcxx=libstdc++11 -s build_type=Release \
     && conan build .. \
+    && rm -rf Release \
+    && conan install .. --build=missing -o with_ut=True -o with_diskann=True -o with_cardinal=True -s compiler.libcxx=libstdc++11 -s build_type=Release \
+    && conan build .. \
```
zhengbuqian, line 48: what is this Dockerfile used for? why do we build twice with and without cardinal?

Issue #456, opened 2024-03-12
Integrate cardinal into knowhere


foxspy:
Intergrate enterprise vector search engine into knowhere directly, and delete knowhere-cloud repo

PR #480, opened 2024-03-25
fix UseDiskLoad interface

foxspy:
cardinal index will be built/load from disk file
issue: #456 
/kind enhancement

### DIFF
--- a/python/knowhere/knowhere.i
+++ b/python/knowhere/knowhere.i
@@ -36,6 +36,7 @@ typedef uint64_t size_t;
 #include <knowhere/comp/knowhere_config.h>
 #include <knowhere/comp/local_file_manager.h>
 #include <knowhere/comp/task.h>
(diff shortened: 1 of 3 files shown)


python/knowhere/knowhere.i
```diff
@@ -145,7 +145,7 @@ class IndexWrap {
  public:
     IndexWrap(const std::string& name, const int32_t& version) {
         GILReleaser rel;
-        if (knowhere::UseDiskLoad(name, version)) {
+        if (name == std::string("DISKANN")) {
```
alexanderguzhva, line 148: a magic constant?

chasingegg: why not use USeDiskLoad api here?

foxspy: > why not use USeDiskLoad api here?

UseDiskLoad will cause hnsw running with filemanager, this will lead ci fail

PR #483, opened 2024-03-27
Add iterator support to sparse inverted index

zhengbuqian:
Also added iterator for brute force sparse search.

For now both inverted index and wand uses a brute force approach.

/kind feature
https://github.com/zilliztech/knowhere/issues/193

### DIFF
--- a/include/knowhere/index_node.h
+++ b/include/knowhere/index_node.h
@@ -120,18 +120,21 @@ class PrecomputedDistanceIterator : public IndexNode::iterator {
  public:
     PrecomputedDistanceIterator(std::vector<std::pair<float, int64_t>>&& distances_ids, bool larger_is_closer)
         : comp_(larger_is_closer), results_(std::move(distances_ids)) {
-        sort_size_ = std::max((size_t)50000, results_.size() / 10);
+        sort_size_ = get_sort_size(results_.size());
         sort_next();
     }
 
     // Construct an iterator from a list of distances with index being id, filtering out zero distances.
-    PrecomputedDistanceIterator(std::vector<float> distances, bool larger_is_closer) : comp_(larger_is_closer) {
+    PrecomputedDistanceIterator(const std::vector<float>& distances, bool larger_is_closer) : comp_(larger_is_closer) {
+        // 30% is a ratio guesstimate of non-zero distances: probability of 2 random sparse splade vectors(100 non zero
+        // dims out of 30000 total dims) sharing at least 1 common non-zero dimension.
+        results_.reserve(distances.size() * 0.3);
         for (size_t i = 0; i < distances.size(); i++) {
             if (distances[i] != 0) {
                 results_.push_back(std::make_pair(distances[i], i));
             }
         }
-        sort_size_ = std::max((size_t)50000, results_.size() / 10);
+        sort_size_ = get_sort_size(results_.size());
         sort_next();
     }
 
@@ -148,6 +151,11 @@ class PrecomputedDistanceIterator : public IndexNode::iterator {
     }
 
  private:
+    static inline size_t
+    get_sort_size(size_t rows) {
+        return std::max((size_t)50000, rows / 10);
+    }
+
     // sort the next sort_size_ elements
     inline void
     sort_next() {
@@ -170,6 +178,10 @@ class PrecomputedDistanceIterator : public IndexNode::iterator {
             }
             if (b.second == -1) {
                 return true;
+            }
+            // to ensure deterministic behavior
+            if (a.first == b.first) {
+                return a.second < b.second;
             }
             return larger_is_closer ? a.first > b.first : a.first < b.first;
         }
--- a/include/knowhere/index_node.h
+++ b/include/knowhere/index_node.h
@@ -115,6 +115,72 @@ class IndexNode : public Object {
     Version version_;
 };
 
+// An iterator implementation that accepts a list of distances and ids and returns them in order.
+class PrecomputedDistanceIterator : public IndexNode::iterator {
+ public:
+    PrecomputedDistanceIterator(std::vector<std::pair<float, int64_t>>&& distances_ids, bool larger_is_closer)
+        : comp_(larger_is_closer), results_(std::move(distances_ids)) {
+        sort_size_ = std::max((size_t)50000, results_.size() / 10);
+        sort_next();
+    }
+
+    // Construct an iterator from a list of distances with index being id, filtering out zero distances.
+    PrecomputedDistanceIterator(std::vector<float> distances, bool larger_is_closer) : comp_(larger_is_closer) {
+        for (size_t i = 0; i < distances.size(); i++) {
+            if (distances[i] != 0) {
+                results_.push_back(std::make_pair(distances[i], i));
+            }
+        }
+        sort_size_ = std::max((size_t)50000, results_.size() / 10);
+        sort_next();
+    }
+
+    std::pair<int64_t, float>
+    Next() override {
+        sort_next();
+        auto& result = results_[next_++];
+        return std::make_pair(result.second, result.first);
+    }
+
+    [[nodiscard]] bool
+    HasNext() const override {
+        return next_ < results_.size() && results_[next_].second != -1;
+    }
+
+ private:
+    // sort the next sort_size_ elements
+    inline void
+    sort_next() {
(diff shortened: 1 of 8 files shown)


include/knowhere/index_node.h
```diff
@@ -115,6 +115,72 @@ class IndexNode : public Object {
     Version version_;
 };
 
+// An iterator implementation that accepts a list of distances and ids and returns them in order.
+class PrecomputedDistanceIterator : public IndexNode::iterator {
+ public:
+    PrecomputedDistanceIterator(std::vector<std::pair<float, int64_t>>&& distances_ids, bool larger_is_closer)
+        : comp_(larger_is_closer), results_(std::move(distances_ids)) {
+        sort_size_ = std::max((size_t)50000, results_.size() / 10);
```
alexanderguzhva, line 123: would it be possible to move `std::max((size_t)50000, results_.size() / 10)` into a separate static function, bcz the same code here and below is identical?

include/knowhere/index_node.h
```diff
@@ -115,6 +115,72 @@ class IndexNode : public Object {
...
+        sort_next();
+    }
+
+    // Construct an iterator from a list of distances with index being id, filtering out zero distances.
+    PrecomputedDistanceIterator(std::vector<float> distances, bool larger_is_closer) : comp_(larger_is_closer) {
```
alexanderguzhva, line 128: `const std::vector<float>& distances`

include/knowhere/index_node.h
```diff
@@ -115,6 +115,72 @@ class IndexNode : public Object {
...
+
+        bool
+        operator()(const std::pair<float, int64_t>& a, const std::pair<float, int64_t>& b) const {
+            if (a.second == -1) {
+                return false;
+            }
+            if (b.second == -1) {
+                return true;
+            }
+            return larger_is_closer ? a.first > b.first : a.first < b.first;
```
alexanderguzhva, line 174: what if `a.first == b.first`? Typically, ppl write something like this
````
if (a.first > b.first) return true;
if (a.first < b.first) return false;

// a.first == b.first case
return (a.second > b.second);
```

This is needed to ensure the deterministic behavior.

include/knowhere/index_node.h
```diff
@@ -115,6 +115,72 @@ class IndexNode : public Object {
...
+ public:
+    PrecomputedDistanceIterator(std::vector<std::pair<float, int64_t>>&& distances_ids, bool larger_is_closer)
+        : comp_(larger_is_closer), results_(std::move(distances_ids)) {
+        sort_size_ = std::max((size_t)50000, results_.size() / 10);
+        sort_next();
+    }
+
+    // Construct an iterator from a list of distances with index being id, filtering out zero distances.
+    PrecomputedDistanceIterator(std::vector<float> distances, bool larger_is_closer) : comp_(larger_is_closer) {
+        for (size_t i = 0; i < distances.size(); i++) {
```
alexanderguzhva, line 129: you could add `results_.reserve(distances.size());`, if it is expected that the number of zeros among `distances` is small

include/knowhere/sparse_utils.h
```diff
@@ -152,6 +152,7 @@ class SparseRow {
     dot(const SparseRow<T>& other) const {
         float product_sum = 0.0f;
         size_t i = 0, j = 0;
```
alexanderguzhva, line 154: Please use a separate line for introducing every variable
```
size_t i = 0;
size_t j = 0;
```

src/common/comp/brute_force.cc
```diff
@@ -685,6 +631,85 @@ BruteForce::AnnIterator(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+    auto xq = static_cast<const sparse::SparseRow<float>*>(query_dataset->GetTensor());
+    auto nq = query_dataset->GetRows();
+
+    BruteForceConfig cfg;
+    std::string msg;
+    auto status = Config::Load(cfg, config, knowhere::ITERATOR, &msg);
+    if (status != Status::success) {
+        LOG_KNOWHERE_ERROR_ << "Failed to load config, msg is: " << msg;
+        return expected<std::vector<std::shared_ptr<IndexNode::iterator>>>::Err(
+            status, "failed to brute force search sparse for iterator");
```
alexanderguzhva, line 652: the error message in somewhat unclear.

src/common/comp/brute_force.cc
```diff
@@ -685,6 +631,85 @@ BruteForce::AnnIterator(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+        span->SetAttribute(meta::NQ, nq);
+    }
+#endif
+
+    std::string metric_str = cfg.metric_type.value();
+    auto result = Str2FaissMetricType(metric_str);
+    if (result.error() != Status::success) {
+        LOG_KNOWHERE_ERROR_ << "Invalid metric type: " << cfg.metric_type.value();
+        return expected<std::vector<std::shared_ptr<IndexNode::iterator>>>::Err(
+            result.error(), "failed to brute force search sparse for iterator");
```
alexanderguzhva, line 672: same

src/common/comp/brute_force.cc
```diff
@@ -685,6 +631,85 @@ BruteForce::AnnIterator(const DataSetPtr base_dataset, const DataSetPtr query_da
...
+    }
+    if (!IsMetricType(metric_str, metric::IP)) {
+        LOG_KNOWHERE_ERROR_ << "Invalid metric type: " << cfg.metric_type.value();
+        return expected<std::vector<std::shared_ptr<IndexNode::iterator>>>::Err(
+            Status::invalid_metric_type, "failed to brute force search sparse for iterator");
```
alexanderguzhva, line 677: same

src/index/sparse/sparse_inverted_index.h
```diff
@@ -186,6 +186,29 @@ class InvertedIndex {
...
+        auto q_threshold = *pos;
+        std::shared_lock<std::shared_mutex> lock(mu_);
+        auto distances = compute_all_distances(query, q_threshold);
+        for (size_t i = 0; i < distances.size(); ++i) {
+            if (bitset.empty() || !bitset.test(i)) {
+                continue;
+            }
+            distances[i] = 0.0f;
+        }
+        return distances;
```
alexanderguzhva, line 209: technically, the code above is not performance-optimal.
There is no need to compute distances inside `compute_all_distances` for rows with `!bitset.test(i)`.

include/knowhere/index_node.h
zhengbuqian: Done

include/knowhere/index_node.h
zhengbuqian: I used `results_.reserve(distances.size() * 0.3);`. 30% is a ratio guesstimate of non-zero distances: probability of 2 random sparse splade vectors(100 non zero dims out of 30000 total dims) sharing at least 1 common non-zero dimension.

src/common/comp/brute_force.cc
zhengbuqian: including the msg now

src/index/sparse/sparse_inverted_index.h
zhengbuqian: `compute_all_distances` proceeds column by column. We can't skip entire rows directly in `compute_all_distances`. We'd have more bitset tests if we want to skip the computes which is slower.

```cpp
std::vector<float>
    compute_all_distances2(const SparseRow<T>& q_vec, T q_threshold, const BitsetView& bitset) const {
        std::vector<float> scores(n_rows_internal(), 0.0f);
        for (size_t idx = 0; idx < q_vec.size(); ++idx) {
            auto [i, v] = q_vec[idx];
            if (v < q_threshold || i >= n_cols_internal()) {
                continue;
            }
            auto lut_it = inverted_lut_.find(i);
            if (lut_it == inverted_lut_.end()) {
                continue;
            }
            // TODO: improve with SIMD
            auto& lut = lut_it->second;
            for (size_t j = 0; j < lut.size(); j++) {
                auto [idx, val] = lut[j];
                // branching for each value, instead of each row
                if (bitset.empty() || !bitset.test(idx)) {
                    scores[idx] += v * float(val);
                }
            }
        }
        return scores;
    }
```

alexanderguzhva:
/lgtm
overall

PR #487, opened 2024-04-02
add operator= for index

Presburger:


include/knowhere/index.h
```diff
@@ -43,6 +43,17 @@ class Index {
         node = idx.node;
     }
 
+    Index<T1>&
+    operator=(const Index<T1>& idx) {
+        if (idx.node == nullptr) {
```
alexanderguzhva, line 48: Please add `if (*this == idx)` case as well

alexanderguzhva (changes requested):

Issue #500, opened 2024-04-16
improve the iterator implementation


zhengbuqian:
so that we can avoid as much duplicate code between indexes as possible and form a good foundation for Cardinal to add iterator support

zhengbuqian:
TODOs after #501:

- [ ] hnsw iterator use brute force for high filter rate
- [ ] hnsw iterator to limit `to_visit` size by search 1 step every several `Next()` calls.
- [ ] ivf iterator to copy and own query data
- [ ] DiskANN impl
- [ ] IVF-SQ8 impl

PR #501, opened 2024-04-16
Add a common subclass for IndexNode::iterator

zhengbuqian:
so that different indexes only need to implement next_batch. Any index that gradually adds new points as the `iterator::Next()` is called should implement their iterator by inheriting this new class.

Updated the impl for supported ivf and hnsw index.

issue: https://github.com/zilliztech/knowhere/issues/500
/kind: enhancement

### DIFF
--- a/include/knowhere/index_node.h
+++ b/include/knowhere/index_node.h
@@ -12,6 +12,11 @@
 #ifndef INDEX_NODE_H
 #define INDEX_NODE_H
 
+#include <functional>
+#include <queue>
+#include <utility>
+#include <vector>
+
 #include "knowhere/binaryset.h"
 #include "knowhere/bitsetview.h"
 #include "knowhere/config.h"
@@ -22,6 +27,7 @@
 #include "knowhere/version.h"
 
 namespace knowhere {
+
 class IndexNode : public Object {
  public:
     IndexNode(const int32_t ver) : version_(ver) {
@@ -57,7 +63,7 @@ class IndexNode : public Object {
         virtual std::pair<int64_t, float>
         Next() = 0;
         [[nodiscard]] virtual bool
-        HasNext() const = 0;
+        HasNext() = 0;
         virtual ~iterator() {
         }
     };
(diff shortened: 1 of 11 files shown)


include/knowhere/index_node.h
```diff
@@ -115,6 +124,114 @@ class IndexNode : public Object {
...
+    // will be called only if refine_ratio_ is not 0.
+    virtual float
+    raw_distance(int64_t id) {
+        if (!refine_) {
+            throw std::runtime_error("raw_distance should not be called for indexes without quantization");
+        }
+        throw std::runtime_error("raw_distance not implemented");
+    }
+
+    const bool refine_ratio_;
```
alexanderguzhva, line 198: should be `float`

include/knowhere/index_node.h
```diff
@@ -115,6 +124,114 @@ class IndexNode : public Object {
...
+    const bool refine_;
+
+    std::priority_queue<std::pair<float, int64_t>, std::vector<std::pair<float, int64_t>>, PairComparator> res_;
```
alexanderguzhva, line 201: there are too many `std::pair<float, int64_t>` structs in the code. I'd create a simple 2-field structure with clear field names, because the current code uses `.first` or `.second` and who knows that they mean

src/index/ivf/ivf.cc
```diff
@@ -247,14 +247,24 @@ class IvfIndexNode : public IndexNode {
     Status
     TrainInternal(const DataSet& dataset, const Config& cfg);
 
+    static bool
+    IsQuantized() {
```
alexanderguzhva, line 251: you can make this function `constexpr`
Also, what about SCANN?

zhengbuqian:
/hold

this branch is currently based on #505, need to rebase after #505 is merged

include/knowhere/index_node.h
```diff
@@ -115,23 +124,108 @@ class IndexNode : public Object {
...
+// with quantization, override `raw_distance`.
+class IndexIterator : public IndexNode::iterator {
+ public:
+    IndexIterator(bool larger_is_closer, float refine_ratio = 0.0f)
+        : refine_ratio_(refine_ratio), refine_(refine_ratio != 0.0f), sign_(larger_is_closer ? -1 : 1) {
+    }
+
+    std::pair<int64_t, float>
+    Next() override {
+        if (!initialized_) {
```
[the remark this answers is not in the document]
zhengbuqian: initialization must be done after constructor of subclass has finished, which we don't know when that'll happen in this class

foxspy (approved):
/lgtm

PR #502, opened 2024-04-16
knowhere light for milvus local

Presburger:
issue: #503


alexanderguzhva:
@Presburger  What is 'knowhere light' exactly?

Presburger:
> @Presburger What is 'knowhere light' exactly?

knowhere light weight version, only keep hnsw and brute force works. this version for project "milvus local", a python based milvus.

PR #505, opened 2024-04-16
make IdVal more generic

zhengbuqian:
/kind improvement

### DIFF
--- a/include/knowhere/object.h
+++ b/include/knowhere/object.h
@@ -13,11 +13,36 @@
 #define OBJECT_H
 
 #include <atomic>
+#include <cassert>
 #include <iostream>
 
 #include "knowhere/file_manager.h"
 
 namespace knowhere {
+
+template <typename I, typename T>
+struct IdVal {
+    I id;
+    T val;
+
+    IdVal() = default;
+    IdVal(I id, T val) : id(id), val(val) {
+    }
+
+    inline friend bool
+    operator<(const IdVal<I, T>& lhs, const IdVal<I, T>& rhs) {
+        return lhs.val < rhs.val || (lhs.val == rhs.val && lhs.id < rhs.id);
+    }
+    inline friend bool
+    operator>(const IdVal<I, T>& lhs, const IdVal<I, T>& rhs) {
+        return !(lhs < rhs);
+    }
+
+    inline friend bool
+    operator==(const IdVal<I, T>& lhs, const IdVal<I, T>& rhs) {
+        return lhs.id == rhs.id && lhs.val == rhs.val;
+    }
+};
 
 class Object {
  public:
--- a/include/knowhere/object.h
+++ b/include/knowhere/object.h
@@ -15,6 +15,7 @@
 #include <atomic>
 #include <cassert>
 #include <iostream>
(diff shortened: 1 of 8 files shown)


include/knowhere/object.h
```diff
@@ -13,12 +13,37 @@
...
 
 #include <atomic>
+#include <cassert>
 #include <iostream>
 
 #include "knowhere/file_manager.h"
 
 namespace knowhere {
 
+template <typename I, typename T>
```
zhengbuqian, line 23: I tried to put this in utils.h but for unknown reason it just won't compile 🧐

include/knowhere/object.h
```diff
@@ -13,12 +13,37 @@
...
+    IdVal(I id, T val) : id(id), val(val) {
+    }
+
+    inline friend bool
+    operator<(const IdVal<I, T>& lhs, const IdVal<I, T>& rhs) {
+        return lhs.val < rhs.val || (lhs.val == rhs.val && lhs.id < rhs.id);
+    }
+    inline friend bool
+    operator>(const IdVal<I, T>& lhs, const IdVal<I, T>& rhs) {
+        return !(lhs < rhs);
```
alexanderguzhva, line 38: `!(lhs < rhs)` implies `lhs >= rhs`, not `lhs > rhs`

alexanderguzhva (changes requested):

include/knowhere/object.h
zhengbuqian: good catch! 

also added a commit to replace std::pair<float, int64_t> in PrecomputedDistanceIterator with IdVal

thirdparty/faiss/faiss/utils/distances.h
```diff
@@ -234,7 +236,7 @@ void all_inner_product(
         size_t d,
         size_t nx,
         size_t ny,
-        std::vector<std::pair<float, int64_t>>& output,
+        std::vector<knowhere::DistId>& output,
```
alexanderguzhva, line 239: Is it possible to avoid modifying Faiss in this case?

thirdparty/faiss/faiss/utils/distances.h
```diff
@@ -305,7 +307,7 @@ void all_cosine(
...
```
alexanderguzhva, line 310: same

thirdparty/faiss/faiss/utils/distances.h
alexanderguzhva: It adds more pain to me whenever I need to propagate the changes from the Faiss baseline. So, I expect and insist that the changes to the faiss core code to be the last resort.
But it should be acceptable if the functions are not from the original faiss.
Could you please check whether you'll be able to put your change on top of #453 and see whether they are compatible? Just check, don't rebase yet :) Thanks

zhengbuqian: I just tried rebasing a copy of `iterator-new-interface` onto `merge_faiss_changes_from_baseline_v2a` and yes they are compatible. The only conflict is `CollectAllResultHandler` which is easy to fix.

PR #509, opened 2024-04-18
make DataSet gen helpers available to pyknowhere

zhengbuqian:
/kind improvement
/hold

This pr only changes the `const int64_t*` and `const size_t*` arguments of `GenIdsDataSet/GenResultDataSet` to `const void*` when compiling for swig.

When `const int64_t*` and `const size_t*` is used as argument type, swig build fails due to `error: invalid conversion from 'const long long int*' to 'const int64_t*' {aka 'const long int*'} [-fpermissive]`.

### DIFF
--- a/include/knowhere/dataset.h
+++ b/include/knowhere/dataset.h
@@ -254,13 +254,16 @@ GenDataSet(const int64_t nb, const int64_t dim, const void* xb) {
     return ret_ds;
 }
 
-#ifdef NOT_COMPILE_FOR_SWIG
-// TOOD: python wheel build error for this API, need check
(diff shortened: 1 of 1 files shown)


zhengbuqian:
I think python3 wheel can now successfully build

/unhold

include/knowhere/dataset.h
```diff
@@ -254,8 +254,8 @@ GenDataSet(const int64_t nb, const int64_t dim, const void* xb) {
     return ret_ds;
 }
 
+// swig won't compile when using int64_t* or size_t* as parameter
```
cydrain, line 257: it's because swig not support api override

PR #512, opened 2024-04-19
The workload of multi-GPU tasks is more evenly distributed

Presburger:
issue: #513

### DIFF
--- a/src/common/raft/integration/raft_knowhere_index.cuh
+++ b/src/common/raft/integration/raft_knowhere_index.cuh
@@ -385,6 +385,8 @@ struct raft_knowhere_index<IndexKind>::impl {
     void
     train(raft_knowhere_config const& config, data_type const* data, knowhere_indexing_type row_count,
           knowhere_indexing_type feature_count) {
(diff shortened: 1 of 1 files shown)


src/common/raft/integration/raft_knowhere_index.cuh
```diff
@@ -468,6 +470,7 @@ struct raft_knowhere_index<IndexKind>::impl {
     search(raft_knowhere_config const& config, data_type const* data, knowhere_indexing_type row_count,
            knowhere_indexing_type feature_count, knowhere_bitset_data_type const* bitset_data,
            knowhere_bitset_indexing_type bitset_byte_size, knowhere_bitset_indexing_type bitset_size) const {
+        RAFT_EXPECTS(device_id >= 0, "Device id has not yet been setted");
```
alexanderguzhva, line 473: I'd write `Device id has not been selected yet. This implies that the train() method has not been called.` 
Otherwise, it is unclear about what needs to be done to fix this warning.

PR #516, opened 2024-04-22
remove no-as-needed

Presburger:


chasingegg:
`no-as-needed` targets to reduce so size, but does not show much difference in our case, and it is not recognized by clang-15 compiler, so remove it

PR #517, opened 2024-04-23
removed unused range util for faiss

zhengbuqian:
/kind improvement

### DIFF
--- a/src/common/range_util.cc
+++ b/src/common/range_util.cc
@@ -18,61 +18,6 @@
 namespace knowhere {
 
 ///////////////////////////////////////////////////////////////////////////////
-// For Faiss index types
-size_t
-CountValidRangeSearchResult(const faiss::RangeSearchResult& res, const bool is_ip, const int64_t nq, const float radius,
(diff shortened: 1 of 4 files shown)


src/common/range_util.cc
```diff
@@ -17,61 +17,6 @@
...
-            o_labels[num] = id;
-            o_distances[num] = dis;
-            num++;
-        }
-    }
-    KNOWHERE_THROW_IF_NOT_FMT(num == o_size, "%" SCNd64 " not equal %" SCNd64, num, o_size);
-}
-
-void
-GetRangeSearchResult(const faiss::RangeSearchResult& res, const bool is_ip, const int64_t nq, const float radius,
```
cydrain, line 58: this API is used in flat.cc and ivf.cc, please double check

zhengbuqian: the `GetRangeSearchResult(const std::vector<std::vector<float>>& result_distances, const std::vector<std::vector<int64_t>>& result_labels` variant is used there

PR #518, opened 2024-04-23
Use specific commit rather than branch

PwzXxm:


liliu-z:
Let use a short one next time

cmake/libs/libcardinal.cmake
```diff
@@ -1,4 +1,4 @@
-set(CARDINAL_VERSION master)
+set(CARDINAL_VERSION 0bb6d5771a7c279362bd3b6ea7dc94b53d0c4de1 )
```
foxspy, line 1: use short commit tag?

Issue #519, opened 2024-04-24
Consistency in the computation logic between growing and sealed segment


foxspy:
Milvus segments have two states: growing and sealed, which may correspond to different types of indexes. These indexes need to maintain consistency in their computation logic; otherwise, when a segment transitions from growing to sealed, the index query results may differ slightly.

PR #521, opened 2024-04-24
[Bugfix] Add patch for bf16 distance computation

foxspy:
issue: #519 
/kind enhancement

### DIFF
--- a/include/knowhere/comp/knowhere_config.h
+++ b/include/knowhere/comp/knowhere_config.h
@@ -41,6 +41,17 @@ class KnowhereConfig {
     SetSimdType(const SimdType simd_type);
 
     /**
+     *The purpose of this interface is: part of the sealed indexes default to using bf16 as the base data to achieve
+     *higher capacity; to ensure consistency in computation between growing and sealed, it is necessary to maintain the
+     *same precision in growing calculations as in sealed.
+     */
+    static void
+    EnablePatchForComputeFP32AsBF16();
+
+    static void
+    DisablePatchForComputeFP32AsBF16();
+
+    /**
      * Set openblas threshold
      *   if nq < use_blas_threshold, calculated by omp
      *   else, calculated by openblas
--- a/src/simd/distances_avx.cc
+++ b/src/simd/distances_avx.cc
@@ -19,6 +19,8 @@
 
 namespace faiss {
 
+bool enable_avx_patch_fp32_bf16 = false;
+
 #define ALIGNED(x) __attribute__((aligned(x)))
 
 // reads 0 <= d < 4 floats as __m128
@@ -38,6 +40,38 @@ masked_read(int d, const float* x) {
     // cannot use AVX2 _mm_mask_set1_epi32
 }
 
+namespace {
+
+inline __m256
+avx2_bf16_patch(__m256 f256) {
+    if (!enable_avx_patch_fp32_bf16) {
+        return f256;
+    } else {
+        __m256i i_f256 = _mm256_castps_si256(f256);
+        __m256i rounding_increment = _mm256_set1_epi32(0x8000);
+
+        __m256i mask = _mm256_set1_epi32(0xFFFF0000);
+
+        return _mm256_castsi256_ps(_mm256_and_si256(_mm256_add_epi32(i_f256, rounding_increment), mask));
+    }
+}
+
+inline __m128
+avx_bf16_patch(__m128 f128) {
+    if (!enable_avx_patch_fp32_bf16) {
+        return f128;
+    } else {
+        __m128i i_f128 = _mm_castps_si128(f128);
+        __m128i rounding_increment = _mm_set1_epi32(0x8000);
+
+        __m128i mask = _mm_set1_epi32(0xFFFF0000);
+
+        return _mm_castsi128_ps(_mm_and_si128(_mm_add_epi32(i_f128, rounding_increment), mask));
+    }
+}
+
+}  // namespace
+
 float
 fvec_inner_product_avx(const float* x, const float* y, size_t d) {
     __m256 msum1 = _mm256_setzero_ps();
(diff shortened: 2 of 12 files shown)


include/knowhere/comp/knowhere_config.h
```diff
@@ -40,6 +40,17 @@ class KnowhereConfig {
     static std::string
     SetSimdType(const SimdType simd_type);
 
+    /**
+     *The purpose of this interface is: part of the sealed indexes default to using bf16 as the base data to achieve
+     *higher capacity; to ensure consistency in computation between growing and sealed, it is necessary to maintain the
+     *same precision in growing calculations as in sealed.
+     */
+    static void
+    EnablePatchForBF16();
```
PwzXxm, line 49: Maybe a better name? Throwing all the context away, solely looking at the name and the comments, I might be confused at what this patch do. `EnableCalculatingFp32AsBf16` or something similar?

tests/ut/utils.h
```diff
@@ -46,6 +46,17 @@ GenDataSet(int rows, int dim, int seed = 42) {
     return ds;
 }
 
+inline knowhere::DataSetPtr
+GenFloatDataSet(int rows, int dim, int seed = 42) {
```
PwzXxm, line 50: Where does this link to?

foxspy: ut; previously all generated data are integer

src/simd/distances_avx512.cc
```diff
@@ -36,6 +38,37 @@ masked_read(int d, const float* x) {
     return _mm_load_ps(buf);
     // cannot use AVX2 _mm_mask_set1_epi32
 }
+namespace {
+inline __m512
+avx512_bf16_patch(__m512&& f512) {
+    if (!enable_avx512_patch_fp32_bf16) {
```
PwzXxm, line 44: Will this runtime branching hurts performance? Is it necessary to template it?

src/simd/distances_avx.cc
```diff
@@ -38,14 +40,38 @@ masked_read(int d, const float* x) {
     // cannot use AVX2 _mm_mask_set1_epi32
 }
 
+namespace {
+
+inline __m256
+avx2_bf16_patch(__m256&& f256) {
```
hhy3, line 46: Why && here? Just passing by value seems better, cause the compiler can optimize it better

src/simd/distances_avx.cc
```diff
@@ -38,14 +40,38 @@ masked_read(int d, const float* x) {
...
+    if (!enable_avx_patch_fp32_bf16) {
+        return f256;
+    } else {
+        __m256i mask = _mm256_set1_epi32(0xFFFF0000);
+        return _mm256_castsi256_ps(_mm256_and_si256(_mm256_castps_si256(f256), mask));
+    }
+}
+
+inline __m128
+avx_bf16_patch(__m128&& f128) {
```
hhy3, line 56: same

src/simd/distances_avx512.cc
PwzXxm: branching before call the dist cmp function, rather than within this function?

src/simd/distances_avx.cc
```diff
@@ -38,14 +40,38 @@ masked_read(int d, const float* x) {
...
 }
 
+namespace {
+
+inline __m256
+avx2_bf16_patch(__m256&& f256) {
+    if (!enable_avx_patch_fp32_bf16) {
+        return f256;
+    } else {
+        __m256i mask = _mm256_set1_epi32(0xFFFF0000);
```
hhy3, line 50: Seems not compatible with sealed index implementation. It does not just truncate fp32 value to bf16.

src/simd/distances_avx.cc
```diff
@@ -38,14 +40,38 @@ masked_read(int d, const float* x) {
...
+        return _mm256_castsi256_ps(_mm256_and_si256(_mm256_castps_si256(f256), mask));
+    }
+}
+
+inline __m128
+avx_bf16_patch(__m128&& f128) {
+    if (!enable_avx_patch_fp32_bf16) {
+        return f128;
+    } else {
+        __m128i mask = _mm_set1_epi32(0xFFFF0000);
```
hhy3, line 60: same

src/simd/distances_avx512.cc
```diff
@@ -36,6 +38,37 @@ masked_read(int d, const float* x) {
     return _mm_load_ps(buf);
     // cannot use AVX2 _mm_mask_set1_epi32
 }
+namespace {
```
hhy3, line 41: same for this file

src/simd/distances_avx512.cc
foxspy: Branch prediction should hit all targets, so the overhead can be ignored

src/simd/distances_avx.cc
foxspy: good catch!

PwzXxm:
lgtm overall. Waiting for @hhy3 double-check.

alexanderguzhva:
This looks Horrible, because it is a really dirty hack.
* It is almost impossible to identify the reason of why such a patch is needed without comments in the code (which are missing). 
* Why the patch is applied to `y`, but not in `x` in `distance(const float* x, const float* y, const size_t dim)`. 
* Why is it needed to apply the patch every time instead of prepatching `y` somewhere else 
* Why it is not possible to patch a query vector on the level of milvus-to-knowhere parameters passing instead of patching all the distance functions.  
* What is going to happen in PPC64, which could have a big-endian format.

Too many questions and too few answers.

alexanderguzhva:
or, at least, one could introduce a different set of distance computation functions (which include this patch) and assign different distance pointers in simdhook. This won't hurt the performance and the code becomes a bit more clear.

include/knowhere/comp/knowhere_config.h
```diff
@@ -40,6 +40,17 @@ class KnowhereConfig {
     static std::string
     SetSimdType(const SimdType simd_type);
 
+    /**
+     *The purpose of this interface is: part of the sealed indexes default to using bf16 as the base data to achieve
+     *higher capacity; to ensure consistency in computation between growing and sealed, it is necessary to maintain the
+     *same precision in growing calculations as in sealed.
+     */
+    static void
+    EnablePatchForComputeFP32AsBF16();
```
PwzXxm, line 49: Another interesting thing to mention is that will this change the behavior of all segments on a QN in Milvus, especially in the multi-tenant situation?

PwzXxm: also the simhook approach

PR #523, opened 2024-04-24
Add loongarch64 support

zhangwenlong8911:


PwzXxm:
@zhangwenlong8911 Thanks for your contribution. Could you also commit it with `git commit -s` to sign off DCO, please?

zhangwenlong8911:
> lgtm overall @zhangwenlong8911 did you have a chance to run knowhere unit tests on a native loongarch64 hardware?

I ran ./Release/tests/ut/knowhere_tests, the result is as follows:

```
===============================================================================
test cases:        36 |        35 passed |  1 failed
assertions: 151434844 | 151434832 passed | 12 failed
```

the fail log is :
```
E0426 01:38:46.616202 224128 ivf_config.h:118] [KNOWHERE][CheckAndAdjust][knowhere_tests] SCANN index is not supported on the current CPU model, avx2 support is needed for x86 arch.
-------------------------------------------------------------------------------
Test Mem Index With Float Vector
  Test Serialize/Deserialize
-------------------------------------------------------------------------------
/home/zwl/anolis/knowhere/tests/ut/test_search.cc:344
...............................................................................
/home/zwl/anolis/knowhere/tests/ut/test_search.cc:365: FAILED:
  REQUIRE( idx.Build(*train_ds, json) == knowhere::Status::success )
with expansion:
  20 == 0
with messages:
  name := "SCANN"
  cfg_json := "{"dim":128,"k":120,"metric_type":"COSINE","nlist":16,"nprobe":
  14,"radius":0.99,"range_filter":1.01,"reorder_k":200,"with_raw_data":true}"
```

alexanderguzhva:
@zhangwenlong8911 This is fine, please add the following pieces of logic into `knowhere/tests/ut/test_search.cc` to ensure that SCANN does not run there. The failure happens because of the comparison like this one. 
```
REQUIRE(idx.Build(*train_ds, json) == knowhere::Status::success);
```
You need to skip the unit test if the returned result is `knowhere::Status::invalid_instruction_set` and add some message to the log as well.
So, something like
```
auto status = idx.Build(*train_ds, json);
if (status == knowhere::Status::invalid_instruction_set) {
    LOG("Skipping the test for SCANN, because it is not supported on this CPU architecture");
    return;
}

REQUIRE(status == knowhere::Status::success);
// continue the test
``` 

This change should be very straightforward and trivial.

liliu-z:
Hi @zhangwenlong8911 Thanks for contribution. We can help check in this code, but cannot promise it will always be well supported since we don't have CI for `loongarch64`. And can I ask the use case in your side to help us know the community better. Thanks!

PR #525, opened 2024-04-24
[fix] Add patch for bf16 distance computation

foxspy:
issue: #519 
/kind enhancement


PwzXxm:
/lgtm
for patch, revisit in Milvus 2.4 and later

PR #530, opened 2024-04-25
sparse float vector index to support mmap when loading from index file

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/29419

/kind: improvement


src/index/sparse/sparse_index_node.cc
```diff
@@ -194,12 +193,29 @@ class SparseInvertedIndexNode : public IndexNode {
...
+        auto reader = knowhere::FileReader(filename);
+        map_size_ = reader.size();
+        int map_flags = MAP_SHARED;
+        if (cfg.enable_mmap_pop.has_value() && cfg.enable_mmap_pop.value()) {
+#ifdef MAP_POPULATE
+            map_flags |= MAP_POPULATE;
+#endif
+        }
+        map_ = static_cast<char*>(mmap(nullptr, map_size_, PROT_READ, map_flags, reader.descriptor(), 0));
+        madvise(map_, map_size_, MADV_RANDOM);
```
alexanderguzhva, line 215: please check the results of `mmap` and `madvise` whether any of these two fails

src/index/sparse/sparse_index_node.cc
```diff
@@ -229,8 +245,29 @@ class SparseInvertedIndexNode : public IndexNode {
...
+    void
+    delete_index() {
+        if (index_ != nullptr) {
+            delete index_;
+            index_ = nullptr;
+        }
+        if (map_ != nullptr) {
+            auto res = munmap(map_, map_size_);
+            if (res != 0) {
+                throw std::runtime_error("failed to munmap when trying to delete index: " +
```
alexanderguzhva, line 257: this may cause an exception to be thrown inside `~SparseInvertedIndexNode()`

zhengbuqian: in the current impl seems the only thing we can do is to swallow the error? we provide no method for the users to explicitly destroy all the resources used so we can only clean in the destructor

PR #532, opened 2024-04-26
Bitset valid check

chasingegg:
Make sure the bitset is valid to avoid possible bugs when we deal with filtered search

### DIFF
--- a/src/index/index.cc
+++ b/src/index/index.cc
@@ -11,6 +11,7 @@
 
 #include "knowhere/index/index.h"
 
+#include "fmt/format.h"
 #include "knowhere/comp/time_recorder.h"
 #include "knowhere/dataset.h"
 #include "knowhere/expected.h"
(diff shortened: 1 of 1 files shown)


src/index/index.cc
```diff
@@ -116,6 +127,16 @@ Index<T>::AnnIterator(const DataSet& dataset, const Json& json, const BitsetView
     if (status != Status::success) {
         return expected<std::vector<std::shared_ptr<IndexNode::iterator>>>::Err(status, msg);
     }
+    // when index is immutable, bitset size should always equal to data count in index
+    // when index is mutable, it could happen that data count larger than bitset size, see
+    // https://github.com/zilliztech/knowhere/issues/70
+    // so something must be wrong at caller side when passed bitset size larger than data count
+    if (bitset_.size() > this->Count()) {
+        msg = fmt::format("bitset size invalid! bitset size: {}, data count: {}", bitset_.size(), this->Count());
```
alexanderguzhva, line 135: nit: I'd write in the error message that the bitset size is expected to be <= than the data size, just to make the error more clear. Otherwise, I'd assume (based on the current message) that these two sizes must be equal.

src/index/index.cc
```diff
@@ -140,6 +161,16 @@ Index<T>::RangeSearch(const DataSet& dataset, const Json& json, const BitsetView
     if (status != Status::success) {
         return expected<DataSetPtr>::Err(status, std::move(msg));
     }
+    // when index is immutable, bitset size should always equal to data count in index
+    // when index is mutable, it could happen that data count larger than bitset size, see
+    // https://github.com/zilliztech/knowhere/issues/70
+    // so something must be wrong at caller side when passed bitset size larger than data count
+    if (bitset_.size() > this->Count()) {
+        msg = fmt::format("bitset size invalid! bitset size: {}, data count: {}", bitset_.size(), this->Count());
```
alexanderguzhva, line 169: same

PR #535, opened 2024-04-28
Add AMX support to speed up Faiss Inner-Product

mellonyou:
Use Intel AMX to speed up Inner-Product algorithm of knowhere::BruteForce::Search(), which can bring more than 10x performance boost.

Build parameter: use "-o with_dnnl=True/False" to control enable/disable AMX feature.
This feature will depends on libdnnl.so.3, you can install it by running scripts/install_deps.sh.

Runtime parameter: if you want use AMX feature, you need set ENV parameter "DNNL_ENABLE=1" at first, otherwise the AMX feature will not work.


mellonyou:
issue: #541

mellonyou:
I can't edit the labels, need any access permissions?

thirdparty/faiss/faiss/utils/onednn_utils.h
```diff
@@ -0,0 +1,216 @@
...
+        f32_md2 = dnnl::memory::desc({yrow, ycol}, dnnl::memory::data_type::f32, dnnl::memory::format_tag::ab);
+        f32_dst_md2 = dnnl::memory::desc({xrow, yrow}, dnnl::memory::data_type::f32, dnnl::memory::format_tag::ab);
+
+        f32_mem1 = dnnl::memory(f32_md1, cpu_engine, in_f32_1);
+        f32_mem2 = dnnl::memory(f32_md2, cpu_engine, in_f32_2);
+        f32_dst_mem = dnnl::memory(f32_dst_md2, cpu_engine);
+
+        // inner memory bf16
+        bf16_md1 = dnnl::memory::desc({xrow, xcol}, dnnl::memory::data_type::bf16, dnnl::memory::format_tag::any);
+        bf16_md2 = dnnl::memory::desc({yrow, ycol}, dnnl::memory::data_type::bf16, dnnl::memory::format_tag::any);
```
liliu-z, line 166: Noob Q, why we use bf16 here?

thirdparty/faiss/faiss/utils/onednn_utils.h
```diff
@@ -0,0 +1,216 @@
...
+        if (this->in_f32_2 != in_f32_2)
+            BaseData::getState().store(BASE_DATA_STATE::MODIFIED);
+    }
+
+    void execut(float** out_f32) {
+        dnnl::reorder(f32_mem1, bf16_mem1).execute(engine_stream, f32_mem1, bf16_mem1);
+        BASE_DATA_STATE expected = BASE_DATA_STATE::MODIFIED;
+
+        if (BaseData::getState().compare_exchange_strong(expected, BASE_DATA_STATE::PREPARE)) {
+            pthread_rwlock_wrlock(&rwlock);
```
liliu-z, line 186: Noob Q, why we need to lock this. Is that because we only have only AMX instruction can run at a time?

thirdparty/faiss/faiss/utils/onednn_utils.h
```diff
@@ -0,0 +1,216 @@
...
+        // update state for new base data
+        if (this->in_f32_2 != in_f32_2)
+            BaseData::getState().store(BASE_DATA_STATE::MODIFIED);
+    }
+
+    void execut(float** out_f32) {
+        dnnl::reorder(f32_mem1, bf16_mem1).execute(engine_stream, f32_mem1, bf16_mem1);
+        BASE_DATA_STATE expected = BASE_DATA_STATE::MODIFIED;
+
+        if (BaseData::getState().compare_exchange_strong(expected, BASE_DATA_STATE::PREPARE)) {
```
liliu-z, line 185: Plz CMIIW. In the first call, `expected` will be `BASE_DATA_STATE::MODIFIED` and changed into `BASE_DATA_STATE::PREPARE` in this line and return false. Then it will loop in line 196 forever

thirdparty/faiss/faiss/utils/distances.cpp
```diff
@@ -211,30 +214,59 @@ void exhaustive_inner_product_seq_impl(
     using SingleResultHandler = typename BlockResultHandler::SingleResultHandler;
     int nt = std::min(int(nx), omp_get_max_threads());
 
+#ifdef FAISS_WITH_DNNL
+    if (is_dnnl_enabled()) {
+        float *res_arr = NULL;
+
+        comput_f32bf16f32_inner_product(nx, d, ny, d, const_cast<float*>(x), const_cast<float*>(y), &res_arr);
```
liliu-z, line 221: Can we implement a dynamic hook like all other simd in Knowhere?

thirdparty/faiss/faiss/utils/onednn_utils.h
```diff
@@ -0,0 +1,216 @@
...
+        inner_product_prim = dnnl::inner_product_forward(inner_product_pd);
+
+        bf16_mem1 = dnnl::memory(inner_product_pd.src_desc(), cpu_engine);
+
+        // update state for new base data
+        if (this->in_f32_2 != in_f32_2)
+            BaseData::getState().store(BASE_DATA_STATE::MODIFIED);
+    }
+
+    void execut(float** out_f32) {
```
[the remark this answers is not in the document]
mellonyou: yes, it's a typo

thirdparty/faiss/faiss/utils/onednn_utils.h
mellonyou: Because AMX can native support for bf16/int8 compute, which can significantly improve performance, and we have done the test, it have little impact on accuracy.

thirdparty/faiss/faiss/utils/onednn_utils.h
mellonyou: The lock is designed for multi-thread scenario, if two threads operate on the same base dataset with different query dataset, the lock prevent the base dataset from being modified by the other thread while working on it.

thirdparty/faiss/faiss/utils/onednn_utils.h
mellonyou: The state is also designed for multi-thread scenario, the state change is INIT->MODIFIED -> PREPARE -> READY. When the first thread have finished the initialization, the other thread will get the state is READY, and then skip line 196.

thirdparty/faiss/faiss/utils/distances.cpp
mellonyou: We have also considered following the other simd interface, but due to the implementation of AMX, it may be a bit incompatible with the current interface:
1. AMX prefers batch data calculation, and it's library will schedule multiple threads on its own.
2. The return value is a array for batch data operation.
So if we use dynamic hook, maybe need add new interface for batch data operation, and call the new interface when AMX is available.

mellonyou: @liliu-z We are planning to port code to adapt dynamic hook, do you have any other suggestions?

thirdparty/faiss/faiss/utils/distances.cpp
```diff
@@ -211,30 +214,59 @@ void exhaustive_inner_product_seq_impl(
     using SingleResultHandler = typename BlockResultHandler::SingleResultHandler;
     int nt = std::min(int(nx), omp_get_max_threads());
 
+#ifdef FAISS_WITH_DNNL
```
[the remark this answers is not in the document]
xtangxtang: @alexanderguzhva The filter is inside Knowhere or in the Milvus?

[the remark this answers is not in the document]
alexanderguzhva: @xtangxtang an external filter (in the form of bitset), provided from Milvus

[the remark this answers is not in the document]
xtangxtang: @alexanderguzhva Thanks for your clarification. Since this PR is just for the Knowhere, The Benchmark for this PR should include the bitset filter? or Just an memory bandwitdth perf data is OK?

mellonyou:
1) port the code to knowhere to follow dynamic hook interface.
2) about filter, write a simple benchmark to compare no filter amx inner product with simd inner product with filer(0.1f, 0.5f, 0.9f), it can be seen that AMX still has a perf. advantage when the filter percentage reaches 0.9.
percentage              0.1	          0.5	         0.9	            amx
result(s) 	               0.432	0.208   	0.043	   0.033
dnnl perf. boost     13.1x   	6.3x	         1.3x	 

For the code, the amx inner product interface is more suitable for producing batch vectors, and it doesn't support a filter interface, I have two ideas:
1) amx inner product just handle no filter scenario.
2) add a percentage parameter to the interface, when it is less than 0.9, we choose amx inner product.

@alexanderguzhva Looking forward to your suggestions.

alexanderguzhva:
@mellonyou Could you please include a benchmark or, at least, its details?
The numbers that you've provided cannot be interpreted properly without knowing
* the exact number of samples
* the dimensionality
* whether it is a single query/batched query requests
* is it a test for this particular function or for a whole index,
* etc.

The results are interesting and are definitely worth checking on my end.

mellonyou:
#include "simd/distances_onednn.h"

#define MAX_LOOP 20
TEST_CASE("Test Brute Force", "[float vector]") {
    using Catch::Approx;

    const int64_t nb = 2000000;
    const int64_t nq = 10;
    const int64_t dim = 512;
    const int64_t k = 100;

    auto metric = GENERATE(as<std::string>{}, knowhere::metric::IP );

    const auto train_ds = GenDataSet(nb, dim);
    const auto query_ds = CopyDataSet(train_ds, nq);

    const knowhere::Json conf = {
        {knowhere::meta::DIM, dim},
        {knowhere::meta::METRIC_TYPE, metric},
        {knowhere::meta::TOPK, k},
        {knowhere::meta::RADIUS, knowhere::IsMetricType(metric, knowhere::metric::IP) ? 10.0 : 0.99},
    };

    SECTION("Test Search Batch") {
     faiss::BaseData::getState().store(faiss::BASE_DATA_STATE::MODIFIED);
     struct timeval t1,t2;
     double timeuse;
     gettimeofday(&t1,NULL);

         std::vector<std::function<std::vector<uint8_t>(size_t, size_t)>> gen_bitset_funcs = {
                 GenerateBitsetWithFirstTbitsSet, GenerateBitsetWithRandomTbitsSet};
         const auto bitset_percentages = {0.1f, 0.5f, 0.9f};
         for (const float percentage : bitset_percentages) {
                 for (const auto& gen_func : gen_bitset_funcs) {
                         auto bitset_data = gen_func(nb, percentage * nb);
                         knowhere::BitsetView bitset(bitset_data.data(), nb);

                         for (int i = 0; i < MAX_LOOP; i++)
                         {
                                 gettimeofday(&t1,NULL);

                                 //    threads.emplace_back(WrapSearch, queryvar1);
                                 auto res = knowhere::BruteForce::Search<knowhere::fp32>(train_ds, query_ds, conf, bitset);
                                 gettimeofday(&t2,NULL);
                                 timeuse = (t2.tv_sec - t1.tv_sec) + (double)(t2.tv_usec - t1.tv_usec)/1000000.0;
                                 std::cout << "elpased: " << timeuse << std::endl;
                         }

                 }
         }

         gettimeofday(&t2,NULL);
         timeuse = (t2.tv_sec - t1.tv_sec) + (double)(t2.tv_usec - t1.tv_usec)/1000000.0;

         std::cout << "All thread finished." << std::endl;

        }
}

mellonyou:
@alexanderguzhva I just add this code to ut as a temporary benchmark, and build it with "-o with_dnnl=True", then run the test:
DNNL_ENABLE=0/1 ./Release/tests/ut/knowhere_tests
The test will run 20 rounds, and the results above are the average after discarding the best 20% and the worst 20%. And I ran the test on Intel SPR platform with Ubuntu 22.04 system.

alexanderguzhva:
@mellonyou I'll take a look. Thanks!

src/simd/distances_onednn.h
```diff
@@ -0,0 +1,218 @@
...
+        inner_product_prim = dnnl::inner_product_forward(inner_product_pd);
+
+        bf16_mem1 = dnnl::memory(inner_product_pd.src_desc(), cpu_engine);
+
+        // update state for new base data
+        if (this->in_f32_2 != in_f32_2)
+            BaseData::getState().store(BASE_DATA_STATE::MODIFIED);
+    }
+
+    void execut(float** out_f32) {
```
godchen0212, line 183: There is a typo.

mellonyou:
Add searchwithbuf and rangesearch interface implementation with AMX onednn. And will submit the related build config into milvus later.

mellonyou:
I am trying to do a manual filter with multithread before AMX IP.
@liliu-z @alexanderguzhva @godchen0212  Do you have any other opinions on the current interface implementation.

mellonyou:
Have tried to do manual filter with multithread before AMX IP, and it have a significant impact on performance. So we only filter the results to ensure their accuracy, which has a relatively small impact on performance.

PR #538, opened 2024-04-30
fix compiler warnings

alexanderguzhva:


thirdparty/faiss/faiss/utils/binary_distances.cpp
```diff
@@ -387,7 +387,14 @@ void binary_knn_hc(
         size_t all_heap_size = thread_heap_size * thread_max_num;
         T* value = new T[all_heap_size];
         int64_t* labels = new int64_t[all_heap_size];
-        T init_value = (typeid(T) == typeid(float)) ? (1.0 / 0.0) : 0x7fffffff;
+        
+        T init_value;
+        if constexpr (std::is_same_v<T, float>) {
+            init_value = std::numeric_limits<float>::quiet_NaN();
```
liliu-z, line 393: `1.0/0 = inf`
`std::numeric_limits<float>::quiet_NaN() = nan`
and `inf > [any float]` is ture but `nan > [any float]` if false.

It is tricky, let's do `std::numeric_limits<float>::max()`

liliu-z (changes requested):

thirdparty/faiss/faiss/utils/binary_distances.cpp
alexanderguzhva: Changed to `infinity()`

liliu-z (approved):
/lgtm

PR #542, opened 2024-05-06
Fix LAPACK not found

chasingegg:
after #448 , LAPACK dependency is limited to 'openblas' vendor and will not be found in some enviroments, and also find_package(LAPACK) will overwrite BLAS_LIBRARIES, move find_package(LAPACK) above and limit BLAS_LIBRARIES to openblas

/kind improvement


zhengbuqian:
this fix works for me, thanks!

Issue #546, opened 2024-05-07
Feature: Binary vector support


chasingegg:
get ready for binary vector support(some of the features already implemented, just make sure), users could insert and search binary vectors as float vectors do

- [ ] Binary FLAT/IVFFLAT/HNSW support
- [ ] Binary index range search
- [ ] Binary growing index
- [ ] Binary index iterator

PR #550, opened 2024-05-07
Binary vector from/to np.uint8 array in pyknowhere api 

chasingegg:
issue #546 

Currently binary vector dataset uses np.int32, change to np.uint8


alexanderguzhva:
do I get it correct that we don't need to maintain the backward compatibility for this particular case?

chasingegg:
> do I get it correct that we don't need to maintain the backward compatibility for this particular case?

I guess so, we never officially manage the pyknowhere release

PR #554, opened 2024-05-10
[Bugfix] Add patch for bf16 distance computation

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/519
/kind enhancement

### DIFF
--- a/tests/ut/utils.h
+++ b/tests/ut/utils.h
@@ -48,6 +48,17 @@ GenDataSet(int rows, int dim, int seed = 42) {
 }
 
 inline knowhere::DataSetPtr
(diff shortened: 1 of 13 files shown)


alexanderguzhva:
/lgtm
both gcc and clang optimize the code properly

tests/ut/utils.h
```diff
@@ -47,6 +47,17 @@ GenDataSet(int rows, int dim, int seed = 42) {
     return ds;
 }
 
+inline knowhere::DataSetPtr
+GenFloatDataSet(int rows, int dim, int seed = 42) {
```
chasingegg, line 51: duplicate?

PR #559, opened 2024-05-13
only search use GPU memory pool

Presburger:
GPU Index, only Search task can use Gpu MEMORY pool.
issue: #515


Presburger:
issue: #515

liliu-z (approved):

PR #562, opened 2024-05-13
Cherry-pick commit #3416 from Faiss baseline

alexanderguzhva:
https://github.com/facebookresearch/faiss/pull/3416

> The code generated for function fvec_L2sqr generated by OpenXL do not perform as good as the codes generated by gcc on Power. The macros to enable imprecise floating point operation don’t cover Power with OpenXL. This patch adds the OpenXL compiler options for the PowerPC macros to achieve better performance.


alexanderguzhva:
rerun ut

chasingegg (approved):
/lgtm

PR #569, opened 2024-05-15
refactor memory management

alexanderguzhva:
Replaced some `new T[]` with `std::make_unique<T[]>` to make the code safer.
/kind improvement

### DIFF
--- a/include/knowhere/dataset.h
+++ b/include/knowhere/dataset.h
@@ -21,6 +21,7 @@
 #include <variant>
 
 #include "comp/index_param.h"
+#include "knowhere/range_util.h"
 #include "knowhere/sparse_utils.h"
 
 namespace knowhere {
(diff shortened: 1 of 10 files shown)


include/knowhere/dataset.h
```diff
@@ -71,18 +72,46 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
...
+    void
+    SetIds(std::unique_ptr<long int[]>&& ids) {
+        static_assert(sizeof(long int) == sizeof(int64_t));
+
+        std::unique_lock lock(mutex_);
+        this->data_[meta::IDS] = Var(std::in_place_index<2>, reinterpret_cast<int64_t*>(ids.release()));
+    }
+
+    void
+    SetIds(std::unique_ptr<long long int[]>&& ids) {
```
liliu-z, line 108: We we don't simply accept a `std::unique_ptr<int64_t[]>&& ids` ?

alexanderguzhva: `int64_t` is different on linux and on macos, this was the root of the problem why I implemented it this way

include/knowhere/dataset.h
```diff
@@ -71,18 +72,46 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
         this->data_[meta::DISTANCE] = Var(std::in_place_index<0>, dis);
     }
```
liliu-z, line 73: We can remove these old raw pointer based API

liliu-z (approved):
/lgtm

PR #574, opened 2024-05-21
Fix kmeans init dead loop when dist is inf

### LINKED ISSUE #575: Bug: kmeans init dead loop when distance is inf
use random init for this case

chasingegg:
/kind bug

Close #575


foxspy (approved):
/lgtm

PR #599, opened 2024-05-24
remove seed_ef and use ef for hnsw iterator

zhengbuqian:
kind/improvement

issue: https://github.com/zilliztech/knowhere/issues/500

### DIFF
--- a/src/index/hnsw/hnsw_config.h
+++ b/src/index/hnsw/hnsw_config.h
@@ -30,7 +30,6 @@ class HnswConfig : public BaseConfig {
     CFG_INT M;
     CFG_INT efConstruction;
     CFG_INT ef;
-    CFG_INT seed_ef;
     CFG_INT overview_levels;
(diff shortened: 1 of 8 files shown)


zhengbuqian:
kind/improvement

issue: #500

src/index/hnsw/hnsw_config.h
```diff
@@ -45,11 +44,6 @@ class HnswConfig : public BaseConfig {
             .set_range(1, std::numeric_limits<CFG_INT::value_type>::max())
             .for_search()
```
alwayslove2013, line 45: seems that `ef` need for_iterator

alwayslove2013 (changes requested):

src/index/hnsw/hnsw_config.h
zhengbuqian: good catch!

zhengbuqian:
kind/improvement

issue: https://github.com/zilliztech/knowhere/issues/500

PR #601, opened 2024-05-24
Fix iterator based range search

zhengbuqian:
kind/bug
issue: #600

/hold

hold for #599 to merge first


zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/600

PR #632, opened 2024-06-04
Upgrade index version

hhy3:


zhengbuqian:
chatted with @foxspy , we'll not bump index version for cardinal hnsw sparse support. closing this

PR #659, opened 2024-06-20
Add following APIs to facilitate DataSet conversion

cydrain:
* ConvertFromDataTypeIfNeeded()
 * ConvertToDataTypeIfNeeded()

### DIFF
--- a/include/knowhere/utils.h
+++ b/include/knowhere/utils.h
@@ -127,6 +127,39 @@ data_type_conversion(const DataSet& src) {
     return des;
 }
 
+// Convert DataSet from DataType to float
+template <typename DataType>
+inline DataSetPtr
+ConvertFromDataTypeIfNeeded(const DataSetPtr ds) {
+    if constexpr (std::is_same_v<DataType, typename MockData<DataType>::type>) {
+        return ds;
+    } else {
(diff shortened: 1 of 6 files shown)


Presburger (approved):

include/knowhere/utils.h
```diff
@@ -127,6 +127,39 @@ data_type_conversion(const DataSet& src) {
...
+        return data_type_conversion<DataType, typename MockData<DataType>::type>(*ds);
+    }
+}
+
+// Convert DataSet from DataType to float
+template <typename DataType>
+inline std::shared_ptr<const DataSet>
+ConvertFromDataTypeIfNeeded(const DataSet& ds) {
+    if constexpr (std::is_same_v<DataType, typename MockData<DataType>::type>) {
+        return ds.shared_from_this();
```
zhengbuqian, line 146: this is very strange: we don't know who owns the dataset, how can we just create a shared ptr from it? This implicitly requires the passed in `ds` to be already managed by a shared ptr. This requirement is satisfied for now, but could lead to a hard to find bug in the future.

but this issue is not introduced by this PR, we can figure out a solution and put up a fix later.

cydrain: plan to file another PR to change all APIs' input parameter to shared_ptr to avoid this issue completely

PR #671, opened 2024-06-27
[range search] iterator stops only if exceeds radius several times consecutively.

alwayslove2013:
A new param - range_search_level, default 0.01
- iterator stops only if it consecutively exceeds the `radius` more than `num_next() * range_search_level` times.


foxspy (approved):
/lgtm

PR #674, opened 2024-07-01
remove additional metric_type definition from diskann_config

alwayslove2013:
just `BaseConfig` is ok.
https://github.com/zilliztech/knowhere/blob/f0c16f4e47ca7d1b90540d870fa2f6cb9f8048a7/include/knowhere/config.h#L649-L655


foxspy (approved):
/lgtm

PR #676, opened 2024-07-01
a theoretical fix for binary data type

alexanderguzhva:
Faiss `IndexBinary`-derived indices prohibit `dim % 8 != 0` use cases. Still, it is needed to correctly handle non-standard `dim` values inside Knowhere, just in hope that a Faiss might lift its restriction one day.

/kind improvement

### DIFF
--- a/src/index/flat/flat.cc
+++ b/src/index/flat/flat.cc
@@ -117,7 +117,8 @@ class FlatIndexNode : public IndexNode {
                         faiss::SearchParameters search_params;
                         search_params.sel = id_selector;
 
-                        index_->search(1, (const uint8_t*)x + index * dim / 8, k, cur_i_dis, cur_ids, &search_params);
+                        index_->search(1, (const uint8_t*)x + index * ((dim + 7) / 8), k, cur_i_dis, cur_ids,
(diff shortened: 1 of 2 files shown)


xiaofan-luan:
do we really need to?

xiaofan-luan:
I thought it make much more sense to align with byte.

chasingegg:
Currently milvus has constraints for dim % 8 == 0 on binary vectors, but this change shall not cause any problems

xiaofan-luan:
> Currently milvus has constraints for dim % 8 == 0 on binary vectors, but this change shall not cause any problems

yes.  i'm just wondering what will be use case for that.

alexanderguzhva:
the use case is the aesthetics of a code that handles the situation correctly, even if the situation is improbable :)

chasingegg:
Agree, we could be ready for possible future change(dim % 8 != 0) even if it is unlikely:) , at least the code is correct and more robust

> the use case is the aesthetics of a code that handles the situation correctly, even if the situation is improbable :)

chasingegg (approved):
/lgtm

Issue #687, opened 2024-07-04
sparse vector search to support BM25 metric


zhengbuqian:
when using such metric, the query vector is expected to be IDF vector, while the doc vector is expected to be TF vector. `k1`, `b` and `avgdl` of BM25 formula must also be provided.

For both index, k1 and b must be provided at build/load time. When using inverted index, k1 and b can be overridden at search time, but when using WAND, the same k1 and b must be used.

PR #688, opened 2024-07-04
Add bm25 metric type to inverted index/wand

zhengbuqian:
Also reduced number of sparse unit tests so it won't take too long.

/kind feature
issue: #687

### DIFF
--- a/src/common/comp/brute_force.cc
+++ b/src/common/comp/brute_force.cc
@@ -37,6 +37,30 @@ namespace knowhere {
 
 class BruteForceConfig : public BaseConfig {};
 
+namespace {
+
+template <typename T>
+expected<sparse::DocValueComputer<T>>
+GetDocValueComputer(const BruteForceConfig& cfg) {
+    if (IsMetricType(cfg.metric_type.value(), metric::IP)) {
+        return sparse::GetDocValueOriginalComputer<T>();
+    } else if (IsMetricType(cfg.metric_type.value(), metric::BM25)) {
+        if (!cfg.bm25_k1.has_value() || !cfg.bm25_b.has_value() || !cfg.bm25_avgdl.has_value()) {
+            return expected<sparse::DocValueComputer<T>>::Err(
+                Status::invalid_args, "bm25_k1, bm25_b, bm25_avgdl must be set when searching for bm25 metric");
+        }
+        auto k1 = cfg.bm25_k1.value();
+        auto b = cfg.bm25_b.value();
+        auto avgdl = cfg.bm25_avgdl.value();
+        return sparse::GetDocValueBM25Computer<T>(k1, b, avgdl);
+    } else {
+        return expected<sparse::DocValueComputer<T>>::Err(Status::invalid_metric_type,
+                                                          "metric type not supported for sparse vector");
+    }
+}
+
+}  // namespace
+
 template <typename DataType>
 expected<DataSetPtr>
 BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset, const Json& config,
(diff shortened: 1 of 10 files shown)


src/common/comp/brute_force.cc
```diff
@@ -37,6 +37,30 @@ namespace knowhere {
...
+expected<sparse::DocValueComputer<T>>
+GetDocValueComputer(const BruteForceConfig& cfg) {
+    if (IsMetricType(cfg.metric_type.value(), metric::IP)) {
+        return sparse::GetDocValueOriginalComputer<T>();
+    } else if (IsMetricType(cfg.metric_type.value(), metric::BM25)) {
+        if (!cfg.bm25_k1.has_value() || !cfg.bm25_b.has_value() || !cfg.bm25_avgdl.has_value()) {
+            return expected<sparse::DocValueComputer<T>>::Err(
+                Status::invalid_args, "bm25_k1, bm25_b, bm25_avgdl must be set when searching for bm25 metric");
+        }
+        auto k1 = cfg.bm25_k1.value();
```
alexanderguzhva, line 52: Please add constraints on possible value ranges for `k1`, `b` and `avgdl` as well. For example, `avgdl` should be non-zero, as far as I see from the code.
Also, there is a duplicate code down below that also needs such a check, is it possible to add something like 
`static bool CheckBM25Params(const BaseConfig& config) {...}` ?

src/common/comp/brute_force.cc
```diff
@@ -352,7 +385,14 @@ BruteForce::RangeSearch(const DataSetPtr base_dataset, const DataSetPtr query_da
...
                     }
-                    auto dist = cur_query->dot(xb_sparse[j]);
+                    float row_sum = 0;
+                    if (is_bm25) {
+                        for (size_t k = 0; k < xb_sparse[j].size(); ++k) {
+                            auto [d, v] = xb_sparse[j][k];
+                            row_sum += v;
+                        }
+                    }
+                    auto dist = cur_query->dot(xb_sparse[j], sparse_computer, row_sum);
```
alexanderguzhva, line 395: Please move this code inside `SparseRow` to have the caller to use it as follows:
```
float dist = 0;
if (is_bm25) {
    dist = cur_query->dot_sparse(xb_sparse[j], sparse_computer);
} else {
    dist = cur_query->dot(xb_sparse[j]);
}
```
The code from the change belongs to `SparseRow` class.

src/common/comp/brute_force.cc
```diff
@@ -500,7 +538,14 @@ BruteForce::SearchSparseWithBuf(const DataSetPtr base_dataset, const DataSetPtr
...
                 }
-                float dist = row.dot(base[j]);
+                float row_sum = 0;
+                if (is_bm25) {
+                    for (size_t k = 0; k < base[j].size(); ++k) {
+                        auto [d, v] = base[j][k];
+                        row_sum += v;
+                    }
+                }
+                float dist = row.dot(base[j], computer, row_sum);
```
alexanderguzhva, line 548: same

src/common/comp/brute_force.cc
```diff
@@ -699,7 +739,14 @@ BruteForce::AnnIterator<knowhere::sparse::SparseRow<float>>(const DataSetPtr bas
...
                     }
-                    auto dist = row.dot(base[j]);
+                    float row_sum = 0;
+                    if (is_bm25) {
+                        for (size_t k = 0; k < base[j].size(); ++k) {
+                            auto [d, v] = base[j][k];
+                            row_sum += v;
+                        }
+                    }
+                    auto dist = row.dot(base[j], computer, row_sum);
```
alexanderguzhva, line 749: same

src/common/comp/brute_force.cc
zhengbuqian: the value ranges are defined in `config.h` and automatically checked, for example:

```cpp
KNOWHERE_CONFIG_DECLARE_FIELD(bm25_k1)
    .allow_empty_without_default()
    .set_range(0.0, 3.0)
    .description("BM25 k1 to tune the term frequency scaling factor")
    .for_train_and_search()
    .for_iterator()
    .for_deserialize()
    .for_deserialize_from_file();
```

src/common/comp/brute_force.cc
zhengbuqian: the reason why I made the caller to compute and pass in the row sum: 

* we don't need to compute the row sum for IP metric, thus I don't want to make `row_sum` an always computed member of `SparseRow`. 
* if we want to do some "lazy compute row sum and store in SparseRow" thing, we need to protect from concurrent row sum update which might be an overkill? also `dot` is a const method which makes it harder
* with bm25 on index we can precompute and cache the row sums, so that in `InvertedIndex::refine_and_collect` we pass the cached row sum to `dot`
* in the case of brute force, the incoming Dataset is a const one and we cannot modify it(create/update cached row sum)

if we still want to have the `SparseRow` to own the row sum computation logic:

```cpp
// dot_sparse to accept an optional row sum:
float dot_sparse(other, computer, std::optional<T> other_row_sum = std::nullopt) const {
    T row_sum = other_sum.has_value() ? other_sum.value() : row_sum();
    ...
}

// computes row sum on the fly, does not cache
float row_sum() const {}
```

WDYT?

zhengbuqian (commented):
thanks Alex!

src/common/comp/brute_force.cc
alexanderguzhva: your arguments totally make sense.

PR #689, opened 2024-07-04
fix possible memory leak caused by iterator

zhengbuqian:
/kind bug
issue: https://github.com/zilliztech/knowhere/issues/690


zhengbuqian:
issue: #690

PR #711, opened 2024-07-17
fix knowhere-hnsw filter strategy

alwayslove2013:
- update alpha from filter_rate / 2 to filter_rate * 0.7
- set the initial value of acc_alpha from 0.0 to 1.0
- switch to brute force if results are less than k.

issue: https://github.com/milvus-io/milvus/issues/34548

results: recall++, qps--
image

### DIFF
--- a/thirdparty/hnswlib/hnswlib/hnswalg.h
+++ b/thirdparty/hnswlib/hnswlib/hnswalg.h
@@ -435,7 +435,7 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
             metric_hops++;
             metric_distance_computations += size;
         }
-        float kAlpha = bitset.filter_ratio() / 2.0f;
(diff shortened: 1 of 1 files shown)


thirdparty/hnswlib/hnswlib/hnswalg.h
```diff
@@ -1441,6 +1441,10 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
                 searchBaseLayerST<false, true>(currObj, query_data, std::max(ef, k), visited, bitset, feder_result);
         }
         std::vector<std::pair<dist_t, labeltype>> result;
+        // switch to brute-force when insufficient k
```
PwzXxm, line 1444: move L1444-1447 before L1443 plz

alwayslove2013: got it~

Issue #714, opened 2024-07-17
knowhere print out TraceID in a different format as Milvus


cydrain:
Build Milvus in Debug mode, then you can see TracdID is printed out in a different way between knowhere and Milvus.

```
[2024/07/17 17:52:34.610 +08:00] [DEBUG] [querynodev2/services.go:680] ["search segments..."] [traceID=c56dd6de3329e8baf581744b0660381b] [msgID=451204844043770863] [collectionID=451204844043370785] [channel=by-dev-rootcoord-dml_15_451204844043370785v0] [scope=Historical]
[2024/07/17 17:52:34.614 +08:00] [DEBUG] [segments/validate.go:50] ["read target partitions"] [traceID=c56dd6de3329e8baf581744b0660381b] [collectionID=451204844043370785] [partitionIDs="[451204844043370786]"]
[2024/07/17 17:52:34.614 +08:00] [DEBUG] [segments/segment.go:575] ["search segment..."] [traceID=c56dd6de3329e8baf581744b0660381b] [collectionID=451204844043370785] [segmentID=451204844043570801] [segmentType=Sealed] [withIndex=true]
I20240717 17:52:34.616413 317924 future_c.cpp:58] [SERVER][executor_set_thread_num][milvus] future executor setup cpu executor with thread num: 24
I20240717 17:52:34.625875 317928 index.cc:32] [KNOWHERE][LoadConfig][MILVUS_FUTURE_C] Search config dump: {"ef":16,"k":10,"metric_type":"COSINE","span_id":[163,228,15,76,126,44,218,97],"trace_flags":1,"trace_id":[197,109,214,222,51,41,232,186,245,129,116,75,6,96,56,27]}
```

PR #715, opened 2024-07-18
Encode & Decode traceID and spanID as hex string

cydrain:
Issue: #714

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -641,8 +641,8 @@ class BaseConfig : public Config {
     CFG_BOOL enable_mmap_pop;
     CFG_BOOL for_tuning;
     CFG_BOOL shuffle_build;
-    CFG_BYTES trace_id;
-    CFG_BYTES span_id;
+    CFG_STRING trace_id;
+    CFG_STRING span_id;
     CFG_INT trace_flags;
     CFG_MATERIALIZED_VIEW_SEARCH_INFO_TYPE materialized_view_search_info;
     CFG_STRING opt_fields_path;
--- a/include/knowhere/tracer.h
+++ b/include/knowhere/tracer.h
@@ -14,7 +14,6 @@
 #include <memory>
 #include <string>
 
-#include "knowhere/config.h"
 #include "opentelemetry/trace/provider.h"
 
 #define TRACE_SERVICE_KNOWHERE "knowhere"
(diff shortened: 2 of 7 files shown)


src/common/tracer.cc
```diff
@@ -143,12 +143,24 @@ EmptySpanID(const TraceContext* ctx) {
...
+    for (size_t i = 0; i < len; i++) {
+        ss << std::hex << std::setw(2) << std::setfill('0') << static_cast<int>(data[i]);
+    }
+    return ss.str();
+}
+
+std::string
+GetIDFromHexStr(const std::string& hexStr) {
+    std::stringstream ss;
+    for (size_t i = 0; i < hexStr.length(); i += 2) {
```
alexanderguzhva, line 158: are there any checks whether the input string is valid? What happens if I pass an input 'qqq'?

cydrain: this API is only used to decode tracdID and spanID, so I add no validation check

include/knowhere/config.h
```diff
@@ -641,8 +641,8 @@ class BaseConfig : public Config {
     CFG_BOOL enable_mmap_pop;
     CFG_BOOL for_tuning;
     CFG_BOOL shuffle_build;
-    CFG_BYTES trace_id;
-    CFG_BYTES span_id;
+    CFG_STRING trace_id;
```
foxspy, line 644: Why do we need to include trace in the parameter set? Logically, the index doesn't consume this parameter; it's just used for some chain tracing. Can we consider constructing the TraceContext externally and passing it through the interface, extending the parameters on the interface to achieve this?

src/common/tracer.cc
alexanderguzhva: I definitely would add at least 
`for (size_t i = 0; i < (hexStr.length() / 2) * 2; i += 2) {}`
because otherwise we may get a possible case of a hard-to-find segmentation fault error

include/knowhere/config.h
cydrain: Adding traceID and spanID into config is the most economical way to achieve this.
If change the interfaces for Build()/Search()/RangeSearch(), then pyknowhere API need update (if I remember correctly, pyknowhere API does not support default input parameter), and Milvus code also need change.

src/common/tracer.cc
cydrain: OK, I'll update my code

foxspy (approved):
/lgtm

foxspy:
/hold
wait 2.3.7 release

include/knowhere/tracer.h
```diff
@@ -62,7 +61,10 @@ EmptyTraceID(const TraceContext* ctx);
 bool
 EmptySpanID(const TraceContext* ctx);
 
-tracer::TraceContext
-GetTraceCtxFromCfg(const BaseConfig* cfg);
+std::string
+BytesToHexStr(const uint8_t* data, size_t len);
```
foxspy, line 65: hold until milvus 2.4.7 release

foxspy (changes requested):

foxspy (approved):
/lgtm

PR #717, opened 2024-07-18
[improvement] Add more cardinal index metrics

foxspy:
issue: #716 
/kind improvement


alexanderguzhva:
we don't have 2.4.7 Cardinal yet, do we?

PR #722, opened 2024-07-22
x86 fp16/bf16

cqy123456:


cqy123456:
Neon support is not complete yet in this pr, there will be a new one later.

PR #731, opened 2024-07-25
Prerequisites for FAISS HNSW

alexanderguzhva:
A collection of small tweaks that might be needed in the future.
* Adds another statistics for HNSW called `nhops`, which is a number of different nodes visited through a single query search
* AVX-512 code for `PQDecoder8`
* Binary heap utilities for data stored in `std::pair<dis,ids>` format
* Fix for `HNSW::upper_beam != 1` search use case
* AVX-512 code `HNSW::MinimaxHeap::pop_min`

Additionally, fixes a bug in `ScalarQuantizerCoder_avx512`

Corresponding baseline Faiss PR: https://github.com/facebookresearch/faiss/pull/3692

/kind improvement


thirdparty/faiss/faiss/utils/Heap.h
```diff
@@ -200,6 +201,110 @@ inline void maxheap_replace_top(
...
+    bh[i] = bh[k];
+}
+
+/** Pushes the element (val, ids) into the heap bh_val[0..k-2] and
+ * bh_ids[0..k-2].  on output the element at k-1 is defined.
+ */
+template <class C>
+inline void heap_push(
+        size_t k,
+        std::pair<typename C::T, typename C::TI>* bh,
```
liliu-z, line 249: Curios about why we need to add another function with `pair`

liliu-z (approved):
/lgtm

Issue #743, opened 2024-07-30
Introduce SQ6 quantization


alexanderguzhva:
SQ6 seems to be competitive with SQ8 in certain cases.
SQ6 seems to be definitely competitive if used together with a refining.

PR #744, opened 2024-07-30
improve SQ6 code for AVX512

alexanderguzhva:
Significantly improves SQ6 performance for AVX512.

Also, the new code contains two commented out version candidates which are even faster, but require a proper platform (`-mbmi2` for Intel or AMD Zen4+ CPU families). TODO.

issue: #743 
/kind improvement

### DIFF
--- a/cmake/libs/libfaiss.cmake
+++ b/cmake/libs/libfaiss.cmake
@@ -96,6 +96,7 @@ if(__X86_64)
             -mavx512f
             -mavx512dq
             -mavx512bw
+            -mavx512vl
             -mpopcnt>)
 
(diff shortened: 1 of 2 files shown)


liliu-z:
Looks like the e2e was timeout. Just triggered the rerun.

Plz correct me if I am wrong, Milvus suer cannot benefit from this for now right. Are this for the preparation of HNSW migration?

alexanderguzhva:
@liliu-z yes
On the other hand, this PR opens an opportunity for someone else to add an IVF_SQ6 index independently

liliu-z (approved):
/lgtm

PR #746, opened 2024-08-01
FAISS HNSW

alexanderguzhva:
Introducing modified FAISS HNSW version, which outperforms current hnswlib HNSW implementation.
* HNSW,Flat for fp32.
* HNSW,SQ + optional refine. 
* * SQ6, SQ8, BF16, FP16 for sq. 
* * SQ8, FP16, BF16, FP32/FLAT for refine.
* HNSW,PQ + optional refine. 
* * TODO: bring speedup code. 
* * SQ8, FP16, BF16, FP32/FLAT for refine.
* HNSW,PRQ + optional refine.
* * TODO: bring speedup code.
* * SQ8, FP16, BF16, FP32/FLAT for refine.
* * The question is whether I need to keep this, because it provides some recall improvements at the cost of the training speed while keeping the same size and QPS.

The code is somewhat dirty and is subject to further possibly significant changes.

/kind improvement


src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1002 @@
...
+                        } else if (whether_bf_search) {
+                            // use brute-force wrapper
+                            base_wrapper = &wrapper_bf;
+                        } else {
+                            // use hnsw-search wrapper
+                            base_wrapper = &wrapper_hnsw_search;
+                        }
+
+                        // check if used wants a refined result
+                        if (hnsw_cfg.refine_k.has_value()) {
```
liliu-z, line 411: Milvus allows user to not provide any params, meaning any param we need to define its by default performance.

For this one: we set the `refine_k` as empty by default, and it means that the default performance for a user is not refine.

Do you think giving an experience number (can be ratio to k) to this `refine_k` to the refine as default is more intuitive? Also, we can define `refine_k = 0` as skipping refine, or even give a `skip_refine` param to make it more clear.

liliu-z (commented):
Just a first glance comments. More detailed review will follow.

Can I ask some summary about the code structure for faster reviewing? 
1. What is the Wrappers and why we need them
2. The relationship between Wrappers in Knowhere & Faiss
3. Difference between the overrided HNSW and Faiss'

Thanks

include/knowhere/index/index_table.h
```diff
@@ -66,6 +66,10 @@ static std::set<std::pair<std::string, VecType>> legal_knowhere_index = {
     {IndexEnum::INDEX_HNSW_SQ8_REFINE, VecType::VECTOR_FLOAT},
     {IndexEnum::INDEX_HNSW_SQ8_REFINE, VecType::VECTOR_FLOAT16},
     {IndexEnum::INDEX_HNSW_SQ8_REFINE, VecType::VECTOR_BFLOAT16},
+    // faiss hnsw
+    {IndexEnum::INDEX_FAISS_HNSW_FLAT, VecType::VECTOR_FLOAT},
+
+    {IndexEnum::INDEX_FAISS_HNSW_SQ, VecType::VECTOR_FLOAT},
```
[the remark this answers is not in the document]
alexanderguzhva: this is temporary

include/knowhere/comp/index_param.h
```diff
@@ -53,6 +53,10 @@ constexpr const char* INDEX_HNSW_SQ8 = "HNSW_SQ8";
 constexpr const char* INDEX_HNSW_SQ8_REFINE = "HNSW_SQ8_REFINE";
 constexpr const char* INDEX_DISKANN = "DISKANN";
 
+constexpr const char* INDEX_FAISS_HNSW_FLAT = "FAISS_HNSW_FLAT";
+constexpr const char* INDEX_FAISS_HNSW_SQ = "FAISS_HNSW_SQ";
+constexpr const char* INDEX_FAISS_HNSW_PQ = "FAISS_HNSW_PQ";
+
```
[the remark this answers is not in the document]
alexanderguzhva: not before we make sure that baseline HNSW from hnswlib canbe safely and fully replaced

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1002 @@
...
+    virtual Status
+    TrainInternal(const DataSetPtr dataset, const Config& cfg) = 0;
+
+    // add impl
+    virtual Status
+    AddInternal(const DataSetPtr dataset, const Config& cfg) = 0;
+};
+
+//
+class BaseFaissRegularIndexNode : public BaseFaissIndexNode {
```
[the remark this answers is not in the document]
alexanderguzhva: 1. There will be `BaseFaissBinaryIndexNode`
2. This is the goal. I wanted to get rid of `if constexpr()` spoiled ivf.cc

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: This is the draft version. Someone who uses the system needs to let me know about the best way to expose params, the default way that Milvus users prefer.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1002 @@
...
+
+//
+class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
+ public:
+    BaseFaissRegularIndexHNSWNode(const int32_t& version, const Object& object)
+        : BaseFaissRegularIndexNode(version, object) {
+    }
+
+    bool
+    HasRawData(const std::string& metric_type) const override {
```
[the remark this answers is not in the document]
alexanderguzhva: no, there will be `BaseFaissBinaryIndexNode`. This is a draft, anyway.

include/knowhere/comp/index_param.h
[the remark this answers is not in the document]
liliu-z: Since this index type name is external, we use config instead of the name itself as the fallback method to avoid compatibility trouble in the future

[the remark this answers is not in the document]
alexanderguzhva: this is a minor issue anyways. I need to make the code fully operational, and then we may consider changing all these strings.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1398 @@
...
+            final_index = std::move(final_index_cnd.value());
+        } else {
+            // no refine
+
+            // assign
+            final_index = std::move(hnsw_index);
+        }
+
+        // we have to convert the data to float, unfortunately, which costs extra RAM
+        auto float_ds_ptr = convert_ds_to_float(dataset, data_format);
```
liliu-z, line 1273: Does FAISS have a plan to support FP/BF16?

liliu-z (commented):
Still work in progress

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: We can ask Matthijs about supporting fp16 / bf16, but I doubt that he will go for it. So, I'm using a workaround for now.

alexanderguzhva:
> Can I ask why we need a have a binary class instead of templating it
@liliu-z because templating is inappropriate for this situation. Inheritance should be used for this particular use case.

include/knowhere/utils.h
```diff
@@ -109,44 +109,69 @@ GetKey(const std::string& name) {
...
+// * no start, no count, float -> returns the source without cloning
+// * no start, no count, no float -> returns a clone with a different type
+// * start, no count -> returns a clone that starts from a given row 'start'
+// * no start, count -> returns a clone that starts from a row 0 and has 'count' rows
+// * start, count -> returns a clone that start from a given row 'start' and has 'count' rows
+// * invalid start, count values -> returns nullptr
 template <typename DataType>
 inline DataSetPtr
-ConvertFromDataTypeIfNeeded(const DataSetPtr ds) {
+ConvertFromDataTypeIfNeeded(const DataSetPtr& ds, const std::optional<int64_t> start = std::nullopt,
```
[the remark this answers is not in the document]
alexanderguzhva: I think I've added this function for some code that I've deleted...

src/index/hnsw/faiss_hnsw_config.h
```diff
@@ -0,0 +1,250 @@
...
+#include "knowhere/comp/index_param.h"
+#include "knowhere/config.h"
+
+namespace knowhere {
+
+namespace {
+
+constexpr const CFG_INT::value_type kIteratorSeedEf = 40;
+constexpr const CFG_INT::value_type kEfMinValue = 16;
+constexpr const CFG_INT::value_type kDefaultRangeSearchEf = 16;
```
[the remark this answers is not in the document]
alexanderguzhva: Updated

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1398 @@
...
+    if (!is_fp32_flat.has_value()) {
+        return expected<std::unique_ptr<faiss::Index>>::Err(Status::invalid_args, "");
+    }
+
+    const bool is_fp32_flat_v = is_fp32_flat.value();
+
+    // check input data_format
+    if (data_format == DataFormatEnum::fp16) {
+        // make sure that we're using fp16 refine
+        auto refine_sq_type = get_sq_quantizer_type(refine_type.value());
```
liliu-z, line 838: Can we put this check to config? Milvus will check parameter validation through a CGO call to Knowhere in proxy for:
1. Early termination for bad request.
2. Build is an async request, it will be great to prevent bad request at beginning.

liliu-z: We can put it in `CheckAndAdjust`

src/index/hnsw/faiss_hnsw_config.h
```diff
@@ -0,0 +1,250 @@
...
+                return Status::invalid_value_in_json;
+            }
+
+            // check refine
+            if (refine_type.has_value()) {
+                if (!WhetherAcceptableRefineType(refine_type.value())) {
+                    *err_msg = "invalid refine type type";
+                    LOG_KNOWHERE_ERROR_ << *err_msg;
+                    return Status::invalid_value_in_json;
+                }
```
liliu-z, line 196: Can we create a new `.set_range()` function for this. This table lookup check looks very common used

src/index/hnsw/faiss_hnsw.cc
liliu-z, line 838: BTW what if this `refine_type` is `nullopt`? From the config definition, it is allowed to be empty

src/index/hnsw/impl/IndexBruteForceWrapper.cc
```diff
@@ -0,0 +1,80 @@
...
+
+        // try knowhere-specific filter
+        const knowhere::BitsetViewIDSelector* bw_idselector = dynamic_cast<const knowhere::BitsetViewIDSelector*>(sel);
+
+        knowhere::BitsetFilter filter(bw_idselector->bitset_view);
+
+        if (is_similarity_metric(index->metric_type)) {
+            using C = faiss::CMin<float, idx_t>;
+
+            faiss::cppcontrib::knowhere::brute_force_search_impl<C, faiss::DistanceComputer, knowhere::BitsetFilter>(
```
liliu-z, line 69: Did we consider reuse this? https://github.com/liliu-z/knowhere/blob/2006b77c5b005a7cacdaaa1d567be363d44ef55e/src/common/comp/brute_force.cc

liliu-z: Or merge them somehow

src/index/hnsw/impl/BitsetFilter.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <faiss/MetricType.h>
+
+#include "knowhere/bitsetview.h"
+
+namespace knowhere {
+
+// specialized override for knowhere
+struct BitsetFilter {
```
liliu-z, line 21: A noob Q. Why we need `BitsetFilter` and `DefaultIDSelectorFilter` instead of using `Selector` directly

liliu-z:
> > Can I ask why we need a have a binary class instead of templating it
> > @liliu-z because templating is inappropriate for this situation. Inheritance should be used for this particular use case.

Can you help elaborate this a little bit? Thanks!

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
```diff
@@ -0,0 +1,415 @@
...
+
+// Accomodates all the search logic and variables.
+/// * DistanceComputerT is responsible for computing distances
+/// * GraphVisitorT records visited edges
+/// * VisitedT is responsible for tracking visited nodes
+/// * FilterT is resposible for filtering unneeded nodes
+/// Interfaces of all templates are tweaked to accept standard Faiss structures
+///   with dynamic dispatching. Custom Knowhere structures are also accepted.
+template<typename DistanceComputerT, typename GraphVisitorT, typename VisitedT, typename FilterT>
+struct v2_hnsw_searcher {
```
liliu-z, line 53: Plz correct me if I am wrong, this is the same logic as what Knowhere' hnsw is right now

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1398 @@
...
+        if (!this->index->is_trained) {
+            return expected<DataSetPtr>::Err(Status::index_not_trained, "index not trained");
+        }
+
+        auto dim = Dim();
+        auto rows = dataset->GetRows();
+        auto ids = dataset->GetIds();
+
+        try {
+            auto data = std::make_unique<float[]>(dim * rows);
```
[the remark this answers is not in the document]
alexanderguzhva: Faiss always produces `float` output data in `index->reconstruct()`.

Do I get it correct that `GetVectorByIds()` method expects `fp16` output data for `fp16` version and `bf16` output data for `bf16` version? If so, then I need to add a data conversion.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1398 @@
...
+
+                    if (index_refine != nullptr) {
+                        // yes, it is possible to refine results.
+
+                        // cast a base index to IndexHNSW-based index
+                        faiss::IndexHNSW* const index_hnsw = dynamic_cast<faiss::IndexHNSW*>(index_refine->base_index);
+
+                        if (index_hnsw == nullptr) {
+                            // this is unexpected
+                            throw std::runtime_error("Expecting faiss::IndexHNSW");
```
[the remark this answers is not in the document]
alexanderguzhva: We cannot, because the exception is thrown inside one of folly's threads. 
Realistically speaking, this is a safe-guard error, similar to assert, which should never get triggered under normal conditions.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1398 @@
...
+
+        // create an index
+        const bool is_cosine = IsMetricType(hnsw_cfg.metric_type.value(), metric::COSINE);
+
+        std::unique_ptr<faiss::IndexHNSW> hnsw_index;
+        if (is_cosine) {
+            if (data_format == DataFormatEnum::fp32) {
+                hnsw_index = std::make_unique<faiss::IndexHNSWFlatCosine>(dim, hnsw_cfg.M.value());
+            } else if (data_format == DataFormatEnum::fp16) {
+                hnsw_index = std::make_unique<faiss::IndexHNSWSQCosine>(dim, faiss::ScalarQuantizer::QT_fp16,
```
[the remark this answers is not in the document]
alexanderguzhva: Yes and no.
Adding data into index consists of two steps:
1. training an index
2. adding the data.

Training HNSW, HNSWFlat, HNSWSQ(fp16) and HNSWSQ(bf16) does literally nothing. So, line [763](https://github.com/zilliztech/knowhere/blob/094fa745ef20472ff6f4c7a5fc5f79a81be0c44a/src/index/hnsw/faiss_hnsw.cc#L763)
`hnsw_index->train(rows, (const float*)data);` does literally nothing and is just a formality, as mentioned in the comment on lines 758-762.
 
Adding the data is performed via `add_to_index()` function on [line 346](https://github.com/zilliztech/knowhere/blob/094fa745ef20472ff6f4c7a5fc5f79a81be0c44a/src/index/hnsw/faiss_hnsw.cc#L346). If the input data format is not `fp32`, then it will split the input data into virtual chunks of `65536` lines ([line 356](https://github.com/zilliztech/knowhere/blob/094fa745ef20472ff6f4c7a5fc5f79a81be0c44a/src/index/hnsw/faiss_hnsw.cc#L356)). Every chunk will be converted into fp32 and added into an index. So, the overhead is `65536 * dim` bytes. If you believe that this is a large value, then please let me know what a better value should be. `16384`? `8192`? even `4096` maybe? `4096` for a `1536 dim` OpenAI dataset is just extra `24 MB`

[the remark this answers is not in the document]
alexanderguzhva: Also, this memory overconsumption is a problem for HNSWPQ and HNSWPRQ at this moment, because Faiss processed everything in floats. But there is a relatively simple way to fix it in the future because Faiss has all the needed things in place, thus trading training speed for the memory, if needed (we can make it a config parameter).

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: Makes sense

alexanderguzhva: A valid point.

src/index/hnsw/impl/BitsetFilter.h
alexanderguzhva: `Selector` is based on `virtual` methods. Technically, I haven't checked whether a compiler de-virtualized such calls. I need to check this.

alexanderguzhva:
> > > Can I ask why we need a have a binary class instead of templating it
> > > @liliu-z because templating is inappropriate for this situation. Inheritance should be used for this particular use case.
> 
> Could you please help elaborate this a little bit? Thanks!



> > > Can I ask why we need a have a binary class instead of templating it
> > > @liliu-z because templating is inappropriate for this situation. Inheritance should be used for this particular use case.
> 
> Could you please help elaborate this a little bit? Thanks!

Basically, templatizing is used in C++ for particular use cases, which are targeted at performance. In our particular case, I consider templatizing `IndexNode` as a definite design error. It provides zero benefits, leads to generating the same code multiple times and overcomplicates the code structure. Traditional OOP approach needs to be used here.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
alexanderguzhva: Originally, hnswlib contained a code for a brute-force search. So, I kinda reimplemented it. My goal was to make sure that the QPS value matches.

Reusing what's available in Faiss needs to be evaluated.

src/index/hnsw/impl/IndexBruteForceWrapper.cc
alexanderguzhva: This needs to be evaluated, no doubt.

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
alexanderguzhva: changed to 4096

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
alexanderguzhva: modified `GetVectorByIds()` accordingly

src/index/hnsw/impl/BitsetFilter.h
alexanderguzhva: updated. I had to use a small hack, because `GCC` seems to have problems devirtualizing. There were no problems with `clang`.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
alexanderguzhva: technically, I'm not putting things in Faiss. My future goal is to use a vanilla Faiss and have a dedicated set of additional components that belong to knowhere. I'd like to keep these Knowhere components in `faiss/cppcontrib/knowhere` directory.

liorf95:
How is it possible to run the benchmark /UT for this PR?

alexanderguzhva:
@liorf95 We have benchmarks that are done in https://github.com/zilliztech/vdbbench, they will be published as soon as the code is reviewed. Basically, at this moment:
* Faiss HNSW Flat outperforms our current hnswlib Flat up to 20%, depending on the filtering rate
* Faiss HNSW SQ8 is slower than hnswlib SQ8, but the recall is higher, because of difference in SDC / ADC operating modes
* Faiss HNSW PQ and especially Faiss HNSW PRQ are not at theirs peak performance, because they require some additional components that are in a beta stage, plus some additional but tiny changes. This will be done as well. But overall, the QPS for HNSW PQ and HNSW PRQ is somewhat close to HNSW Flat / HNSW SQ

Alternatively, if you'd like to test the code yourself, please use vdbbench. I'm just not sure whether we've added a support to Milvus yet.

liorf95:
> @liorf95 We have benchmarks that are done in https://github.com/zilliztech/vdbbench, they will be published as soon as the code is reviewed. Basically, at this moment:
> 
> * Faiss HNSW Flat outperforms our current hnswlib Flat up to 20%, depending on the filtering rate
> * Faiss HNSW SQ8 is slower than hnswlib SQ8, but the recall is higher, because of difference in SDC / ADC operating modes
> * Faiss HNSW PQ and especially Faiss HNSW PRQ are not at theirs peak performance, because they require some additional components that are in a beta stage, plus some additional but tiny changes. This will be done as well. But overall, the QPS for HNSW PQ and HNSW PRQ is somewhat close to HNSW Flat / HNSW SQ
> 
> Alternatively, if you'd like to test the code yourself, please use `vdbbench`, but I'm not sure tbh whether it might require some manual changes in `vdbbench` (I don't remember tbh lol). I'm just not sure whether we've added a support to Milvus yet.
> 
> hnswlib flat for 768 dim 1M dataset (current)
> 
> ```
> 
> 2024-08-20 20:37:45,815 | INFO |DB       | db_label case            label      | load_dur    qps          latency(p99)    recall        max_load_count | label (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |-------- | -------- --------------- ---------- | ----------- ------------ --------------- ------------- -------------- | ----- (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-None     2024082020 | 665.2817    4665.1736    0.0             0.9214        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-0p   2024082020 | 665.2817    4655.2296    0.0             0.9214        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-1p   2024082020 | 665.2817    4512.0473    0.0             0.9208        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-5p   2024082020 | 665.2817    4595.7376    0.0             0.9152        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-10p  2024082020 | 665.2817    4683.7024    0.0             0.9105        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-20p  2024082020 | 665.2817    4778.249     0.0             0.8996        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-30p  2024082020 | 665.2817    4779.8595    0.0             0.8886        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-40p  2024082020 | 665.2817    4639.7362    0.0             0.8801        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-50p  2024082020 | 665.2817    4296.0571    0.0             0.8759        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-60p  2024082020 | 665.2817    3753.4411    0.0             0.878         0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-70p  2024082020 | 665.2817    2985.7408    0.0             0.8844        0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-80p  2024082020 | 665.2817    2078.155     0.0             0.901         0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-90p  2024082020 | 665.2817    1354.3858    0.0             0.891         0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-95p  2024082020 | 665.2817    479.1046     0.0             1.0           0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-98p  2024082020 | 665.2817    2098.4618    0.0             1.0           0              | :)    (models.py:234)
> 2024-08-20 20:37:45,815 | INFO |Knowhere |          Cohere-int-99p  2024082020 | 665.2817    3511.0782    0.0             1.0           0              | :)    (models.py:234)
> ```
> 
> faiss hnsw flat for 768 dim 1M dataset (candidate)
> 
> ```
> 2024-08-20 20:16:55,630 | INFO |DB       | db_label case            label      | load_dur    qps          latency(p99)    recall        max_load_count | label (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |-------- | -------- --------------- ---------- | ----------- ------------ --------------- ------------- -------------- | ----- (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-None     2024082019 | 627.703     5017.7925    0.0             0.9265        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-0p   2024082019 | 627.703     5012.7896    0.0             0.9265        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-1p   2024082019 | 627.703     4955.4562    0.0             0.9253        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-5p   2024082019 | 627.703     5085.4225    0.0             0.9198        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-10p  2024082019 | 627.703     5265.8881    0.0             0.9139        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-20p  2024082019 | 627.703     5485.4166    0.0             0.9028        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-30p  2024082019 | 627.703     5575.8401    0.0             0.8906        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-40p  2024082019 | 627.703     5440.1608    0.0             0.883         0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-50p  2024082019 | 627.703     5054.3116    0.0             0.8799        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-60p  2024082019 | 627.703     4426.2809    0.0             0.8803        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-70p  2024082019 | 627.703     3552.6454    0.0             0.8896        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-80p  2024082019 | 627.703     2510.4212    0.0             0.9021        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-90p  2024082019 | 627.703     1665.417     0.0             0.8952        0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-95p  2024082019 | 627.703     474.7159     0.0             1.0           0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-98p  2024082019 | 627.703     2242.8977    0.0             1.0           0              | :)    (models.py:234)
> 2024-08-20 20:16:55,637 | INFO |Knowhere |          Cohere-int-99p  2024082019 | 627.703     3969.9828    0.0             1.0           0              | :)    (models.py:234)
> ```

Wow!Thanks a lot for all that helpful information.
I did not see any additional UT for that PR in tests/ut/knowhere_tests - had I missed anything here?

alexanderguzhva:
@liorf95 `benchmark/hdf5/benchmark_faiss_hnsw.cpp`. The reason why it is done this way because it introduces way too many options to test. As of now, this benchmark generates 2.7 GB of temporary data and tests all new possible combinations of indices. You may wish to quickly tweak it for your own needs.

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
liliu-z: IC, should we take care of the folly thread failures?
What I usually do is avoiding throw exception in a folly thread(since the thread will get killed), and set a status that can be verified out side of the folly thread?

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
liliu-z: IC, so after calling `add_to_index()`, FAISS HNSWSQ(BF16) will only store BF16 data right?

Also, can you help elaborate a little bit why HNSWPQ/PRQ need more memory. Actually building memory is not that critical, but search is

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
liliu-z: > technically, I'm not putting things in Faiss. My future goal is to use a vanilla Faiss and have a dedicated set of additional components that belong to knowhere. I'd like to keep these Knowhere components in `faiss/cppcontrib/knowhere` directory.

OK, it means that this implementation is from Knowhere's HNSW. To my understanding, this implementation is some kind of temporary fall back plan and need to be removed after the main implementation is stable. So let it be where it used to be (knowhere/thirdparty/hnswlib) can make the implementation easier? 

Just curios about any concerns here, I don't have a strong opinion to put it anywhere.

src/index/hnsw/impl/IndexWrapperCosine.cc
```diff
@@ -0,0 +1,29 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#include "index/hnsw/impl/IndexWrapperCosine.h"
+
+#include <faiss/IndexCosine.h>
+
+namespace knowhere {
+
+// a wrapper that overrides a distance computer
+IndexWrapperCosine::IndexWrapperCosine(faiss::Index* index, const float* inverse_l2_norms_in)
```
liliu-z, line 19: Why we need a wrapper instead of putting logics in `IndexHNSWxxxxCosine`?

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
alexanderguzhva: this is exactly the situation there a thread needs to be killed and report an exception.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1480 @@
...
+                auto data = std::make_unique<knowhere::fp16[]>(dim * rows);
+
+                // faiss produces fp32 data format, we need some other format.
+                // Let's create a temporary fp32 buffer for this.
+                auto tmp = std::make_unique<float[]>(dim);
+
+                for (int64_t i = 0; i < rows; i++) {
+                    const int64_t id = ids[i];
+                    assert(id >= 0 && id < index->ntotal);
+                    index->reconstruct(id, tmp.get());
```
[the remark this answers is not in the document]
alexanderguzhva: if HNSWSQ is used and no refinement is available, then the reconstruction will be inaccurate.
Please let me know the requirements for this particular method. What is it expected to return exactly?

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -0,0 +1,1480 @@
...
+
+                    if (!convert_rows_from_fp32(tmp.get(), data.get(), data_format, i, 1, dim)) {
```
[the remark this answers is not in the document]
alexanderguzhva: I haven't checked that in a brute-force validation (65536 values for FP16 to check), but the logic tells me that there should be zero loss, so it must be safe to convert from FP16 -> FP32 -> FP16 and have the same result.

include/knowhere/tolower.h
```diff
@@ -0,0 +1,29 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+#pragma once
+
+#include <algorithm>
+#include <string>
+
+namespace knowhere {
```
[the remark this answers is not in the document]
alexanderguzhva: I don't like `utils.h` style files which are bloated with 100500 include headers :)

src/index/hnsw/impl/IndexBruteForceWrapper.cc
```diff
@@ -0,0 +1,110 @@
...
+
+#include "knowhere/bitsetview.h"
+#include "knowhere/bitsetview_idselector.h"
+
+namespace knowhere {
+
+using idx_t = faiss::idx_t;
+
+// the following structure is a hack, because GCC cannot properly
+//   de-virtualize a plain BitsetViewIDSelector.
```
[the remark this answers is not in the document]
alexanderguzhva: the performance for selector impact may have an impact in brture force search, because it implies 90+% of filtering. In other cases, the cost of calling a virtual method is negligible compares to the cost of distance computations.

src/index/hnsw/impl/IndexWrapperCosine.cc
alexanderguzhva: Ultimately, it was needed for refine and to enable/disable refine for the search wirh a config parameter.

thirdparty/faiss/faiss/cppcontrib/knowhere/IndexBruteForceWrapper.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <faiss/impl/FaissAssert.h>
+#include <faiss/impl/IDSelector.h>
+
+#include <faiss/cppcontrib/knowhere/impl/Bruteforce.h>
+
+namespace faiss {
+namespace cppcontrib {
+namespace knowhere {
+
+IndexBruteForceWrapper::IndexBruteForceWrapper(Index* underlying_index) :
```
liliu-z, line 30: Why we have two BruteForce Wrappers? one in Knowhere and one in Faiss. And looks like nowhere call this one

alexanderguzhva: I expect to merge one, a generic one into Faiss, if they accept it. Another one is a specialized version for our needs. The other one might be removed in Faiss guys reject it.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
alexanderguzhva: We can always change locations later, this is not important at this moment.

src/index/hnsw/faiss_hnsw_config.h
alexanderguzhva: I'd rather keep it as is for now

src/index/hnsw/impl/IndexBruteForceWrapper.cc
alexanderguzhva: This implementation uses `faiss::DistanceComputer` for distances

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
liliu-z: From the code, this error will kill the folly thread, and exception will throw from `WaitAllSuccess`. Then the catch block outside will catch this error and return an error code to Milvus. User will get error report, do some modification and then retry. But the folly thread is permenentally dead.

1. If this is an unexpected error, we need to crash the Milvus
2. If it is a fixable problem (mis operation order, retryable), then we should not kill the thread.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HnswSearcher.h
liliu-z: More than location. 
1. Since we have three HNSW for now: The new implemented one, /knohwere/thirdpart/hnswlib, and this searcher.
2. Given the purpose is fallback to old solution, old code can always provide more promising old solution

It is a little bit hard for keep them updated (We may need to do so for temporary).

I will create a issue to follow and unblock this PR. We can have a further discussion there.

liliu-z: https://github.com/zilliztech/knowhere/issues/776

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
liliu-z: Created a issue https://github.com/zilliztech/knowhere/issues/777 to track this and unblock this PR

src/index/hnsw/faiss_hnsw.cc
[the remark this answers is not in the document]
liliu-z: `GetVectorById` need to return an error code if there is no raw data available (If the data type is FP16, then the raw data need to be FP16).

So we need a raw data existence check there.

include/knowhere/tolower.h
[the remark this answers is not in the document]
liliu-z: hhhhh,  why 100500

src/index/hnsw/impl/IndexBruteForceWrapper.cc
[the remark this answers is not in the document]
liliu-z: Saying that it is also a concern.
1. IVF can have super high filtering rate but still uses virtual method.
2. To my experience the, HNSW validation is not negligible. We noticed 10% performance regression in cardinal if we try to do one more deference for that.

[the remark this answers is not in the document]
liliu-z: Created an issue https://github.com/zilliztech/knowhere/issues/779 to track this and unblock this PR

liliu-z (approved):
/lgtm

Issue #750, opened 2024-08-07
Provide a HasRawData interface to statically determine whether the index maintains raw data


foxspy:
According to Milvus's requirements, it is necessary to determine whether the index contains raw data before loading it for capacity evaluation. In this context, it is necessary to provide a static interface to meet verification requirements. This solution is only temporary; as the complexity of the index increases, a unified resource management interface will be needed to support the evaluation of index resources.

PR #751, opened 2024-08-07
[Enhancement] add statically hasRawData interface

foxspy:
issue: #750

### DIFF
--- a/include/knowhere/comp/knowhere_check.h
+++ b/include/knowhere/comp/knowhere_check.h
@@ -16,6 +16,10 @@
 
 #include "knowhere/comp/index_param.h"
 #include "knowhere/index/index_factory.h"
+#ifdef KNOWHERE_WITH_CARDINAL
+#include "cardinal/cardinal_utils.h"
+#endif
+
 namespace knowhere {
 namespace KnowhereCheck {
 static bool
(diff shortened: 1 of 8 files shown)


foxspy:
issue: #750

include/knowhere/comp/knowhere_check.h
```diff
@@ -38,6 +42,93 @@ SuppportMmapIndexTypeCheck(const std::string& index_name) {
         return false;
     }
 }
+
+inline bool
+CheckBooleanInJson(const knowhere::Json& json, std::string key, bool target) {
```
alexanderguzhva, line 47: any checks for `1` or `0` as well?
Something like `value == "1"` or `value == 1`

include/knowhere/comp/knowhere_check.h
```diff
@@ -38,6 +42,93 @@ SuppportMmapIndexTypeCheck(const std::string& index_name) {
...
+            return IsMetricType(metricType, metric::L2) || IsMetricType(metricType, metric::COSINE);
+#endif
+        } else if (indexType == IndexEnum::INDEX_FAISS_SCANN) {
+            return !CheckBooleanInJson(params, indexparam::WITH_RAW_DATA, false);
+            // INDEX_FAISS_IVFSQ_CC is not online yet
+        } else if (indexType == IndexEnum::INDEX_FAISS_IVFSQ_CC) {
+            return params.find(indexparam::RAW_DATA_STORE_PREFIX) != params.end();
+        } else if (indexType == IndexEnum::INDEX_FAISS_IDMAP) {
+            if (knowhere::Version(version) <= Version::GetMinimalVersion()) {
+                return !IsMetricType(metricType, metric::COSINE);
```
alexanderguzhva, line 119: why does not cosine have a raw data? O_o

alexanderguzhva:
lgtm overall, just a couple of questions

include/knowhere/comp/knowhere_check.h
foxspy: This is consistent with the constraint represented by the bool in the configuration. Currently, something like 1 would be considered an invalid parameter.

include/knowhere/comp/knowhere_check.h
foxspy: There are some reasons, mainly because at one point, cosine was directly normalizing the raw data, which resulted in the raw data being unavailable.

PR #755, opened 2024-08-09
iterator with order

alwayslove2013:
issue: #758

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -636,6 +636,7 @@ class BaseConfig : public Config {
     CFG_INT range_search_k;
     CFG_FLOAT range_filter;
     CFG_FLOAT range_search_level;
+    CFG_BOOL retain_iterator_order;
     CFG_BOOL trace_visit;
     CFG_BOOL enable_mmap;
     CFG_BOOL enable_mmap_pop;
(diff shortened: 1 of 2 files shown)


include/knowhere/index/index_node.h
```diff
@@ -243,6 +246,16 @@ class IndexIterator : public IndexNode::iterator {
         auto ret = q.top();
         q.pop();
         UpdateNext();
+        if (with_order_) {
+            while (HasNext()) {
+                auto next_ret = q.top();
+                if (next_ret.val >= ret.val) {
```
PwzXxm, line 252: What about other metrics? Cosine/IP?

include/knowhere/config.h
```diff
@@ -636,6 +636,7 @@ class BaseConfig : public Config {
     CFG_INT range_search_k;
     CFG_FLOAT range_filter;
     CFG_FLOAT range_search_level;
+    CFG_BOOL iterator_with_order;
```
zhengbuqian, line 639: retain_iterator_order

include/knowhere/index/index_node.h
PwzXxm: Now Explained in the comments.

alwayslove2013:
issue: #758

PR #756, opened 2024-08-09
expose knowhere config as go package

Presburger:
/kind improvement
issue: #757

### DIFF
--- a/knowhere.go
+++ b/knowhere.go
@@ -1,0 +1,83 @@
+package knowhere
+
+/*
+#cgo pkg-config: knowhere
+#include <stdlib.h>
+int CheckConfig(int index_type, char const* str, int n, int param_type);
+*/
+import "C"
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"unsafe"
+)
+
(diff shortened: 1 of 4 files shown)


Presburger:
> @Presburger may I ask why this is needed? p.s. the code looks good

@alexanderguzhva   Thanks, We want to unify and reuse the config check logic of the Milvus proxy frontend with the checking code in Knowhere. Currently, some of the checks in the Golang frontend are completely independent, which can lead to inconsistencies between the two sides.

Presburger:
issue: #757

foxspy (approved):
/lgtm

knowhere.go
```diff
@@ -0,0 +1,83 @@
...
+import "C"
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"unsafe"
+)
+
+func CheckConfig(name string, params map[string]string, ptype string) error {
```
liliu-z, line 17: Suggest to put the cgo function in segcore aligning with all other APIs.
We can put the `IndexType - number` mapping in Knowhere side.

PR #764, opened 2024-08-16
enhance: increase range search default param

PwzXxm:
issue: https://github.com/milvus-io/milvus/issues/35464


PwzXxm:
kind/improvement

PR #773, opened 2024-08-22
Show illegal vector value in diskann

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/35594

### DIFF
--- a/thirdparty/DiskANN/src/partition_and_pq.cpp
+++ b/thirdparty/DiskANN/src/partition_and_pq.cpp
@@ -206,7 +206,10 @@ int generate_pq_pivots(const float *passed_train_data, size_t num_train,
   for (uint64_t i = 0; i < num_train; i++) {
     for (uint64_t j = 0; j < dim; j++) {
       if (passed_train_data[i * dim + j] != train_data[i * dim + j])
-        LOG(ERROR) << "error in copy";
+        LOG(ERROR) << "error in copy, (" << i << ", " << j << ") "
+                   << train_data[i * dim + j] << " vs. "
+                   << passed_train_data[i * dim + j] << " (0x" << std::hex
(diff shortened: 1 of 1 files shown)


thirdparty/DiskANN/src/partition_and_pq.cpp
```diff
@@ -206,7 +206,7 @@ int generate_pq_pivots(const float *passed_train_data, size_t num_train,
   for (uint64_t i = 0; i < num_train; i++) {
     for (uint64_t j = 0; j < dim; j++) {
       if (passed_train_data[i * dim + j] != train_data[i * dim + j])
-        LOG(ERROR) << "error in copy";
+        LOG(ERROR) << "error in copy (0x" << std::hex << *(int32_t*)(&train_data[i]) << ")";
```
alexanderguzhva, line 209: Please log `i` and `j` as well, because a message `error in copy 0x10` is not that more informative compared to `error in copy`. 

Also, have you checked what gets printed in case of `Nan` or `Inf`.
Also, if this code is a weird way of checking for `Nan` or `Inf`, then why doesn't a function like `isnanf()` is used?

alexanderguzhva (changes requested):

thirdparty/DiskANN/src/partition_and_pq.cpp
cydrain: log updated, the message will like this
```
E20240823 18:49:21.503752 3378223 partition_and_pq.cpp:209] error in copy, (3062, 254) -nan vs. -nan (0xffc00000)
E20240823 18:49:21.503759 3378223 partition_and_pq.cpp:209] error in copy, (3062, 255) -nan vs. -nan (0xffc00000)
```
I explained in details in https://github.com/milvus-io/milvus/issues/35594

liliu-z (approved):
/lgtm

Issue #774, opened 2024-08-23
Sparse InvertedIndex::Train SigSegv if all vectors are empty


zhengbuqian:

zhengbuqian:
https://github.com/milvus-io/milvus/issues/35548

PR #775, opened 2024-08-23
fix InvertedIndex::Train when all input vectors are empty

zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/774
/kind bug

### DIFF
--- a/src/index/sparse/sparse_inverted_index.h
+++ b/src/index/sparse/sparse_inverted_index.h
@@ -200,6 +200,9 @@ class InvertedIndex : public BaseInvertedIndex<T> {
         for (size_t i = 0; i < rows; ++i) {
             amount += data[i].size();
         }
+        if (amount == 0) {
+            return Status::success;
+        }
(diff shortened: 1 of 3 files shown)


src/index/sparse/sparse_inverted_index.h
```diff
@@ -200,6 +200,9 @@ class InvertedIndex : public BaseInvertedIndex<T> {
         for (size_t i = 0; i < rows; ++i) {
             amount += data[i].size();
         }
+        if (amount == 0) {
+            return Status::success;
```
alexanderguzhva, line 204: this early exit does not set rhe following values (lines 222-223)
```
        value_threshold_ = *pos;
        drop_during_build_ = true;
```
is it expected?

zhengbuqian: yes, they have default values of 0 and false.

liliu-z (approved):
lgtm

PR #781, opened 2024-08-26
[iterator_range_search] remove min_exceed_cnt

alwayslove2013:
For RangeSearch (based on iterator).
The `num_consecutive_over_further_bound` will be fully controlled by the “range_search_level”.

/kind improvement

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -162,7 +162,6 @@ class IndexNode : public Object {
          * - terminate iterator if next distance [consecutively] exceeds `further_bound` several times.
          * - if get enough results (`range_search_k`), update a `tighter_further_bound`, to early terminate iterator.
          * */
-        constexpr size_t k_min_num_consecutive_over_further_bound = 16;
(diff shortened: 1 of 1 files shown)


include/knowhere/index/index_node.h
```diff
@@ -185,7 +184,7 @@ class IndexNode : public Object {
                 if (same_or_too_far(dist)) {
                     num_consecutive_over_further_bound++;
                     if (num_consecutive_over_further_bound >
-                        std::max(k_min_num_consecutive_over_further_bound, (size_t)(num_next * range_search_level))) {
+                        static_cast<size_t>(std::ceil(num_next * range_search_level))) {
```
PwzXxm, line 187: Just to double-check, the `num_next` will at least be `1` and we at least need 2 consecutive jumps over further bound to stop. Is it intentional?

alwayslove2013: Yes, that's why I use `ceil`.

liliu-z (approved):
/lgtm

PR #791, opened 2024-08-28
fix uncompilable test_sparse UT

alexanderguzhva:
`test_sparse.cc` produces a compilation error in the master branch. This PR fixes the problem.
/kind bug


foxspy (approved):
/lgtm

PR #799, opened 2024-08-29
Set norm to 1.0 for all-0 vectors

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/35594

### DIFF
--- a/src/simd/distances_avx.cc
+++ b/src/simd/distances_avx.cc
@@ -485,7 +485,7 @@ fp16_vec_norm_L2sqr_avx(const knowhere::fp16* x, size_t d) {
         msum_0 = _mm256_fmadd_ps(mx, mx, msum_0);
     }
     auto res = _mm256_reduce_add_ps(msum_0);
-    return res;
+    return (res == 0.0 ? 1.0 : res);
 }
 
 float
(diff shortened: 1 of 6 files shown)


src/simd/distances_avx.cc
```diff
@@ -485,7 +485,7 @@ fp16_vec_norm_L2sqr_avx(const knowhere::fp16* x, size_t d) {
         msum_0 = _mm256_fmadd_ps(mx, mx, msum_0);
     }
     auto res = _mm256_reduce_add_ps(msum_0);
-    return res;
+    return (res == 0.0 ? 1.0 : res);
```
alexanderguzhva, line 488: Such a modification in `distances-*.cpp` is absolutely and surely inappropriate. Checks for non-zero needs to be conducted in the code that calls `vec_norm` functions, similar to how it is done in diskann `utils.h` change:
```
    float norm = fvec_norm_L2(...);
    if (absf(norm) < SOME_EPSILON) { norm = 1.0f }
```

alexanderguzhva (changes requested):
unacceptable

src/simd/distances_avx.cc
cydrain: I pull another PR https://github.com/zilliztech/knowhere/pull/803, please check

PR #802, opened 2024-08-30
Optimize NormalizeVec

cydrain:


src/common/utils.cc
```diff
@@ -41,7 +41,9 @@ NormalizeVec(DataType* x, int32_t d) {
     } else if constexpr (std::is_same_v<DataType, bf16>) {
         norm_l2_sqr = faiss::bf16_vec_norm_L2sqr(x, d);
     } else {
-        LOG_KNOWHERE_ERROR_ << "unknown datatype";
+        for (int32_t i = 0; i < d; i++) {
+            norm_l2_sqr += (float)x[i] * (float)x[i];
+        }
```
liliu-z, line 46: why we don't throw exception here? the float cast is an undefined operation here.

thirdparty/hnswlib/hnswlib/space_l2.h
```diff
@@ -8,7 +8,13 @@ namespace hnswlib {
...
 NormSqr(const void* pVect1v, const void* qty_ptr) {
-    if constexpr (!std::is_same_v<DataType, float>) {
+    if constexpr (std::is_same_v<DataType, knowhere::fp32>) {
+        return faiss::fvec_norm_L2sqr((const DataType*)pVect1v, *(size_t*)(qty_ptr));
+    } else if constexpr (std::is_same_v<DataType, knowhere::fp16>) {
+        return faiss::fp16_vec_norm_L2sqr((const DataType*)pVect1v, *(size_t*)(qty_ptr));
+    } else if constexpr (std::is_same_v<DataType, knowhere::bf16>) {
+        return faiss::bf16_vec_norm_L2sqr((const DataType*)pVect1v, *(size_t*)(qty_ptr));
+    } else {
         auto pVect1 = (DataType*)pVect1v;
```
liliu-z, line 18: Same here

src/common/utils.cc
cydrain: updated

thirdparty/hnswlib/hnswlib/space_l2.h
cydrain: updated

liliu-z (commented):
/lgtm

liliu-z:
Looks like we get many places didn't take care of the exception throwing like https://github.com/zilliztech/knowhere/blob/613f06f15597ff0abea79801c6995af5e2fb7713/src/common/comp/brute_force.cc#L136
We need to try catch is to avoid folly thread crash.

liliu-z:
Create a issue to track the fix https://github.com/zilliztech/knowhere/issues/804

@cydrain

PR #814, opened 2024-09-05
Remove f16c flag in utils_sse

cqy123456:
issue: #815 
Milvus issue: https://github.com/milvus-io/milvus/issues/35877

### DIFF
--- a/src/simd/distances_sse.cc
+++ b/src/simd/distances_sse.cc
@@ -66,24 +66,6 @@ fvec_norm_L2sqr_sse(const float* x, size_t d) {
 }
 
 float
-fp16_vec_norm_L2sqr_sse(const knowhere::fp16* x, size_t d) {
-    __m128 m_res = _mm_setzero_ps();
-    while (d >= 4) {
-        __m128 m_x = _mm_cvtph_ps(_mm_loadl_epi64((const __m128i_u*)x));
-        m_res = _mm_add_ps(m_res, _mm_mul_ps(m_x, m_x));
-        x += 4;
-        d -= 4;
-    }
-    if (d > 0) {
-        __m128 m_x = _mm_cvtph_ps(mm_masked_read_short(d, (uint16_t*)x));
-        m_res = _mm_add_ps(m_res, _mm_mul_ps(m_x, m_x));
-    }
-    m_res = _mm_hadd_ps(m_res, m_res);
-    m_res = _mm_hadd_ps(m_res, m_res);
-    return _mm_cvtss_f32(m_res);
-}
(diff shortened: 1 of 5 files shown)


src/simd/hook.cc
```diff
@@ -89,12 +89,6 @@ cpu_support_sse4_2() {
     InstructionSet& instruction_set_inst = InstructionSet::GetInstance();
     return (instruction_set_inst.SSE42());
 }
-
```
alexanderguzhva, line 92: I would keep this function just in case

src/simd/distances_sse.cc
```diff
@@ -67,20 +68,12 @@ fvec_norm_L2sqr_sse(const float* x, size_t d) {
 
 float
 fp16_vec_norm_L2sqr_sse(const knowhere::fp16* x, size_t d) {
-    __m128 m_res = _mm_setzero_ps();
-    while (d >= 4) {
-        __m128 m_x = _mm_cvtph_ps(_mm_loadl_epi64((const __m128i_u*)x));
-        m_res = _mm_add_ps(m_res, _mm_mul_ps(m_x, m_x));
-        x += 4;
-        d -= 4;
+    std::vector<knowhere::fp32> fp32_x(d);
```
alexanderguzhva, line 71: `std::vector` introduces an enormous overhead.
Please implement these functions manually in the following way:
```
    float res = 0;
        const float xv = float(x[i]);
        res += xv * xv;
    }

    return res;
``` 
Similar for dot products

src/simd/hook.h
```diff
@@ -103,8 +103,6 @@ bool
 cpu_support_avx2();
 bool
 cpu_support_sse4_2();
-bool
```
alexanderguzhva, line 106: I'd keep this function

alexanderguzhva (changes requested):

alexanderguzhva:
@cqy123456 We can use `_ref` versions, sure. But the downside is that a compiler will not use `sse` instructions for compiling `ref` versions, such a change introduces a performance regression. Therefore, the most optimal way is to literally copy-paste `ref` versions code into `distances_sse.cc` under new function names, because it allows a compiler to use `sse` instructions. 
It is not a big problem if we're going to have small pieces of the same code in different files.

cqy123456:
At first I tried to only rely on compiler optimization, but the compiler did not speed up fp16's (float)x * (float)y very well, probably because the fp16->float operation is more complicated.

PwzXxm:
/lgtm
Patch to release for correctness first.

PR #825, opened 2024-09-09
Set 4c16g for e2e-cpu and 8c30g for e2e-gpu

cydrain:


alexanderguzhva:
@cydrain Do I get it correct that "Set 4c16g for e2e-cpu and 8c30g for e2e-gpu" means that you set 4c16g for both e2e-cpu and e2e-gpu, because this is what this PR does?

cydrain:
> @cydrain Do I get it correct that "Set 4c16g for e2e-cpu and 8c30g for e2e-gpu" means that you set 4c16g for both e2e-cpu and e2e-gpu, because this is what this PR does?

@alexanderguzhva, 
Setting limits and requests to be the same is to ensure that these resources are exclusive during the e2e process.
For e2e-cpu and e2e-arm, less resources of 4c16g are required, while for e2e-gpu, more resources of 8c30g are needed

PR #827, opened 2024-09-10
modify the API to use smart pointers to pave the way for the async

Presburger:
The async interface needs to hold the config object even after the API call has ended.

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -60,11 +60,15 @@ class IndexNode : public Object {
      *
      * @note Indexes need to be ready to search after `Build` is called. TODO:@liliu-z DiskANN is an exception and need
      * to be revisited.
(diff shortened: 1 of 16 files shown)


Presburger:
issue: #829

include/knowhere/index/index_node.h
```diff
@@ -62,9 +62,9 @@ class IndexNode : public Object {
      * to be revisited.
      */
     virtual Status
-    Build(const DataSetPtr dataset, const Config& cfg) {
+    Build(const DataSetPtr dataset, std::shared_ptr<Config> cfg) {
```
alexanderguzhva, line 65: Please add comments to all methods on why it is needed to pass `Config` this way. It is clear from the description of this PR, but it not very clear without it. 
Please also explain why you pick `std::shared_ptr<>` or `std::unique_ptr<>` depending on a use case.

alexanderguzhva:
/lgtm 
overall

PR #841, opened 2024-09-14
fix: fail to load ivf_flat(metric = cosine) with DeserializeFromFile

cqy123456:
issue:https://github.com/milvus-io/milvus/issues/36052

When we set IO_FLAG_WITH_NORM = true to read a faiss index,  ArrayInvertedLists will read into a read-only OnDiskInvertedLists. 
OnDiskInvertedLists will mmap all ids and vectors with the index file,  and record the ids and vectors offset in memory. If we use cosine, l1-norms will save in index file and all offsets are wrong.

### DIFF
--- a/thirdparty/faiss/faiss/invlists/OnDiskInvertedLists.cpp
+++ b/thirdparty/faiss/faiss/invlists/OnDiskInvertedLists.cpp
@@ -674,7 +674,25 @@ void OnDiskInvertedLists::set_all_lists_sizes(const size_t* sizes) {
     for (size_t i = 0; i < nlist; i++) {
         lists[i].offset = ofs;
         lists[i].capacity = lists[i].size = sizes[i];
-        ofs += sizes[i] * (sizeof(idx_t) + code_size);
+        if (this->with_norm) {
+            ofs += sizes[i] * (sizeof(idx_t) + code_size + sizeof(float));
+        } else {
+            ofs += sizes[i] * (sizeof(idx_t) + code_size);
+        }
+    }
+}
+
+const float* OnDiskInvertedLists::get_code_norms(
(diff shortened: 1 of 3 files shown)


foxspy (approved):
/lgtm

thirdparty/faiss/faiss/invlists/OnDiskInvertedLists.cpp
```diff
@@ -674,10 +674,33 @@ void OnDiskInvertedLists::set_all_lists_sizes(const size_t* sizes) {
...
+            return nullptr;
+        }
+        return (const float*)(ptr + lists[list_no].offset +
+                            (code_size + sizeof(idx_t)) * lists[list_no].capacity);
+    } else {
+        return nullptr;
     }
 }
 
+void OnDiskInvertedLists::release_code_norms(size_t list_no, const float* codes)
```
alexanderguzhva, line 699: why is it needed to override this method in the provided way?

alexanderguzhva: Just to clarify the question. It seems that the base class has the same implementation as the added one

cqy123456: I thought the behavior in release_code_norms is based on the storage method of InvertedList, so this function is overrided, and this function do nothing like the basic class.

alexanderguzhva: Then I would eliminate it, because it does nothing.
Or I can lgtm the PR, just let me know. :)

liliu-z (approved):
/lgtm

Issue #809, opened 2024-09-03
Refactor knowhere UT


cydrain:
Target:
1. code coverage > 80%
2. runtime for each case < 5s

https://zilliverse.feishu.cn/wiki/S8JtwBcIyiPWYkkdhqdcJalmnM6
Task items:
- [x]  update test_config
- [x]  update test_knowhere_init
- [x]  update test_index_check
- [x]  update test_utils for thread_pool
- [ ]  update test_feder for diskann
- [x]  add ut for index.h / index_node.h / index_node_xxx.h
- [x]  add ut for src/common
- [x]  re-write test_simd
- [ ]  add test for cluster ( @chasingegg )
- [ ]  add test for bruteforce sparse ( @zhengbuqian )
- [ ]  add test for index range search ( @zhengbuqian )
- [ ]  add test for index iterator ( @zhengbuqian )
- [ ]  use trompeloeil to mock index ( @cydrain )

liliu-z:
Let's first introduce some timeout mechanism

PR #850, opened 2024-09-18
Add more unittest for src/common

cydrain:
issue: #809

### DIFF
--- a/tests/ut/test_tracer.cc
+++ b/tests/ut/test_tracer.cc
@@ -9,6 +9,8 @@
 // is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 // or implied. See the License for the specific language governing permissions and limitations under the License
 
+#include <unistd.h>
+
 #include <memory>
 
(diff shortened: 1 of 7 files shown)


tests/ut/test_tracer.cc
```diff
@@ -22,20 +24,47 @@ using namespace opentelemetry::trace;
...
+        AddEvent("sleep");
+        usleep(2000);
+        span->End();
+        CloseRootSpan();
+    }
+
+    SECTION("check jaeger") {
+        auto config = std::make_shared<TraceConfig>();
+        config->exporter = "jaeger";
+        config->jaegerURL = "http://localhost:14268/api/traces";
```
alexanderguzhva, line 44: would it be possible to add some comments about the URL? Because it is unclear whether a unit test is going to launch a local socket/server that listens on this port. Or, maybe, an external optional tracing server might be used if available.

cydrain: add some comment

cydrain:
Overall coverage rate:
  lines......: 75.1% (4679 of 6230 lines)
  functions..: 58.9% (1004 of 1706 functions)

PR #851, opened 2024-09-18
fix BM25 performance issue

zhengbuqian:
/kind improvement


src/index/sparse/sparse_inverted_index.h
```diff
@@ -613,7 +619,7 @@ class InvertedIndex : public BaseInvertedIndex<T> {
         float b;
         // row_sums is used to cache the sum of values of each row, which
         // corresponds to the document length of each doc in the BM25 formula.
-        std::unordered_map<size_t, T> row_sums;
+        std::vector<T> row_sums;
```
alexanderguzhva, line 622: Prior to the change, `row_sums` utilized `::at()` method that throws `std::out_of_range` exception in case of problems. Thus, it is needed to keep accessing elements of `std::vector`-based `rows_sums` using `::at()` method if we want to keep consistency

alexanderguzhva (changes requested):

src/index/sparse/sparse_inverted_index.h
zhengbuqian: good call!

Issue #861, opened 2024-09-23
IVF_FLAT build time degrade


cydrain:
![Screenshot from 2024-09-23 10-42-44](https://github.com/user-attachments/assets/32b3bd15-8e23-4472-94a9-d90bebae43ca)

cydrain:
| Commit time | Commit ID | IVF_FLAT build time |
-------------------------------------------------------------------
| Sep 8 22:24:59 | 67c4a64 |  9.5s |
| Sep 14 11:45:03 | 352ccc7 | 46.6s |

cydrain:
back to normal
![Screenshot from 2024-09-26 16-20-18](https://github.com/user-attachments/assets/e535b0be-83d4-4b1a-946e-ff5e239f9cd4)

PR #863, opened 2024-09-23
Set scoped omp to fix IVF index build degrade

cydrain:
issue: #861

### DIFF
--- a/benchmark/hdf5/benchmark_knowhere.h
+++ b/benchmark/hdf5/benchmark_knowhere.h
@@ -27,6 +27,9 @@
 #include "knowhere/index/index_factory.h"
 #include "knowhere/version.h"
 
+static const size_t default_build_thread_num = 8;
+static const size_t default_search_thread_num = 8;
+
 namespace fs = std::filesystem;
 std::string kDir = fs::current_path().string() + "/diskann_test";
 std::string kRawDataPath = kDir + "/raw_data";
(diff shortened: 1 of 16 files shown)


benchmark/hdf5/benchmark_knowhere.h
```diff
@@ -27,6 +27,9 @@
 #include "knowhere/index/index_factory.h"
 #include "knowhere/version.h"
 
+static const size_t default_build_thread_num = 8;
```
alexanderguzhva, line 30: https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency ? Please don't forget to check for 0 if you decide to go this path.

cydrain: Hi Alex, this variable is only used in benchmark

Presburger (approved):

src/common/thread/thread.cc
```diff
@@ -72,9 +72,14 @@ GetBuildThreadPoolSize() {
     return ThreadPool::GetGlobalBuildThreadPool()->size();
 }
 
-std::unique_ptr<ThreadPool::ScopedOmpSetter>
-CreateScopeOmpSetter(int num_threads) {
-    return std::make_unique<ThreadPool::ScopedOmpSetter>(num_threads);
+std::unique_ptr<ThreadPool::ScopedBuildOmpSetter>
```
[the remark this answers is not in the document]
cydrain: this API will be removed later

PR #873, opened 2024-09-27
async build support for cardinal

Presburger:
support early stop and timeout

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -34,6 +34,8 @@
 
 namespace knowhere {
 
+class Interrupt;
+
 class IndexNode : public Object {
  public:
     IndexNode(const int32_t ver) : version_(ver) {
@@ -70,6 +72,16 @@ class IndexNode : public Object {
         RETURN_IF_ERROR(Train(dataset, cfg));
         return Add(dataset, std::move(cfg));
     }
+
+/*
+ *@ @brief Builds the index using the provided dataset,configuration and handle.
+ */
+#ifdef KNOWHERE_WITH_CARDINAL
+    virtual Status
+    BuildAsync(const DataSetPtr dataset, std::shared_ptr<Config> cfg, const Interrupt* = nullptr) {
+        return Build(dataset, std::move(cfg));
+    }
+#endif
 
     /**
      * @brief Trains the index model using the provided dataset and configuration.
--- a/src/index/index.cc
+++ b/src/index/index.cc
@@ -12,6 +12,8 @@
 #include "knowhere/index/index.h"
 
 #include "fmt/format.h"
+#include "folly/futures/Future.h"
+#include "knowhere/comp/thread_pool.h"
 #include "knowhere/comp/time_recorder.h"
 #include "knowhere/dataset.h"
 #include "knowhere/expected.h"
@@ -33,6 +35,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
     RETURN_IF_ERROR(res);
     return Config::Load(*cfg, json_, param_type, msg);
 }
+
+#ifdef KNOWHERE_WITH_CARDINAL
+template <typename T>
+inline const std::shared_ptr<Interrupt>
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json, const std::chrono::seconds timeout) {
+    auto pool = ThreadPool::GetGlobalBuildThreadPool();
+    auto interrupt = std::make_shared<Interrupt>(timeout);
+    interrupt->Set(pool->push([this, dataset, &json, &interrupt]() {
+        auto cfg = this->node->CreateConfig();
+        RETURN_IF_ERROR(LoadConfig(cfg.get(), json, knowhere::TRAIN, "Build"));
+
+#if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOWHERE_WITH_LIGHT)
+        TimeRecorder rc("BuildAsync index ", 2);
+        auto res = this->node->BuildAsync(dataset, std::move(cfg), interrupt.get());
+        auto time = rc.ElapseFromBegin("done");
+        time *= 0.000001;  // convert to s
+        knowhere_build_latency.Observe(time);
+#else
+        auto res = this->node->BuildAsync(dataset, std::move(cfg), Interrupt.get());
+#endif
+        return res;
+    }));
+    return interrupt;
+}
+#else
+template <typename T>
+inline const std::shared_ptr<Interrupt>
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json) {
+    auto pool = ThreadPool::GetGlobalBuildThreadPool();
+    auto interrupt = std::make_shared<Interrupt>();
+    interrupt->Set(pool->push([this, &dataset, &json]() { return this->Build(dataset, json); }));
+    return interrupt;
+}
+#endif
 
 template <typename T>
 inline Status
--- a/src/index/interrupt.cc
+++ b/src/index/interrupt.cc
@@ -53,4 +53,7 @@ Interrupt::Set(folly::Future<Status>&& future) {
     this->future = std::make_unique<folly::Future<Status>>(std::move(future));
 }
 
+Interrupt::~Interrupt() {
+}
+
 }  // namespace knowhere
--- a/src/index/interrupt.cc
+++ b/src/index/interrupt.cc
@@ -11,6 +11,7 @@
 
 #include "knowhere/index/interrupt.h"
 
+#include "folly/futures/Future.h"
 namespace knowhere {
 #ifdef KNOWHERE_WITH_CARDINAL
 Interrupt::Interrupt(const std::chrono::seconds& timeout) : start(std::chrono::steady_clock::now()), timeout(timeout) {
--- a/src/index/interrupt.cc
+++ b/src/index/interrupt.cc
@@ -1,0 +1,55 @@
+// Copyright (C) 2019-2023 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#include "knowhere/index/interrupt.h"
+
+namespace knowhere {
+#ifdef KNOWHERE_WITH_CARDINAL
+Interrupt::Interrupt(const std::chrono::seconds& timeout) : start(std::chrono::steady_clock::now()), timeout(timeout) {
+}
+#else
+Interrupt::Interrupt() = default;
+#endif
+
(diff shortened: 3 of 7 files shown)


include/knowhere/index/handle.h
```diff
@@ -0,0 +1,60 @@
...
+    void
+    Set(folly::Future<Status>&& future);
+
+    ~Handle();
+
+ private:
+#ifdef KNOWHERE_WITH_CARDINAL
+    std::chrono::steady_clock::time_point start;
+    std::chrono::seconds timeout;
+    mutable std::atomic_bool flag = false;
```
alexanderguzhva, line 55: Could you please add the comments to Handle class about the way it is designed to work?

include/knowhere/index/handle.h
```diff
@@ -0,0 +1,60 @@
...
+namespace folly {
+template <typename T>
+class Future;
+
+}
+
+namespace knowhere {
+class Handle {
+ public:
+#ifdef KNOWHERE_WITH_CARDINAL
```
alexanderguzhva, line 28: why is it needed to add pragmas for `KNOWHERE_WITH_CARDINAL` for this particular case at all? I think it can be avoided by just ignoring timeouts in non-cardinal based implementations

include/knowhere/index/handle.h
```diff
@@ -0,0 +1,60 @@
...
+#include "knowhere/expected.h"
+
+namespace folly {
+template <typename T>
+class Future;
+
+}
+
+namespace knowhere {
+class Handle {
```
alexanderguzhva, line 26: would it be possible to give a more specific name? `Handle` is a very generic one. Maybe, make the class name contain a word `Interrupt` or `Timeout`?

include/knowhere/index/handle.h
```diff
@@ -0,0 +1,60 @@
...
+
+    ~Handle();
+
+ private:
+#ifdef KNOWHERE_WITH_CARDINAL
+    std::chrono::steady_clock::time_point start;
+    std::chrono::seconds timeout;
+    mutable std::atomic_bool flag = false;
+#endif
+    std::unique_ptr<folly::Future<Status>> future = nullptr;
```
alexanderguzhva, line 57: same here. The name `future` is way too generic. Is it a handler, a callback, just a future? Also, please add some comments on what this is.

src/index/handle.cc
```diff
@@ -0,0 +1,58 @@
...
+        return Status::timeout;
+    return std::move(*future).get();
+}
+
+void
+Handle::Set(folly::Future<Status>&& future) {
+    this->future = std::make_unique<folly::Future<Status>>(std::move(future));
+}
+
+Handle::~Handle() {
```
alexanderguzhva, line 55: The destructor does nothing. Is it needed at all?

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
     return Config::Load(*cfg, json_, param_type, msg);
 }
 
+#ifdef KNOWHERE_WITH_CARDINAL
+template <typename T>
+inline const std::unique_ptr<Handle>
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json, const std::chrono::seconds timeout) {
+    auto pool = ThreadPool::GetGlobalBuildThreadPool();
+    auto handle = std::make_unique<Handle>(timeout);
```
alexanderguzhva, line 44: Please switch to `std::shared_ptr<Handle>` here. The following `handle->Set()` line does not convince me that we won't have a future lifetime problems, and I'm not sure whether Rust compiler would easily allow this form of self-reference, for example. Making `pool->push()` capture `std::shared_ptr` (not `std::shared_ptr&`) seems more safe.

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
+    handle->Set(pool->push([this, dataset, &json, &handle]() {
+        auto cfg = this->node->CreateConfig();
+        RETURN_IF_ERROR(LoadConfig(cfg.get(), json, knowhere::TRAIN, "Build"));
+
+#if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOWHERE_WITH_LIGHT)
+        TimeRecorder rc("BuildAsync index ", 2);
+        auto res = this->node->BuildAsync(dataset, std::move(cfg), handle.get());
```
alexanderguzhva, line 51: must be 
```C++
std::shared_ptr<Handle> handle;
...
auto res = this->node->BuildAsync(dataset, std::move(cfg), handle);
```

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
+        auto time = rc.ElapseFromBegin("done");
+        time *= 0.000001;  // convert to s
+        knowhere_build_latency.Observe(time);
+#else
+        auto res = this->node->BuildAsync(dataset, std::move(cfg), handle.get());
```
alexanderguzhva, line 56: same

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
+#endif
+        return res;
+    }));
+    return handle;
+}
+#else
+template <typename T>
+inline const std::unique_ptr<Handle>
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json) {
```
alexanderguzhva, line 65: same. Please switch to `std::shared_ptr<Handle>`

alexanderguzhva:
/hold
let's figure out the situation with lifetimes and make sure that we won't have dangling pointers and references in the future.

src/index/index.cc
Presburger: Each `async build` call generates only one handle object, and all references are just pointers. Moreover, the called functions are all `const` functions.

include/knowhere/index/index_node.h
```diff
@@ -71,6 +73,16 @@ class IndexNode : public Object {
         return Add(dataset, std::move(cfg));
     }
 
+/*
+ *@ @brief Builds the index using the provided dataset,configuration and handle.
+ */
+#ifdef KNOWHERE_WITH_CARDINAL
+    virtual Status
+    BuildAsync(const DataSetPtr dataset, std::shared_ptr<Config> cfg, const Handle* = nullptr) {
```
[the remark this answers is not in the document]
Presburger: I want to maintain compatibility with the existing code.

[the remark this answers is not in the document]
Presburger: The open-source version only supports asynchronous operations but doesn't allow the build process to be stopped proactively, whereas Cardinal supports this feature. To enable this in the open-source version, it would require a significant amount of code changes.

src/index/index.cc
alexanderguzhva: the reason why is that 
```
    interrupt->Set(pool->push([this, dataset, &json, &interrupt]() {
```

Basically, an object captures a reference to smart pointer. This needs to be implemented as capturing the smart pointer itself, not a reference to it. Thus, `shared_ptr` is needed.

I understand the suggestion, but the reality is that C++ async brings to the territory of extremely nasty bugs. Thus, we need to make very pessimistic assumptions on code usage.

src/index/interrupt.cc
```diff
@@ -0,0 +1,57 @@
...
+#ifdef KNOWHERE_WITH_CARDINAL
+Interrupt::Interrupt(const std::chrono::seconds& timeout) : start(std::chrono::steady_clock::now()), timeout(timeout) {
+}
+#else
+Interrupt::Interrupt() = default;
+#endif
+
+#ifdef KNOWHERE_WITH_CARDINAL
+void
+Interrupt::Stop() {
```
liliu-z, line 25: Why we switch the name from `Handle` to `Interrupt`? `Handler` sounds more precise to me, since it is responsible for interrupting, status check, waiting, etc.

include/knowhere/index/index_node.h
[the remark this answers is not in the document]
liliu-z: > The open-source version only supports asynchronous operations but doesn't allow the build process to be stopped proactively, whereas Cardinal supports this feature. To enable this in the open-source version, it would require a significant amount of code changes.

Do you mean the change in different algos to interrupt? How about keeping the API sync and simply throw exception when the `stop` get called?

include/knowhere/index/handle.h
Presburger: The asynchronous interface is universally supported, but only Cardinal supports the timeout and proactive stop mechanisms.

include/knowhere/index/handle.h
Presburger: The changes have been made.

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
+#else
+        auto res = this->node->BuildAsync(dataset, std::move(cfg), Interrupt.get());
+#endif
+        return res;
+    }));
+    return interrupt;
+}
+#else
+template <typename T>
+inline const std::unique_ptr<Handle>
```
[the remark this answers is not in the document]
Presburger: fixed,hh

include/knowhere/index/index_node.h
```diff
@@ -71,6 +73,16 @@ class IndexNode : public Object {
...
     }
 
+/*
+ *@ @brief Builds the index using the provided dataset,configuration and handle.
+ */
+#ifdef KNOWHERE_WITH_CARDINAL
+    virtual Status
+    BuildAsync(const DataSetPtr dataset, std::shared_ptr<Config> cfg, const Interrupt* = nullptr) {
+        return Build(dataset, std::move(cfg));
+    }
```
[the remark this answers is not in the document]
Presburger: It will be executed in other threads, so it won't block the calling code.

src/index/index.cc
Presburger: Alex, you are correct. There's a potential risk here. If the code that calls `buildAsync` directly discards the return value, the thread actually executing the build might encounter illegal access issues.

src/index/interrupt.cc
Presburger: @liliu-z Alex suggested in the corresponding PR for Cardinal that the name should be more specific, as "handle" is too generic.

include/knowhere/index/index_node.h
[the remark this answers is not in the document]
Presburger: This is the implementation of the base class, which serves as a fallback for the subclasses of `IndexNode`. Adding an `interrupt` parameter to the `block build` interface would make it even more awkward.

foxspy (approved):
/lgtm

src/index/index.cc
```diff
@@ -34,6 +36,40 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
     return Config::Load(*cfg, json_, param_type, msg);
 }
 
+#ifdef KNOWHERE_WITH_CARDINAL
+template <typename T>
+inline const std::shared_ptr<Interrupt>
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json, const std::chrono::seconds timeout) {
+    auto pool = ThreadPool::GetGlobalBuildThreadPool();
+    auto interrupt = std::make_shared<Interrupt>(timeout);
+    interrupt->Set(pool->push([this, dataset, &json, &interrupt]() {
```
liliu-z, line 45: Maybe copy instead of reference for `json` and `interrupt`

PR #875, opened 2024-09-27
faiss hnsw: get rid of a debugging config option

alexanderguzhva:
Just a small cleanup

/kind improvement


liliu-z (approved):
Talked offline. This is for original HNSW impl removing for now
/lgtm

PR #878, opened 2024-10-08
fix: Remove thread_priority in mac ThreadFactory constructor

congqixia:
In non-linux block, thread pool constructor uses folly `NamedThreadFactory` directly, which does not contain second parameter in its constructors.

This PR removes `thread_priority` from non-linux block and add comments to note this difference.

### DIFF
--- a/include/knowhere/comp/thread_pool.h
+++ b/include/knowhere/comp/thread_pool.h
@@ -97,6 +97,7 @@ class ThreadPool {
     }
 #else
  public:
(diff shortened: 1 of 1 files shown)


zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/879

liliu-z (approved):
/lgtm

zhengbuqian:
I just restarted the e2e-sse check, seems it failed due to ci node/container failure.

PR #881, opened 2024-10-09
Set resources limit for e2e-sse pod

cydrain:


ci/pod/e2e-sse.yaml
```diff
@@ -28,13 +28,13 @@ spec:
...
-    # resources:
-    #   requests:
-    #     memory: "16Gi"
-    #     cpu: "4"
-    #   limits:
-    #     memory: "16Gi"
-    #     cpu: "4"
+    resources:
+      requests:
+        memory: "3Gi"
```
liliu-z, line 33: Per our previous talk, the host is 2C4G. Why we made it 16G before?

cydrain: the host is 8c4g, Huangliang set no resource limit before

PR #886, opened 2024-10-14
Do not check e2e-sse for ci

cydrain:


zhengbuqian:
/lgtm

chatted with cydrain, after this PR:

* if all other checks passed, ci will pass and pr will be auto-merged even if e2e-sse is still running
* after the pr is merged, e2e-sse job will likely continue running
* e2e-sse will later be moved to nightly, so it'll not run for each individual pr, but run periodically on itself.

Issue #795, opened 2024-08-28
decouple knowhere with milvus


foxspy:
related: https://github.com/milvus-io/milvus/issues/34298

PR #887, opened 2024-10-14
enhance: add a bit-based structure to mark index-specific features

foxspy:
issue: #795 
/kind improvement

### DIFF
--- a/include/knowhere/feature.h
+++ b/include/knowhere/feature.h
@@ -1,0 +1,54 @@
+// Copyright (C) 2019-2023 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+#ifndef FEATURE_H
+#define FEATURE_H
+
+// these features have been report to outside (milvus); pls sync the feature code when it needs to be changed.
+namespace knowhere::feature {
+// vector datatype support : binary
+constexpr uint64_t BINARY = 1UL << 0;
+// vector datatype support : float32
+constexpr uint64_t FLOAT32 = 1UL << 1;
+// vector datatype support : fp16
+constexpr uint64_t FP16 = 1UL << 2;
+// vector datatype support : bf16
+constexpr uint64_t BF16 = 1UL << 3;
+// vector datatype support : sparse_float32
+constexpr uint64_t SPARSE_FLOAT32 = 1UL << 4;
+
+// This flag indicates that there is no need to create any index structure (build stage can be skipped)
+constexpr uint64_t NO_TRAIN = 1UL << 16;
+// This flag indicates that the index defaults to KNN search, meaning the recall is 100% (no precision loss compared
+// with original data)
+constexpr uint64_t KNN = 1UL << 17;
+// This flag indicates that the index search stage will be performed on GPU (need GPU devices)
+constexpr uint64_t GPU = 1UL << 18;
+// This flag indicates that the index support using mmap manage its mainly memory, which can significant improve the
+// capacity
+constexpr uint64_t MMAP = 1UL << 19;
+// This flag indicates that the index support using materialized view to accelerate filtering search
+constexpr uint64_t MV = 1UL << 20;
+// This flag indicates that the index need disk during search stage
+constexpr uint64_t DISK = 1UL << 21;
+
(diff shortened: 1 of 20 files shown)


include/knowhere/feature.h
```diff
@@ -0,0 +1,54 @@
...
+// vector datatype support : bf16
+constexpr uint64_t BF16 = 1UL << 3;
+// vector datatype support : sparse_float32
+constexpr uint64_t SPARSE_FLOAT32 = 1UL << 4;
+
+// This flag indicates that there is no need to create any index structure (build stage can be skipped)
+constexpr uint64_t NO_TRAIN = 1UL << 16;
+// This flag indicates that the index defaults to KNN search, meaning the recall is 100% (no precision loss compared
+// with original data)
+constexpr uint64_t KNN = 1UL << 17;
```
alexanderguzhva, line 35: I'd name this flag a bit differently.
Technically, it is not `KNN search`, it is `brute-force search`.

include/knowhere/feature.h
```diff
@@ -0,0 +1,54 @@
...
+// License for the specific language governing permissions and limitations under
+// the License.
+
+#ifndef FEATURE_H
+#define FEATURE_H
+
+// these features have been report to outside (milvus); pls sync the feature code when it needs to be changed.
+namespace knowhere::feature {
+// vector datatype support : binary
+constexpr uint64_t BINARY = 1UL << 0;
```
alexanderguzhva, line 21: can this be implemented as 
```c++
enum class FeatureFlag : uint64_t { 
    BINARY = ..., 
    FLOAT32 = ...,
    ... 
};
```
? Or is the point having a simple int value?

include/knowhere/index/index_factory.h
```diff
@@ -29,10 +29,15 @@ class IndexFactory {
     Create(const std::string& name, const int32_t& version, const Object& object = nullptr);
     template <typename DataType>
     const IndexFactory&
-    Register(const std::string& name, std::function<Index<IndexNode>(const int32_t&, const Object&)> func);
+    Register(const std::string& name, std::function<Index<IndexNode>(const int32_t&, const Object&)> func,
+             const uint64_t features);
     static IndexFactory&
     Instance();
     typedef std::tuple<std::set<std::pair<std::string, VecType>>, std::set<std::string>> GlobalIndexTable;
```
alexanderguzhva, line 36: `typedef` -> `using`
Also, is it worth converting `GlobalIndexTable` from `tuple<>` into a well-defined structure which explains all the details? It is possible that it might be extended in the future.

include/knowhere/index/index_factory.h
```diff
@@ -48,47 +53,98 @@ class IndexFactory {
         std::function<T1(const int32_t&, const Object&)> fun_value;
     };
     typedef std::map<std::string, std::unique_ptr<FunMapValueBase>> FuncMap;
+    typedef std::map<std::string, uint64_t> FeatureMap;
```
alexanderguzhva, line 56: `using` :)

include/knowhere/index/index_factory.h
```diff
@@ -48,47 +53,98 @@ class IndexFactory {
...
         [](const int32_t& version, const Object& object) {                                                 \
             return (Index<IndexNodeDataMockWrapper<data_type>>::Create(                                    \
                 std::make_unique<index_node<MockData<data_type>::type, ##__VA_ARGS__>>(version, object))); \
         },                                                                                                 \
-        data_type)
-#define KNOWHERE_REGISTER_GLOBAL_WITH_THREAD_POOL(name, index_node, data_type, thread_size)              \
+        data_type, typeCheck<data_type>(features), features)
+
+#define KNOWHERE_SIMPLE_REGISTER_ALL_GLOBAL(name, index_node, features, ...)                                          \
+    KNOWHERE_SIMPLE_REGISTER_GLOBAL(name, index_node, bin1, (features | knowhere::feature::ALL_TYPE), ##__VA_ARGS__); \
```
alexanderguzhva, line 98: I'd use `feature::BINARY` here.
Because the way it is done might contain a bug: `KNOWHERE_SIMPLE_REGISTER_ALL_GLOBAL` registers 4 types (bin, bf16, fp16, fp32) while `ALL_TYPE` consists of 5 types (including sparse fp32)

Same question applies to further macro definitions

include/knowhere/feature.h
foxspy: My view is that KNN search provides an index with 100% recall, while brute search is just a retrieval method without an index structure. What I want to express here is the former: users can choose a lossless index. This flag can be used to implement some control logic in Milvus.

include/knowhere/feature.h
foxspy: The reason for choosing uint64 is that this field will be registered in Milvus's Go memory at startup to handle some control logic (pr : https://github.com/milvus-io/milvus/pull/36843). Therefore, a general data type was chosen instead of a complex structure.

include/knowhere/index/index_factory.h
foxspy: There is an issue here: sparse is not currently defined as a data type but is borrowing float. This means it can only be handled this way for now, or it will result in registration failure.

PR #889, opened 2024-10-14
fix async bug

Presburger:
@alexanderguzhva help review


alexanderguzhva:
/lgtm
@Presburger Personally, I would use `std::shared_ptr<Json>` to avoid copying, but passing just `json` is fine as well.

Presburger:
> /lgtm @Presburger Personally, I would use `std::shared_ptr<Json>` to avoid copying, but passing just `json` is fine as well.

Alex, your suggestion is correct, but changing it to `shared_ptr` would require quite a few API modifications. 😂

Presburger:
issue: #890

PR #892, opened 2024-10-15
fix: prevent crash when running on a non-GPU machine

ashkrisk:
issue: #893
Related: https://github.com/milvus-io/milvus/issues/36894
/kind bug

### DIFF
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -186,7 +186,7 @@ KnowhereConfig::FreeGPUResource() {
 }
 
 void
-KnowhereConfig::SetRaftMemPool(size_t init_size, size_t max_size) {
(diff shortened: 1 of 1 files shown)


ashkrisk:
issue: #893
/kind bug

src/common/comp/knowhere_config.cc
```diff
@@ -186,7 +186,7 @@ KnowhereConfig::FreeGPUResource() {
 }
 
 void
-KnowhereConfig::SetRaftMemPool(size_t init_size, size_t max_size) {
+SetRaftMemPoolImpl(std::optional<size_t> init_size, std::optional<size_t> max_size) {
```
Presburger, line 189: Do not modify the function definition here, as there is code in Milvus that calls this interface.

PR #896, opened 2024-10-16
fix: update diskann config for config check

foxspy:
issue: #795 
Removed pq_code_budget_gb/search_cache_budget_gb from DiskANN and replaced them with pq_code_budget_gb_ratio/search_cache_budget_gb_ratio. This will reduce the parameter-filling logic by using a fixed ratio instead of dynamically filling GB values

### DIFF
--- a/src/index/diskann/diskann_config.h
+++ b/src/index/diskann/diskann_config.h
@@ -33,6 +33,11 @@ class DiskANNConfig : public BaseConfig {
     // complexity. Plz set this value larger than the max_degree unless you need to build indices really quickly and can
     // somewhat compromise on quality.
     CFG_INT search_list_size;
+
+    // The ratio of the size reserved for the pq code to the size of the raw data (defined with vec_field_size_gb)
+    // This parameter will replace pq_code_budget_gb to avoid calculating the actual size on the Milvus side.
+    // The index can indirectly obtain pq_code_budget_gb by vec_field_size_gb * pq_code_budget_gb_ratio
+    CFG_FLOAT pq_code_budget_gb_ratio;
     // Limit the size of the PQ code after the raw vector has been PQ-encoded. PQ code is a (pq_code_budget_gb * 1024 *
     // 1024 * 1024) / row_num)-dimensional uint8 vector. If pq_code_budget_gb is too large, it will be adjusted to the
     // size of dim*row_num.
@@ -50,6 +55,12 @@ class DiskANNConfig : public BaseConfig {
     // This is the flag to enable fast build, in which we will not build vamana graph by full 2 round. This can
     // accelerate index build ~30% with an ~1% recall regression.
     CFG_BOOL accelerate_build;
+
+    // The ratio of the size reserved for the search cache to the size of the raw data (defined with vec_field_size_gb)
+    // This parameter will replace pq_code_budget_gb to avoid calculating the actual size on the Milvus side.
+    // The index can indirectly obtain search_cache_budget_gb by vec_field_size_gb * search_cache_budget_gb_ratio
+    CFG_FLOAT search_cache_budget_gb_ratio;
+
     // While serving the index, the entire graph is stored on SSD. For faster search performance, you can cache a few
     // frequently accessed nodes in memory.
     CFG_FLOAT search_cache_budget_gb;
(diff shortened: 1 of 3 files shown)


src/index/diskann/diskann_config.h
```diff
@@ -148,6 +172,10 @@ class DiskANNConfig : public BaseConfig {
                 if (!search_list_size.has_value()) {
                     search_list_size = kDefaultSearchListSizeForBuild;
                 }
+                pq_code_budget_gb =
+                    std::max(pq_code_budget_gb.value(), pq_code_budget_gb_ratio.value() * vec_field_size_gb.value());
+                search_cache_budget_gb = std::max(search_cache_budget_gb.value(),
+                                                  search_cache_budget_gb_ratio.value() * vec_field_size_gb.value());
```
liliu-z, line 178: The purpose is replacement. Instead of `max`, let's directly use `ratio` once it is provided.

src/index/diskann/diskann_config.h
```diff
@@ -50,6 +55,12 @@ class DiskANNConfig : public BaseConfig {
     // This is the flag to enable fast build, in which we will not build vamana graph by full 2 round. This can
     // accelerate index build ~30% with an ~1% recall regression.
     CFG_BOOL accelerate_build;
+
+    // The ratio of the size reserved for the search cache to the size of the raw data (defined with vec_field_size_gb)
+    // This parameter will replace pq_code_budget_gb to avoid calculating the actual size on the Milvus side.
+    // The index can indirectly obtain search_cache_budget_gb by vec_field_size_gb * search_cache_budget_gb_ratio
+    CFG_FLOAT search_cache_budget_gb_ratio;
```
liliu-z, line 62: Plz help add a deprecated tag in the code.

liliu-z (commented):
Any E2E change?

src/index/diskann/diskann_config.h
foxspy: The main consideration here is compatibility in distributed scenarios. For example, if the parameter-filling code is old, it won't have pq_code_budget_gb_ratio/search_cache_budget_gb_ratio

src/index/diskann/diskann_config.h
foxspy: will do it in the next pr.

foxspy:
> Any E2E change?

Yes, Milvus will remove the calculation logic for the pq_code_budget_gb and search_cache_budget_gb parameters. Any resource-related logic will be bound to vec_field_size_gb. Another reason for this PR is that parameter checks would fail without pq_code_budget_gb/search_cache_budget_gb, so default values were added.

foxspy:
> > Any E2E change?
> 
> Yes, Milvus will remove the calculation logic for the pq_code_budget_gb and search_cache_budget_gb parameters. Any resource-related logic will be bound to vec_field_size_gb. Another reason for this PR is that parameter checks would fail without pq_code_budget_gb/search_cache_budget_gb, so default values were added.

This might be another issue. The parameter check during createIndex does not include runtime parameters, and there may be some differences from the build parameters, though they should be minimal. It seems that index parameters should avoid carrying runtime-specific information as much as possible and be more direct (e.g., directly specifying DiskANN PQ parameters rather than limiting via GB resources). However, managing and retracting exposed parameters can be tricky

alexanderguzhva:
@foxspy would you consider adding something like 'inverse ratio', maybe, because our typical ratios are 1/8, 1/12, 1/16?

foxspy:
> @foxspy would you consider adding something like 'inverse ratio', maybe, because our typical ratios are 1/8, 1/12, 1/16?

Do you mean defining it as an int to ensure precision? But it’s not always possible to guarantee that, for example, in cases like 2/3. Generally, for float parameters, slight precision errors are acceptable.

src/index/diskann/diskann_config.h
liliu-z: then you can pick the original one

alexanderguzhva:
> > @foxspy would you consider adding something like 'inverse ratio', maybe, because our typical ratios are 1/8, 1/12, 1/16?
> 
> Do you mean defining it as an int to ensure precision? But it’s not always possible to guarantee that, for example, in cases like 2/3. Generally, for float parameters, slight precision errors are acceptable.

I remember having a funny experience setting budgets with four or five digits after dot in order to ensure that PQ gets exactly size of 768 for 1536 dim data, in order to avoid PQ 767 or PQ 769.

src/index/diskann/diskann_config.h
foxspy: this will be fixed after https://github.com/milvus-io/milvus/pull/36844 merged. A more clear logic will be provided (choose ratio parameter with high priority)

liliu-z (approved):
/lgtm

PR #915, opened 2024-10-30
fix : add config check for open and closed range settings

foxspy:
issue: https://github.com/milvus-io/milvus/issues/37269

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -59,6 +59,32 @@ typedef nlohmann::json Json;
 #endif
 
 template <typename T>
+struct Range {
+    T left;
+    T right;
+    bool include_left;
+    bool include_right;
+
+    Range(T left, T right, bool includeLeft, bool includeRight)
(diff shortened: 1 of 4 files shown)


include/knowhere/config.h
```diff
@@ -115,6 +115,7 @@ struct Entry<CFG_FLOAT> {
     std::optional<CFG_FLOAT::value_type> default_val;
     uint32_t type;
     std::optional<std::pair<CFG_FLOAT::value_type, CFG_FLOAT::value_type>> range;
+    std::optional<std::pair<bool, bool>> range_included;
```
alexanderguzhva, line 118: please consider packing range into a structure `Range`, which would contain all the logic, including something like `bool Range::within(float value) {}` and `std::string Range::to_string() {}` (which produces a string representation)

foxspy: Good idea~ I will apply it

Issue #927, opened 2024-11-05
sparse vector index mmap mode: too little data mmapped


zhengbuqian:
currently only the raw data is mmapped, the index are still occupying memory

PR #928, opened 2024-11-05
improve sparse vector index mmap: to mmap almost everything

zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/927

kind/improvement

### DIFF
--- a/include/knowhere/sparse_utils.h
+++ b/include/knowhere/sparse_utils.h
@@ -274,4 +274,82 @@ class MaxMinHeap {
     std::vector<SparseIdVal<T>> pool_;
 };  // class MaxMinHeap
 
+// A std::vector like container but uses fixed size free memory(typically from
+// mmap) as backing store and can only be appended at the end.
+//
+// Must be initialized with a valid pointer to memory when used. The memory must be
+// valid during the lifetime of this object. After initialization, GrowableVectorView will
+// have space for mmap_element_count_ elements, none of which are initialized.
+//
+// Currently only used in sparse InvertedIndex. Move to other places if needed.
+template <typename T>
+class GrowableVectorView {
+ public:
+    using value_type = T;
+    using size_type = size_t;
+
+    GrowableVectorView() = default;
+
+    void
+    initialize(void* data, size_type byte_size) {
+        if (byte_size % sizeof(T) != 0) {
+            throw std::invalid_argument("GrowableVectorView byte_size must be a multiple of element size");
+        }
+        mmap_data_ = data;
+        mmap_byte_size_ = byte_size;
+        mmap_element_count_ = 0;
+    }
+
+    size_type
+    capacity() const {
+        return mmap_byte_size_ / sizeof(T);
+    }
+
+    size_type
+    size() const {
+        return mmap_element_count_;
+    }
+
+    template <typename... Args>
+    T&
+    emplace_back(Args&&... args) {
+        if (size() == capacity()) {
+            throw std::out_of_range("emplace_back on a full GrowableVectorView");
+        }
+        auto* elem = reinterpret_cast<T*>(mmap_data_) + mmap_element_count_++;
+        return *new (elem) T(std::forward<Args>(args)...);
+    }
+
+    T&
+    operator[](size_type i) {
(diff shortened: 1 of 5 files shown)


zhengbuqian:
kind/improvement

zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/927

include/knowhere/sparse_utils.h
```diff
@@ -274,4 +274,84 @@ class MaxMinHeap {
...
+//
+// It supports limited operations of std::vector.
+//
+// Must be initialized with a valid pointer to memory when used. The memory must be
+// valid during the lifetime of this object. After initialization, MmapVector will
+// have space for mmap_element_count_ elements, none of which are initialized.
+//
+// Currently only used in sparse InvertedIndex. Move to other places if needed.
+template <typename T>
+class MmapVector {
```
alexanderguzhva, line 288: what is the difference between `MmapVector` and https://en.cppreference.com/w/cpp/container/span ?
Also, technically, this is not a `MmapVector`, but rather a `VectorView` (similar to the relation between `std::string` and `std::string_view`)

src/index/sparse/sparse_inverted_index.h
```diff
@@ -187,68 +205,169 @@ class InvertedIndex : public BaseInvertedIndex<T> {
...
         }
-        auto pos = vals.begin() + static_cast<size_t>(drop_ratio_build * vals.size());
-        // pos may be vals.end() if drop_ratio_build is 1.0, in that case we use
-        // the largest value as the threshold.
-        if (pos == vals.end()) {
-            pos--;
+        // reset reader to the beginning
+        reader.seekg(initial_reader_location);
+
+        auto raw_data_size = rows * sizeof(typename decltype(raw_data_)::value_type);
```
alexanderguzhva, line 236: maybe, it would be nice if variables could be called as `raw_data_size_in_bytes` rather than `raw_data_size`, because the default convension is that `*_size` is measured in elements, not in bytes.

No big deal, but it just improves the readibility of the code.

include/knowhere/sparse_utils.h
zhengbuqian: CMIIW `std::span` is more like referencing a list of initialized objects? so `std::span()::size()` always equals to its capacity. while in this case the contents of the containers are built up gradually using `emplace_back` and `size()` grows from 0 to capacity. though the initialization is still all done in the `Load` method. 

If we want to use `std::span`:

```c++
template <typename T>
class FixedVector : public std::span<T> {

    emplace_back() {  // to have a consistent API like vector so we can mix the usage with std::vector
    }

    size() override {
        // return number of constructed elements instead of `byte_size / sizeof(T)`.
    }
}
```

Another option is to avoid overloading the meaning of `size()`(and also avoid subclassing it) by first placement new all objects on the raw ptr, and then just create the spans out of them. But then we won't be able to reuse `add_row_to_index`.

WDYT? 

And yes it is not a `MmapedVector` indeed. But it is also not a view: it is more like an append-only vector with pre-allocated unextendable memory. How about `FixedVector`?

alexanderguzhva: I see. Basically, I got confused by the name of this class :) It totally makes sense to have this separate entity.
Naming is tricky
* it must include the word `view` / `span` to indicate that an object instance does not own the memory (unlike vector or array that do)
* the word `vector` needs to be included to indicate that this is a contiguous data chunk
* the word `array` is less appropriate, bcz it usually indicates a fixed size, not kinda `growable` as of in our use case

Maybe, something like `VectorLikeView` or `GrowableVectorView`?

alexanderguzhva:
@zhengbuqian please let me know if I should lgtm this diff. It looks good, overall.

include/knowhere/sparse_utils.h
zhengbuqian: Using `GrowableVectorView`, liking it over others.

zhengbuqian:
> @zhengbuqian please let me know if I should lgtm this diff. It looks good, overall.

it is ready now, thanks!

liliu-z (approved):
/lgtm

PR #929, opened 2024-11-05
Allows FAISS_HNSW_FLAT index to load & search on hnswlib flat index

alexanderguzhva:
It is impossible to load hnswlib data into a faiss structure, bcz it causes problems with the proper mmap functionality.
So, I've introduced a new `IndexNode` class that holds two underlying `IndexNode` instances and decides which one to use, based on deserialization results. Basically, the one of two `IndexNode` instances that gets successfully deserialized will be used in search, range search and ann iterator. The key is that a serialized FAISS index data contains a four-byte identifier of a type of an index in the very beginning (such as `IxF2`), which allows Faiss deserialization functions to quickly and reliably verify whether an index is a faiss one or not. So, we deserialize the data as Faiss HNSW Flat index first. And if it fails, then the wrapper will try to deserialize as hnswlib one.


src/index/hnsw/faiss_hnsw.cc
```diff
@@ -2053,7 +2290,7 @@ class BaseFaissRegularIndexHNSWPRQNodeTemplate : public BaseFaissRegularIndexHNS
 };
 
 //
-KNOWHERE_SIMPLE_REGISTER_DENSE_FLOAT_ALL_GLOBAL(FAISS_HNSW_FLAT, BaseFaissRegularIndexHNSWFlatNodeTemplate,
+KNOWHERE_SIMPLE_REGISTER_DENSE_FLOAT_ALL_GLOBAL(FAISS_HNSW_FLAT, BaseFaissRegularIndexHNSWFlatNodeTemplateWithSearchFallback,
```
foxspy, line 2293: This name needs to be replaced with "HNSW" so that all HNSW requests will be forwarded here.

foxspy:
It is good to make a unit test to verify that the index built with old HNSW can be loaded and searched with faiss_hnw.

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: would it cause conflicts with an existing registration for hnswlib's implementation?

alexanderguzhva:
> It is good to make a unit test to verify that the index built with old HNSW can be loaded and searched with faiss_hnw.

Is it the only place that need to be patched in the whole project?
p.s. still unpatched, please just let me know and I'll perform the patch

src/index/hnsw/faiss_hnsw.cc
foxspy: > would it cause conflicts with an existing registration for hnswlib's implementation?

You must take this approach; otherwise, faiss_hnsw_flat won’t be called, and we don’t plan to introduce a new name. You can rename the original HNSW as HNSW_DEPRECATED（I’ll work on handling CARDINAL on my end）

foxspy:
> > It is good to make a unit test to verify that the index built with old HNSW can be loaded and searched with faiss_hnw.
> 
> Is it the only place that need to be patched in the whole project? p.s. still unpatched, please just let me know and I'll perform the patch



> It is good to make a unit test to verify that the index built with old HNSW can be loaded and searched with faiss_hnw.

@alexanderguzhva  Is this PR ready to merge? I can help with the naming issues since it might get a bit tricky here. If so, you can remove the WIP in the title

alexanderguzhva:
@foxspy I think that it is ready

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -140,6 +141,20 @@ class BaseFaissIndexNode : public IndexNode {
...
+// returns true if the text of FaissException is about non-recognizing fourcc
+static inline bool
+is_faiss_fourcc_error(const char* what) {
+    if (what == nullptr) {
+        return false;
+    }
+
+    std::string error_msg(what);
+
+    // check if this is fourCC problem
```
foxspy, line 153: Using error_msg for judgment can be risky; I suggest using a specific exception instead.

liliu-z (approved):
/lgtm

Issue #921, opened 2024-11-04
Replace HNSW  with FAISS_HNSW_FLAT


foxspy:
Faiss has added flat support for HNSW and supports SQ/PQ/PRQ, which currently results in two HNSW options within knowhere, both offering identical parameters and capabilities but with different naming (FAISS_HNSW_FLAT vs. HNSW). This dual support increases future maintenance costs (since both sets of HNSW need to be maintained) and adds user complexity (users must rebuild and replace indexes).

If the goal is to use FAISS_HNSW to replace HNSW, a compatibility solution should be provided—allowing FAISS_HNSW to load pre-existing HNSW indexes and support search/range_search/iterator capabilities. This would centralize all HNSW-related logic into FAISS_HNSW, automatically transitioning users to FAISS_HNSW after upgrading.

Two potential approaches to address this:

* Code Compatibility Handling
1. Remove hnsw.cc and associated hnswlib code.
2. Rename FAISS_HNSW_FLAT to "HNSW" and route HNSW-related requests to FAISS_HNSW.
3. When deserializing, FAISS_HNSW should recognize HNSW binaries and load them compatibly, handling iterator/search/range_search requests.

* Version Isolation
1. Rename HNSW in hnsw.cc to HNSWXXX.
2. Upgrade the index version (from 5 to 6).
3. Rename FAISS_HNSW_FLAT to "HNSW" and route HNSW-related requests to FAISS_HNSW.
FAISS_HNSW should delegate requests to HNSWXXX for versions ≤ 5; for versions > 5, FAISS_HNSW should handle them internally.

foxspy:
@alexanderguzhva faiss_hnsw is part of the 2.5.0 release planned for early next week, it would be ideal to address this soon.

alexanderguzhva:
@foxspy already testing

PR #930, opened 2024-11-07
enhance : use faiss hnsw implementation to replace hnswlib

foxspy:
issue: #921


foxspy:
@alexanderguzhva It seems that HNSW_PQ has some problem with its iterator interface. You can replay it when you remove the comment line for `INDEX_HNSW_PQ` within the `TEST_CASE("Test Iterator Mem Index With Float Vector", "[float metrics]")`

liliu-z (approved):
/lgtm

PR #932, opened 2024-11-07
enable refine for FAISS HNSW indices by default, if available

alexanderguzhva:
* Enable refine for FAISS HNSW indices by default, if available. This fixes the problem in UT, where a non-refined search for refined FAISS HNSW index would not match refined iterator search 
* Remove duplicate test cases in UT
* Fixes refining for the range search (https://github.com/facebookresearch/faiss/pull/4022)

### DIFF
--- a/src/index/hnsw/faiss_hnsw_config.h
+++ b/src/index/hnsw/faiss_hnsw_config.h
@@ -43,7 +43,7 @@ class FaissHnswConfig : public BaseHnswConfig {
             .for_static();
         KNOWHERE_CONFIG_DECLARE_FIELD(refine_k)
             .description("refine k")
-            .allow_empty_without_default()
+            .set_default(1)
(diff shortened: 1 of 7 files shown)


src/index/hnsw/faiss_hnsw_config.h
```diff
@@ -43,9 +43,10 @@ class FaissHnswConfig : public BaseHnswConfig {
             .for_static();
         KNOWHERE_CONFIG_DECLARE_FIELD(refine_k)
             .description("refine k")
-            .allow_empty_without_default()
+            .set_default(1)
             .set_range(1, std::numeric_limits<CFG_FLOAT::value_type>::max())
-            .for_search();
+            .for_search()
```
foxspy, line 48: for_iterator?

alexanderguzhva: removed `.for_range_search()`. Technically, `range_k` is applicable only for the search, neither range_search, not iterator.

PR #933, opened 2024-11-11
Add SparseInvertedIndexNodeCC

zhengbuqian:
Add SparseInvertedIndexNodeCC to allow being thread safe growing index

move sparse index Add to build pool

/kind improvement

### DIFF
--- a/src/index/sparse/sparse_index_node.cc
+++ b/src/index/sparse/sparse_index_node.cc
@@ -30,14 +30,16 @@ namespace knowhere {
 
 // Inverted Index impl for sparse vectors. May optionally use WAND algorithm to speed up search.
 //
-// Not overriding RangeSerach, will use the default implementation in IndexNode.
+// Not overriding RangeSearch, will use the default implementation in IndexNode.
+//
+// Thread safety: not thread safe.
 template <typename T, bool use_wand>
 class SparseInvertedIndexNode : public IndexNode {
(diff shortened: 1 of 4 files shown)


src/index/sparse/sparse_index_node.cc
```diff
@@ -316,14 +327,127 @@ class SparseInvertedIndexNode : public IndexNode {
...
+template <typename T, bool use_wand>
+class SparseInvertedIndexNodeCC : public SparseInvertedIndexNode<T, use_wand> {
+ public:
+    explicit SparseInvertedIndexNodeCC(const int32_t& version, const Object& object)
+        : SparseInvertedIndexNode<T, use_wand>(version, object) {
+    }
+
+    Status
+    Add(const DataSetPtr dataset, std::shared_ptr<Config> cfg) override {
+        std::unique_lock<std::mutex> lock(mutex_);
```
alexanderguzhva, line 349: can `std::shared_mutex<>` from C++17 (https://en.cppreference.com/w/cpp/thread/shared_mutex) solve the problem of concurrent access for readers and writers in this particular use case?

zhengbuqian: we want to avoid the starvation of writers. this impl guarantees that new read requests have to wait for already waiting write requests to finish.

alexanderguzhva: lgtm's, I'm assuming that you've checked that it won't deadlock

PR #941, opened 2024-11-13
enhance: BF supports ids of base_data starting from a specific value

cqy123456:
issue: https://github.com/zilliztech/knowhere/issues/942

### DIFF
--- a/include/knowhere/bitsetview_idselector.h
+++ b/include/knowhere/bitsetview_idselector.h
@@ -19,14 +19,16 @@ namespace knowhere {
 
 struct BitsetViewIDSelector final : faiss::IDSelector {
     const BitsetView bitset_view;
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@cqy123456 lgtm, what is the practical use case for this optimization? Basically, I wonder why this change was implemented for the brute-force search only

cqy123456:
Raw data within a field is discontinuous in milvus. If the rows number in a raw data block is not aligned with 128, we need to copy a part of the bitset to knowhere. We also need to transfer Knowhere's default ID to an external ID. This PR is to enable Knowhere to obtain external IDs, reduce the copying of bitsets, and implement ID conversion.

foxspy (approved):
/lgtm

Issue #925, opened 2024-11-05
Optimize BruteForce::Search Performance with cosine metric


foxspy:
BruteForce::Search will compute the norms during search stage, which has some performance degradation for serveral reason
1. miss avx support for `fvec_norm_L2sqr`
2. compute the norms of base vectors multiple times when the number of query vectors is more than one

PR #950, opened 2024-11-15
enhance: optimize get norms function with simd and reduce some calculations

cqy123456:
issue: https://github.com/zilliztech/knowhere/issues/925


foxspy (approved):
/lgtm

PR #952, opened 2024-11-16
fix: Isolate faiss_hnsw and hnsw by index version

foxspy:
issue: https://github.com/milvus-io/milvus/issues/37665
Considering the compatibility during the distributed system upgrade, faiss_hnsw and hnswlib must be isolated by version on both the read and write path; for version <= 5, these indexes will be processed by hnswlib; otherwise, they will be processed by faiss; the maximum availability of index version during the upgrade process has been guaranteed in the early stage


liliu-z (approved):
/lgtm

Issue #967, opened 2024-11-29
Improve WAND algorithm of sparse vector index


zhengbuqian:
this is an umbrella issue for improvements to make for the WAND sparse vector index, including but not limited to:

* algorithm impl improvement
* support for BMW and other variants

PR #968, opened 2024-11-29
wand: optimize cursor class

zhengbuqian:
also simplifed code to drop less important elements

simple benchmark result(time used to search all queries once on the full dataset) with drop ratio build 0.32 and drop ratio search 0.6:

||Before|After|Improvement %|
|---|---|---|---|
|[bigann SPLADE vectors of MSMarco](https://github.com/harsha-simhadri/big-ann-benchmarks/blob/main/dataset_preparation/sparse_dataset.md)|141,473ms|89,604ms|36%|
|Milvus 2.5 Full Text Search Vectors of MSMarco|4,758ms|3,154ms|34%|

### DIFF
--- a/src/index/sparse/sparse_inverted_index.h
+++ b/src/index/sparse/sparse_inverted_index.h
@@ -360,15 +360,7 @@ class InvertedIndex : public BaseInvertedIndex<T> {
                     vals.push_back(fabs(data[i][j].val));
                 }
             }
-            auto pos = vals.begin() + static_cast<size_t>(drop_ratio_build * vals.size());
-            // pos may be vals.end() if drop_ratio_build is 1.0, in that case we use
-            // the largest value as the threshold.
-            if (pos == vals.end()) {
-                pos--;
-            }
-            std::nth_element(vals.begin(), pos, vals.end());
-
-            value_threshold_ = *pos;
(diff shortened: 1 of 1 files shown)


src/index/sparse/sparse_inverted_index.h
```diff
@@ -561,74 +565,77 @@ class InvertedIndex : public BaseInvertedIndex<T> {
...
         float q_value_ = 0.0f;
         const BitsetView bitset_;
-    };  // class Cursor
+        table_t cur_vec_id_ = 0;
+
+     private:
+        inline void
+        update_cur_vec_id() {
+            if (loc_ >= lut_size_) {
+                cur_vec_id_ = total_num_vec_;
```
alexanderguzhva, line 617: just double-checking that this is not `total_num_vec_ - 1`. According to standard naming conventions, `id_` is assumed to be something within `[0, num)` range.

zhengbuqian: yes it is not `total_num_vec_ - 1`. When `id_` is in `[0, num)` range it is still in range and the cursor is pointing at some doc, `loc_ >= lut_size_` means it is out of range and the cursor points at no doc.

zhengbuqian:
issue: https://github.com/zilliztech/knowhere/issues/967

PR #970, opened 2024-12-03
[WIP] sparse: use staged cursor and upper_bound for WAND

sparknack:
Use staged cursor and upper_bound to avoid recalculating the score from cursor 0 every time in the loop. And modify some variable names to be more readable.

Simple benchmark result(time used to search all queries once on the full dataset) with drop ratio build 0.32 and drop ratio search 0.6:

| BM25 |Before|After|Improvement %|
|---|---|---|---|
|[bigann SPLADE vectors of MSMarco](https://github.com/harsha-simhadri/big-ann-benchmarks/blob/main/dataset_preparation/sparse_dataset.md)|81,980ms|66,205ms|19.2%|
|Milvus 2.5 Full Text Search Vectors of MSMarco|1,005ms|1,034ms|~|

| IP |Before|After|Improvement %|
|---|---|---|---|
|[bigann SPLADE vectors of MSMarco](https://github.com/harsha-simhadri/big-ann-benchmarks/blob/main/dataset_preparation/sparse_dataset.md)|85,527ms|75,366ms|12%|
|Milvus 2.5 Full Text Search Vectors of MSMarco(using IP/TF-IDF metric though, not BM25)|2,839ms|3,191ms|-12%|

This optimization seems to have an improvement on learned sparse embedding like SPLADE, which is denser, but leads to a little bit of performance degration on more sparse embeddings.

Hold the review of this PR for now until more investigation.

### DIFF
--- a/src/index/sparse/sparse_inverted_index.h
+++ b/src/index/sparse/sparse_inverted_index.h
@@ -654,16 +654,16 @@ class InvertedIndex : public BaseInvertedIndex<T> {
             std::sort(cursors.begin(), cursors.end(), [](auto& x, auto& y) { return x->cur_vec_id_ < y->cur_vec_id_; });
         };
         sort_cursors();
+        float upper_bound = 0;
+        size_t pivot_cursor_id = 0;
         while (true) {
             float threshold = heap.full() ? heap.top().val : 0;
-            float upper_bound = 0;
-            size_t pivot;
             bool found_pivot = false;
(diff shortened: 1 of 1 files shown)


sparknack:
issue: #967

sparknack:
> @sparknack is it good to review the PR?

since maxscore algorithum is better than wand, this optimization is downgraded.

Issue #971, opened 2024-12-03
ut failed to run after the latest submission


sparknack:
![image](https://github.com/user-attachments/assets/63c6b7bc-d690-42cc-922f-6841b16ab35d)

https://github.com/zilliztech/knowhere/actions/runs/12131996132/job/33825138701?pr=970

alexanderguzhva:
@sparknack @Presburger  the story isn't over, now it is causing problems in pyknowhere
```
+ pytest -v -m L0

ImportError while loading conftest '/home/jenkins/agent/workspace/knowhere_kn2_sse_PR-969/tests/conftest.py'.

conftest.py:1: in <module>

    import knowhere

/usr/local/lib/python3.11/dist-packages/knowhere/__init__.py:1: in <module>

    from . import swigknowhere

/usr/local/lib/python3.11/dist-packages/knowhere/swigknowhere.py:13: in <module>

    from . import _swigknowhere

E   ImportError: libglog.so.1: cannot open shared object file: No such file or directory

script returned exit code 4
```

PR #972, opened 2024-12-03
fix: conan: make glog shared

sparknack:
This change is aimed to fix knowhere_tests running error after latest modification.


sparknack:
issue: #971

sparknack:
kind bug

PR #990, opened 2024-12-17
Optimize feature registry

cydrain:


src/index/index_factory.cc
```diff
@@ -84,7 +84,7 @@ IndexFactory::Register(const std::string& name, std::function<Index<IndexNode>(c
         feature_mapping_[name] = features;
     } else {
         // All data types should have the same features; please try to avoid breaking this rule.
-        feature_mapping_[name] = feature_mapping_[name] & features;
+        feature_mapping_[name] |= features;
```
foxspy, line 87: There is a problem here; if a certain data type supports a feature, then all types will be considered to support this feature; this is unreasonable.This is why `ALL_TYPE` is used above.

cydrain: Current feature mark mechanism cannot handle different data type support different feature for a same index type.
If it happens, the whole feature mark mechanism need be re-designed.

foxspy (approved):
/lgtm

Issue #977, opened 2024-12-10
Faiss native support multi data types


cydrain:
can refer https://github.com/zilliztech/knowhere/pull/746

- [x] 1. Faiss HNSW support int8 (#991)
- [ ] 2. BruteForce support int8 (wait Flat_NM support)
- [ ] 3. Faiss IVF_FLAT support fp16/bf16/int8
- [ ] 4. Faiss IVF_SQ8 change to IVF_SQ
- [ ] 5. Faiss IVF_SQ support fp16/bf16/int8
- [ ] 6. Faiss IVF_PQ support fp16/bf16/int8

PR #991, opened 2024-12-17
faiss_hnsw support INT8

cydrain:
Issue: #977

### DIFF
--- a/src/index/hnsw/faiss_hnsw.cc
+++ b/src/index/hnsw/faiss_hnsw.cc
@@ -292,7 +292,7 @@ class BaseFaissRegularIndexNode : public BaseFaissIndexNode {
 };
 
 //
-enum class DataFormatEnum { fp32, fp16, bf16 };
+enum class DataFormatEnum { fp32, fp16, bf16, int8 };
 
 template <typename T>
 struct DataType2EnumHelper {};
@@ -309,14 +309,16 @@ template <>
 struct DataType2EnumHelper<knowhere::bf16> {
     static constexpr DataFormatEnum value = DataFormatEnum::bf16;
 };
+template <>
+struct DataType2EnumHelper<knowhere::int8> {
+    static constexpr DataFormatEnum value = DataFormatEnum::int8;
+};
 
 template <typename T>
 static constexpr DataFormatEnum datatype_v = DataType2EnumHelper<T>::value;
 
-//
 namespace {
 
-//
 bool
 convert_rows_to_fp32(const void* const __restrict src_in, float* const __restrict dst,
                      const DataFormatEnum src_data_format, const size_t start_row, const size_t nrows,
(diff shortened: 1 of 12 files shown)


src/index/hnsw/faiss_hnsw.cc
```diff
@@ -357,21 +362,27 @@ convert_rows_from_fp32(const float* const __restrict src, void* const __restrict
...
         knowhere::fp32* const dst = reinterpret_cast<knowhere::fp32*>(dst_in);
         for (size_t i = 0; i < nrows * dim; i++) {
             dst[i + start_row * dim] = src[i];
         }
-
+        return true;
+    } else if (dst_data_format == DataFormatEnum::int8) {
+        knowhere::int8* const dst = reinterpret_cast<knowhere::int8*>(dst_in);
+        for (size_t i = 0; i < nrows * dim; i++) {
+            KNOWHERE_THROW_IF_NOT_MSG(src[i] >= std::numeric_limits<knowhere::int8>::min() &&
```
alexanderguzhva, line 381: it is better to use `std::numeric_limilts<knowhere::int8>::lowest()` here

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -1327,6 +1350,9 @@ class BaseFaissRegularIndexHNSWFlatNode : public BaseFaissRegularIndexHNSWNode {
             } else if (data_format == DataFormatEnum::bf16) {
                 hnsw_index = std::make_unique<faiss::IndexHNSWSQCosine>(dim, faiss::ScalarQuantizer::QT_bf16,
                                                                         hnsw_cfg.M.value());
+            } else if (data_format == DataFormatEnum::int8) {
+                hnsw_index = std::make_unique<faiss::IndexHNSWSQCosine>(
+                    dim, faiss::ScalarQuantizer::QT_8bit_direct_signed, hnsw_cfg.M.value());
```
alexanderguzhva, line 1355: please **DO** confirm that you want to use `QT_8bit_direct_signed` here, because the use case is not clear to me. Basically, I can imagine a use case that works with the input data of `[0..255]` range (`QT_8bit_direct`), or the traditional `QT_8bit` that remaps input float values into `[0..255]` range, but what is the use case for the input data of `[-128..127]` range? Or is it just the requirement from Milvus?

alexanderguzhva (commented):
lgtm overall, but please confirm about `QT_8bit_direct_signed`. Thanks.

src/index/hnsw/faiss_hnsw.cc
cydrain: Hi Alex, what's the difference between min() and lowest() here?

src/index/hnsw/faiss_hnsw.cc
cydrain: it's the requirement from Milvus, since vespa and qdrant already support Vector_Int8 now

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: there's no difference for this particular use case. But `lowest()` is better to use, because of the connotations with `std::numeric<float>::lowest()` (which is -1e+40, the least value) and `std::numeric<float>::min()` (which is 1e-40, the least representable positive value)

alexanderguzhva:
@cydrain lgtm
please let me know if you'd like to change `min()` to `lowest()` and I'll lgtm this diff in either cases

src/index/hnsw/faiss_hnsw.cc
cydrain: Hi Alex, I see no obvious difference between min() and lowest(), I prefer to use min() and max() in pair.

cydrain:
sift-128-euclidean

                         | index file size | build time | QPS
---------------------|--------------------|---------------|-----------
HNSW (FP32)  | 626 M             | 103.594s  |
HNSW (FP16)  | 382 M             | 86.145s    |
HNSW (INT8)  | 260 M              | 74.623s    |

PR #996, opened 2024-12-18
Enable mmap-ped Faiss indices (IndexFlatCodes-based and HNSW)

alexanderguzhva:


thirdparty/faiss/faiss/impl/mapped_io.cpp
```diff
@@ -0,0 +1,127 @@
...
+        void* address = mmap(nullptr, filesize, PROT_READ, MAP_SHARED, fileno(f), 0);
+        FAISS_THROW_IF_NOT_FMT(
+            address != nullptr, "could not mmap(): %s", strerror(errno)
+        );
+
+        // btw, fd can be closed here
+
+        // set 'random' access pattern
+        // todo: check the error
+        madvise(address, filesize, MADV_RANDOM);
```
foxspy, line 43: a warn log needs to be added here

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -512,7 +594,8 @@ static void read_HNSW(HNSW* hnsw, IOReader* f) {
     READVECTOR(hnsw->cum_nneighbor_per_level);
     READVECTOR(hnsw->levels);
```
foxspy, line 595: Don't these require mmap? Although these only account for a small part

foxspy (approved):
/lgtm

thirdparty/faiss/faiss/impl/index_read.cpp
chasingegg: same question, what do we exactly mmap for hnsw @alexanderguzhva

PR #1001, opened 2024-12-20
fix : fix cardinal compile error

foxspy:
/root/milvus/cmake_build/thirdparty/knowhere/knowhere-src/include/knowhere/index/index_factory.h:71:38: error: redefinition of 'const knowhere::IndexFactory& knowhere::index_factory_ref_HNSWfp32'


foxspy:
verified locally

Issue #1004, opened 2024-12-24
sparse index not properly closing index file when using mmap


zhengbuqian:
`knowhere::FileReader` does not close fd in destructor, instead it provides a close method.

PR #1005, opened 2024-12-24
make FileReader to close fd using RAII

zhengbuqian:
/kind bug

issue: https://github.com/zilliztech/knowhere/issues/1004
#1004


alexanderguzhva:
issue: https://github.com/zilliztech/knowhere/issues/1004

Issue #997, opened 2024-12-19
Use search_pool to control iterator->Next()


alwayslove2013:
The current `AnnIterator` function utilizes the search pool for concurrency control only when initializing the `Iterator`. Once the returned iterator is handed over to the upper layer, the `->next()` calls are not subjected to any thread restrictions, which may lead to issues such as OMP conflicts.

To address this, we propose the following considerations:
- All iterators will accept a `use_knowhere_search_pool` parameter during construction. 
  - When set to `True` (the default), the `iterator->next()` will be scheduled by the `knowhere_search_thread_pool`. 
  - When set to `False`, `iterator->next()` will not involve thread scheduling internally, so please take caution.
- The initialization of `iterator` in the `AnnIterator` function will no longer be concurrent, which helps to avoid potential deadlocks.
  - Furthermore, to enhance performance for large_nq, we will try to streamline the initialization process for all iterators by **deferring heavier pre-computation tasks** to the first call of `->next()`.

PR #1008, opened 2024-12-26
Use search_pool to control iterator->Next()

alwayslove2013:
issue: #997 

> The current AnnIterator function utilizes the search pool for concurrency control only when initializing the Iterator. Once the returned iterator is handed over to the upper layer, the ->next() calls are not subjected to any thread restrictions, which may lead to issues such as OMP conflicts.
> 
> To address this, we propose the following considerations:
> - All iterators will accept a use_knowhere_search_pool parameter during construction.
>   - When set to True (the default), the iterator->next() will be scheduled by the knowhere_search_thread_pool.
>   - When set to False, iterator->next() will not involve thread scheduling internally, so please take caution.
> - The initialization of iterator in the AnnIterator function will no longer be concurrent, which helps to avoid potential deadlocks.
>   - Furthermore, to enhance performance for large_nq, we will try to streamline the initialization process for all iterators by deferring heavier pre-computation tasks to the first call of ->next().

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -146,7 +146,8 @@ class IndexNode : public Object {
     using IteratorPtr = std::shared_ptr<iterator>;
 
     virtual expected<std::vector<IteratorPtr>>
-    AnnIterator(const DataSetPtr dataset, std::unique_ptr<Config> cfg, const BitsetView& bitset) const {
+    AnnIterator(const DataSetPtr dataset, std::unique_ptr<Config> cfg, const BitsetView& bitset,
+                bool use_knowhere_search_pool = true) const {
         return expected<std::vector<std::shared_ptr<iterator>>>::Err(
             Status::not_implemented, "annIterator not supported for current index type");
     }
@@ -202,7 +203,10 @@ class IndexNode : public Object {
             return GenResultDataSet(nq, std::move(range_search_result));
         }
 
-        auto its_or = AnnIterator(dataset, std::move(cfg), bitset);
+        // The range_search function has utilized the search_pool to concurrently handle various queries.
+        // To prevent potential deadlocks, the iterator for a single query no longer requires additional thread
+        //   control over the next() call.
+        auto its_or = AnnIterator(dataset, std::move(cfg), bitset, false);
         if (!its_or.has_value()) {
             return expected<DataSetPtr>::Err(its_or.error(),
                                              "RangeSearch failed due to AnnIterator failure: " + its_or.what());
@@ -290,13 +294,17 @@ class IndexNode : public Object {
         futs.reserve(nq);
         if (retain_iterator_order) {
             for (size_t i = 0; i < nq; i++) {
-                futs.emplace_back(
-                    ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() { task_with_ordered_iterator(idx); }));
+                futs.emplace_back(ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() {
+                    ThreadPool::ScopedSearchOmpSetter setter(1);
+                    task_with_ordered_iterator(idx);
+                }));
             }
         } else {
             for (size_t i = 0; i < nq; i++) {
-                futs.emplace_back(ThreadPool::GetGlobalSearchThreadPool()->push(
-                    [&, idx = i]() { task_with_unordered_iterator(idx); }));
+                futs.emplace_back(ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() {
+                    ThreadPool::ScopedSearchOmpSetter setter(1);
+                    task_with_unordered_iterator(idx);
+                }));
             }
         }
         WaitAllSuccess(futs);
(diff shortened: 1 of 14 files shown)


include/knowhere/index/index_node.h
```diff
@@ -202,7 +203,7 @@ class IndexNode : public Object {
             return GenResultDataSet(nq, std::move(range_search_result));
         }
 
-        auto its_or = AnnIterator(dataset, std::move(cfg), bitset);
+        auto its_or = AnnIterator(dataset, std::move(cfg), bitset, false);
```
alexanderguzhva, line 206: Please add a comment about why this parameter is `false`

include/knowhere/index/index_node.h
```diff
@@ -437,6 +442,10 @@ class IndexNode : public Object {
 //   with quantization, override `raw_distance`.
 // Internally, this structure uses the same priority queue class, but may multiply all
 //   incoming distances to (-1) value in order to turn max priority queue into a min one.
+// If use_knowhere_search_pool is True (the default), the iterator->Next() will be scheduled by the
+//   knowhere_search_thread_pool.
+//   If False, will Not involve thread scheduling internally, so please take caution.
+template <bool use_knowhere_search_pool = true>
```
alexanderguzhva, line 448: What are the justifications for adding `use_knowhere_search_pool` as a template parameter instead of making a regular `IndexIterator::use_knowhere_search_pool` field? I'm not sure I follow

src/common/comp/brute_force.cc
```diff
@@ -786,84 +786,85 @@ BruteForce::AnnIterator(const DataSetPtr base_dataset, const DataSetPtr query_da
...
-                    LOG_KNOWHERE_ERROR_ << "Invalid metric type: " << cfg.metric_type.value();
-                    return Status::invalid_metric_type;
                 }
-            }
-            if (xb_id_offset != 0) {
-                for (auto& distances_id : distances_ids) {
-                    distances_id.id = distances_id.id == -1 ? -1 : distances_id.id + xb_id_offset;
+                if (xb_id_offset != 0) {
+                    for (auto& distances_id : distances_ids) {
+                        distances_id.id = distances_id.id == -1 ? -1 : distances_id.id + xb_id_offset;
```
alexanderguzhva, line 853: just to confirm: this is evaluated as `distances_id.id = a ? b : c`, not `distances_id.id = (a ? b : c) + xb_id_offset'?

alwayslove2013: @cqy123456 help check~

include/knowhere/index/index_node.h
alwayslove2013: added. 
> The range_search function has utilized the search_pool to concurrently handle various queries.
> To prevent potential deadlocks, the iterator for a single query no longer requires additional thread control over the next() call.

include/knowhere/index/index_node.h
alwayslove2013: Just to reduce the judgment of `use_knowhere_search_pool` in each `iterator->next()` call.

src/common/comp/brute_force.cc
```diff
@@ -786,84 +782,88 @@ BruteForce::AnnIterator(const DataSetPtr base_dataset, const DataSetPtr query_da
...
-                    break;
-                }
-                case faiss::METRIC_INNER_PRODUCT: {
-                    if (is_cosine) {
+    std::shared_ptr<float[]> norms = GetVecNorms<DataType>(base_dataset);
+
+    try {
+        for (int i = 0; i < nq; ++i) {
+            auto compute_dist_func = [=]() -> std::vector<DistId> {
+                auto xb = base_dataset->GetTensor();
```
PwzXxm, line 792: maybe add comments on why we cannot do it outside

src/common/comp/brute_force.cc
PwzXxm: yes, `-1`  or `distances_id.id + xb_id_offset`

src/common/comp/brute_force.cc
alwayslove2013: added.

include/knowhere/index/index_node.h
alexanderguzhva: I've re-read the code again and I don't believe I see a reason for making `use_knowhere_search_pool` a template parameter. Basically, it leads to increase of the size of the compiled code, also this parameter is not generic enough (like `T` in `std::vector<T>`) and it does not have a critical-performance impact. Please consider demoting `use_knowhere_search_pool` to a variable. Or I can accept the code as is.

alwayslove2013: Thanks for your suggestion~ I will use `use_knowhere_search_pool` as an iterator variable rather than a template param.

include/knowhere/index/index_node.h
```diff
@@ -290,13 +294,17 @@ class IndexNode : public Object {
         futs.reserve(nq);
         if (retain_iterator_order) {
             for (size_t i = 0; i < nq; i++) {
-                futs.emplace_back(
-                    ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() { task_with_ordered_iterator(idx); }));
+                futs.emplace_back(ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() {
+                    ThreadPool::ScopedSearchOmpSetter setter(1);
```
liliu-z, line 298: Can we rewrite the push to include this omp setter instead of calling it everywhere

include/knowhere/index/index_node.h
```diff
@@ -437,48 +445,71 @@ class IndexNode : public Object {
...
           retain_iterator_order_(retain_iterator_order),
+          use_knowhere_search_pool_(use_knowhere_search_pool),
           sign_(larger_is_closer ? -1 : 1) {
     }
 
     std::pair<int64_t, float>
     Next() override {
         if (!initialized_) {
-            throw std::runtime_error("Next should not be called before initialization");
+            initialize();
```
liliu-z, line 465: Curios why we need this

include/knowhere/index/index_node.h
```diff
@@ -540,46 +571,102 @@ class IndexIterator : public IndexNode::iterator {
...
-    }
-
-    // Construct an iterator from a list of distances with index being id, filtering out zero distances.
-    PrecomputedDistanceIterator(const std::vector<float>& distances, bool larger_is_closer)
-        : larger_is_closer_(larger_is_closer) {
-        // 30% is a ratio guesstimate of non-zero distances: probability of 2 random sparse splade vectors(100 non zero
-        // dims out of 30000 total dims) sharing at least 1 common non-zero dimension.
-        results_.reserve(distances.size() * 0.3);
-        for (size_t i = 0; i < distances.size(); i++) {
-            if (distances[i] != 0) {
```
liliu-z, line 562: Why we do this? I think this `PrecomputedDistanceIterator` is used for both dense/sparse BF
@zhengbuqian

include/knowhere/index/index_node.h
alwayslove2013: @cqy123456 Is this feasible? Are there any scenarios where we do not need the `OmpSetter`?

include/knowhere/index/index_node.h
alwayslove2013: Throwing error is a bit too strict?

include/knowhere/index/index_node.h
alwayslove2013: submitted a new commit to reduce misunderstanding.

include/knowhere/index/index_node.h
cqy123456: OmpSetter only use in faiss index.

alwayslove2013:
issue: https://github.com/zilliztech/knowhere/issues/997

foxspy (approved):
/lgtm

alexanderguzhva (approved):

PR #1011, opened 2024-12-30
memory usage optimization for sparse vector

sparknack:
This patch series adds two memory optimization methods:

1. Remove the raw data cache in sparse inverted index. Note that config param `drop_ratio_build` and GetVectorByIds() are also removed.
2. Force quantize float to uint16_t for BM25 inverted index.

PR #1017, opened 2025-01-08
sparse: accelerate the writing of index files

sparknack:
Since we remove forward index in #1011 , sparse row data for index file should reconstructed from inverted index. But it is too slow to construct the index file using the DAAT cursor method. Switch to the TAAT traversal method instead.


sparknack:
issue: #967

PR #1018, opened 2025-01-08
sparse: add max score ratio downscaling for approximate searching

sparknack:
Reuse `wand_bm25_max_score_ratio` for approximate searching.

```
wand_bm25_max_score_ratio is assigned two functions:

1. to upscale the max score to compensate for avgdl changes
The term frequency part of score of BM25 is:
tf * (k1 + 1) / (tf + k1 * (1 - b + b * (doc_len / avgdl)))
as more documents being added to the collection, avgdl can also
change. In WAND index we precompute and cache this score in order to
speed up the search process, but if avgdl changes, we need to
re-compute such score which is expensive. To avoid this, we upscale
the max score by a ratio to compensate for avgdl changes. This will
make the max score larger than the actual max score, it makes the
filtering less aggressive, but guarantees the correctness.

2. to downscale the max score for approximate searching
In the searching process, we use the sum of the max scores to
filter the candidate vectors. If the sum is smaller than the
threshold, skip current vector. If approximate searching is enabled,
we can make the skipping more aggressive by downscaling the max
score. Since the possibility that the maxscore of all dims in the
query appears on the same vector is relatively small, it won't lead
to a sharp change in the recall rate within a certain range.
```

# Test Result

## MSMARCO BM25

| max score ratio | Recall rate of WAND(%) | Recall rate of MaxScore(%) | Query time of WAND(ms) | Query time of MaxScore(ms) |
| --- | --- | --- | --- | --- |
| 1 | 0.996691 | 0.996576 | 3375 | 3456 |
| 0.9 | 0.988983 | 0.992708 | 2599 | 2908 |
| 0.8 | 0.942851 | 0.981490 | 1999 | 2417 |
| 0.7 | 0.797994 | 0.936948 | 1474 | 1912 |

## HotpotQA BM25

| max score ratio | Recall rate of WAND(%) | Recall rate of MaxScore(%) | Query time of WAND(ms) | Query time of MaxScore(ms) |
| --- | --- | --- | --- | --- |
| 1 | 0.991757 | 0.991463 | 11885 | 6763 |
| 0.9 | 0.990802 | 0.991445 | 9729 | 6216 |
| 0.8 | 0.981109 | 0.991243 | 8244 | 5592 |
| 0.7 | 0.931632 | 0.988985 | 6752 | 4937 |

### DIFF
--- a/src/index/sparse/sparse_inverted_index_config.h
+++ b/src/index/sparse/sparse_inverted_index_config.h
@@ -45,20 +45,32 @@ class SparseInvertedIndexConfig : public BaseConfig {
         /**
          * The term frequency part of score of BM25 is:
          * tf * (k1 + 1) / (tf + k1 * (1 - b + b * (doc_len / avgdl)))
-         * as more documents being added to the collection, avgdl can also
-         * change. In WAND index we precompute and cache this score in order to
-         * speed up the search process, but if avgdl changes, we need to
-         * re-compute such score which is expensive. To avoid this, we upscale
-         * the max score by a ratio to compensate for avgdl changes. This will
-         * make the max score larger than the actual max score, it makes the
-         * filtering less aggressive, but guarantees the correctness.
-         * The larger the ratio, the less aggressive the filtering is.
+         * WAND algorithm uses the max score of each dim for pruning, which is
+         * precomputed and cached in our implementation. The cached max score
+         * is actually not equal to the actual max score. Instead, it is a
+         * scaled one based on the wand_bm25_max_score_ratio.
+         * We should use different scale strategy for different reasons.
+         * 1. As more documents being added to the collection, avgdl could
+         *    be changed. Re-computing such score for each segment is
+         *    expensive. To avoid this, we should upscale the actual max score
+         *    by a ratio greater than 1.0 to compensate for avgdl changes.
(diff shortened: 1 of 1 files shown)


src/index/sparse/sparse_inverted_index_config.h
```diff
@@ -43,22 +43,31 @@ class SparseInvertedIndexConfig : public BaseConfig {
...
-         * tf * (k1 + 1) / (tf + k1 * (1 - b + b * (doc_len / avgdl)))
-         * as more documents being added to the collection, avgdl can also
-         * change. In WAND index we precompute and cache this score in order to
-         * speed up the search process, but if avgdl changes, we need to
-         * re-compute such score which is expensive. To avoid this, we upscale
-         * the max score by a ratio to compensate for avgdl changes. This will
-         * make the max score larger than the actual max score, it makes the
-         * filtering less aggressive, but guarantees the correctness.
-         * The larger the ratio, the less aggressive the filtering is.
+         * wand_bm25_max_score_ratio is assigned two functions:
```
zhengbuqian, line 46: The term frequency part of score of BM25 is:... . but if avgdl changes, the max score changes. this can be used to adjust the max scored used to compute the max score.
1. if set to a value greater than 1, ...
2. if set to a value less than 1, ...

Issue #1031, opened 2025-01-15
Eyrie support for vector index


foxspy:
Eyrie is a centralized memory management framework. Through Eyrie, some specific operations can be implemented, such as memory statistics, lazy load, etc. Vector index support Eyrie can be broken down into two steps
1. Support memory-based zero copy
2. Implement lazy load based on Eyrie

1. Zero copy support
    - [x] IVF_FLAT
    - [x] IVF_SQ
    - [x] IVF_PQ
    - [ ] SCANN
    - [x] HNSW(faiss)
    - [x] HNSW(hnswlib)
    - [x] HNSW_SQ
    - [x] HNSW_PQ
    - [x] HNSW_PRQ
    - [ ] SPARSE_WAND
    - [ ] SPARSE_INVERTED_INDEX
    - [ ] GPU_INDEXES
    - [ ] Cardinal
    - [ ] Load Resource Estimation
3. Partial Load 
    - [ ] Cardinal

PR #1032, opened 2025-01-15
[WIP] feature : add zero copy support for vector index

foxspy:
issue: #1031

### DIFF
--- a/src/index/flat/flat.cc
+++ b/src/index/flat/flat.cc
@@ -13,6 +13,7 @@
 #include "faiss/IndexBinaryFlat.h"
 #include "faiss/IndexFlat.h"
 #include "faiss/impl/AuxIndexStructures.h"
+#include "faiss/impl/zerocopy_io.h"
 #include "faiss/index_io.h"
 #include "index/flat/flat_config.h"
 #include "io/memory_io.h"
@@ -313,23 +314,25 @@ class FlatIndexNode : public IndexNode {
     }
 
     Status
-    Deserialize(const BinarySet& binset, std::shared_ptr<Config>) override {
+    Deserialize(BinarySet&& binset, std::shared_ptr<Config>) override {
         std::vector<std::string> names = {"IVF",        // compatible with knowhere-1.x
                                           "BinaryIVF",  // compatible with knowhere-1.x
                                           Type()};
-        auto binary = binset.GetByNames(names);
+        binarySet_ = std::move(binset);
+        auto binary = binarySet_.GetByNames(names);
         if (binary == nullptr) {
             LOG_KNOWHERE_ERROR_ << "Invalid binary set.";
             return Status::invalid_binary_set;
         }
 
-        MemoryIOReader reader(binary->data.get(), binary->size);
+        int io_flags = faiss::IO_FLAG_ZERO_COPY;
+        faiss::ZeroCopyIOReader reader(binary->data.get(), binary->size);
         if constexpr (std::is_same<IndexType, faiss::IndexFlat>::value) {
-            faiss::Index* index = faiss::read_index(&reader);
+            faiss::Index* index = faiss::read_index(&reader, io_flags);
             index_.reset(static_cast<IndexType*>(index));
         }
         if constexpr (std::is_same<IndexType, faiss::IndexBinaryFlat>::value) {
-            faiss::IndexBinary* index = faiss::read_index_binary(&reader);
+            faiss::IndexBinary* index = faiss::read_index_binary(&reader, io_flags);
             index_.reset(static_cast<IndexType*>(index));
         }
         return Status::success;
--- a/thirdparty/faiss/faiss/Index.h
+++ b/thirdparty/faiss/faiss/Index.h
@@ -15,6 +15,7 @@
 #include <sstream>
 #include <string>
 #include <typeinfo>
+#include "faiss/impl/mapped_io.h"
 
 #define FAISS_VERSION_MAJOR 1
 #define FAISS_VERSION_MINOR 8
@@ -80,6 +81,8 @@ struct Index {
     /// type of metric this index uses for search
     MetricType metric_type;
     float metric_arg; ///< argument of the metric type
+
+    std::shared_ptr<MmappedFileMappingOwner> mmap_owner;
 
     explicit Index(idx_t d = 0, MetricType metric = METRIC_L2)
             : d(d),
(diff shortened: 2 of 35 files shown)


thirdparty/faiss/faiss/Index.h
```diff
@@ -81,6 +82,8 @@ struct Index {
     MetricType metric_type;
     float metric_arg; ///< argument of the metric type
 
+    std::shared_ptr<MmappedFileMappingOwner> mmap_owner;
```
alexanderguzhva, line 85: this is a very tricky point. Although, I understand why this was introduced, I'd like to avoid adding it here. Basically, the approach is similar to a std::string_view, which does not have a business to maintain the memory. Similar, here: It is not Index's instance business to maintain the view. 

I'm open for discussions.

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -114,10 +134,28 @@ void read_vector(VectorT& target, IOReader* f) {
...
-            );
-            target = std::move(mmapped);
+            VectorT mmapped_view = VectorT::createView(address, nread);
+            target = std::move(mmapped_view);
+
+            return;
+        }
+    }
+
+    ZeroCopyIOReader* zr = dynamic_cast<ZeroCopyIOReader*>(f);
```
alexanderguzhva, line 144: this is a todo for the future: basically, a reader need to instantiate an object. But this is a later business, because it needs to be properly synchronized with the vanilla faiss first.
So, let's keep the change in the way you proposed it.

alexanderguzhva (commented):
one topic to discuss. Otherwise, lgtm

thirdparty/faiss/faiss/impl/index_read.cpp
foxspy: Yes, so I put the PR up first; zero copy is needed here mainly for
1. Optimizing the memory enlargement caused by copying during loading
2. The memory will be managed uniformly through Eyrie later, and the index does not hold data, but only views
3. The implementation of mmap and zero copy can be unified (mmap externally first, load through zero copy)
If you have any ideas about this, please feel free to communicate; has faiss considered supporting this mode?

src/index/flat/flat.cc
```diff
@@ -313,23 +314,25 @@ class FlatIndexNode : public IndexNode {
...
+        binarySet_ = std::move(binset);
+        auto binary = binarySet_.GetByNames(names);
         if (binary == nullptr) {
             LOG_KNOWHERE_ERROR_ << "Invalid binary set.";
             return Status::invalid_binary_set;
         }
 
-        MemoryIOReader reader(binary->data.get(), binary->size);
+        int io_flags = faiss::IO_FLAG_ZERO_COPY;
+        faiss::ZeroCopyIOReader reader(binary->data.get(), binary->size);
```
alexanderguzhva, line 329: Such an approach won't lead to a memory leak, correct?

thirdparty/faiss/faiss/Index.h
alexanderguzhva: @foxspy any comments on this? Basically, I doubt that we'll able to introduce this variable in vanilla Faiss when porting back

thirdparty/faiss/faiss/impl/zerocopy_io.cpp
```diff
@@ -0,0 +1,61 @@
+#include <faiss/impl/zerocopy_io.h>
+#include <cstring>
+
+namespace faiss {
+
+ZeroCopyIOReader::ZeroCopyIOReader(uint8_t* data, size_t size) : data_(data), rp_(0), total_(size) {
```
alexanderguzhva, line 6: A duplicate code, similar to `src/io/memory_io.h`

src/index/flat/flat.cc
foxspy: The binary will be held by the index; so there will be no memory leak

thirdparty/faiss/faiss/impl/zerocopy_io.cpp
foxspy: This is because Knowhere is a hub that is adapted  to multiple indexes. For indexes other than faiss, it uses the memory_io.h. Faiss needs to remain independent zerocopy_io.cpp and not rely on related objects of knowhere.

alexanderguzhva:
> Zero copy may cause data misalignment, because data serialization in Faiss does not require alignment, which may cause performance degradation or even crashes in some scenarios. @alexanderguzhva Do you have any suggestions for this?

I'm not sure I follow. Any examples?
As far as was able to understand your code, it is about avoiding memory duplication, so I'm not sure where a data misalignment and crash might occur

alexanderguzhva:
I mean that this PR totally makes sense, with the exception of adding a memory owner directly to the `faiss::Index` object

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -88,7 +89,37 @@ bool read_is_mv(const char* fname) {
...
+                size_t(size),                        \
+                strerror(errno));                    \
+
+            VectorT view = VectorT::createView(address, nread);
+            target = std::move(view);
+
+            return;
+        }
+    }
+    target.resize(size);
```
alexanderguzhva, line 118: this particular line 118 is not needed

foxspy:
> > Zero copy may cause data misalignment, because data serialization in Faiss does not require alignment, which may cause performance degradation or even crashes in some scenarios. @alexanderguzhva Do you have any suggestions for this?
> 
> I'm not sure I follow. Any examples? As far as was able to understand your code, it is about avoiding memory duplication, so I'm not sure where a data misalignment and crash might occur

A simple example is, if https://github.com/zilliztech/knowhere/pull/1032/files#diff-d05f16d9b1473d49c536582f9433cca57d6a2909a5a729dec54c383f598b57c4R119 assign 0x5ffffff3 to float *, it will cause misalignment of memory access. Although I have tried it on x64 and arm machines, it will not cause a crash, but I know that this will generally lead to performance degradation (may cause multiple memory accesses); some older machines may also have strange problems; for example, scann also has SIMD requirements for alignment https://github.com/zilliztech/knowhere/blob/main/thirdparty/faiss/faiss/utils/AlignedTable.h#L115, which will cause memory expansion (duplicated binary + memory copy) if not handled

PR #1033, opened 2025-01-15
sparse: disable refinement by default

sparknack:
Because of the removal of forward index, the current refine searching is
slower than before. To prevent performance degradation when using a
non-zero value of drop_ratio_search, disable refinement by default.

### DIFF
--- a/src/index/sparse/sparse_inverted_index_config.h
+++ b/src/index/sparse/sparse_inverted_index_config.h
@@ -38,11 +38,19 @@ class SparseInvertedIndexConfig : public BaseConfig {
             .for_search()
             .for_range_search()
             .for_iterator();
+        /**
+         * refine_factor is used for approximate search.
+         * refine_factor == 1 means no refinement, and is the default value.
+         * refine_factor > 1 means refinement. The larger the value, the more
+         * accurate the approximate result will be, but the slower the
+         * performance.
+         * Be aware that if you opt to use a large drop_ratio_search, it is
+         * necessary for you to manually modify this value.
+         */
         KNOWHERE_CONFIG_DECLARE_FIELD(refine_factor)
-            .description("refine factor")
-            .set_default(10)
-            .for_search()
-            .for_range_search();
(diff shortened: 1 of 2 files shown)


sparknack:
> lgtm, but should it affect existing users?

Yes. For the current code in main branch, if refinement is used, it will lead to a significant degradation in performance, thus affecting the usability of all range of the drop_ratio_search. Although turning off refinement by default will cause a certain decline in the recall rate with the same drop_ratio_search, it can still ensure that it is basically usable. Users still can use small drop_ratio_search like 0.1.

src/index/sparse/sparse_inverted_index_config.h
```diff
@@ -38,9 +38,19 @@ class SparseInvertedIndexConfig : public BaseConfig {
...
+         * Be aware that if you opt to use a large drop_ratio_search, it is
+         * necessary for you to manually modify this value.
+         */
         KNOWHERE_CONFIG_DECLARE_FIELD(refine_factor)
             .description("refine factor")
-            .set_default(10)
+            .set_default(1)
+            .set_range(1, 10, true, true)
             .for_search()
             .for_range_search();
```
zhengbuqian, line 55: remove for_range_search

src/index/sparse/sparse_inverted_index_config.h
```diff
@@ -38,9 +38,19 @@ class SparseInvertedIndexConfig : public BaseConfig {
...
+         * accurate the approximate result will be, but the slower the
+         * performance.
+         * Be aware that if you opt to use a large drop_ratio_search, it is
+         * necessary for you to manually modify this value.
+         */
         KNOWHERE_CONFIG_DECLARE_FIELD(refine_factor)
             .description("refine factor")
-            .set_default(10)
+            .set_default(1)
+            .set_range(1, 10, true, true)
```
zhengbuqian, line 53: no need to set a upper limit

sparknack:
issue: #1035

Issue #1019, opened 2025-01-09
MV only support for HNSW index


chasingegg:
milvus issue: https://github.com/milvus-io/milvus/issues/34332

When milvus partition key isolation is enabled, one hnsw index will become a "fusion index" where each part stands for one tenant and could possibly reduce search latency.

PR #1036, opened 2025-01-16
Add IsAdditionalScalarSupported check for mv_only

chasingegg:
issue: #1019 
`IsAdditionalScalarSupported` was used to indicate whether vector indexes support building with scalar info, but we need a special path for mv_only

/kind improvement


chasingegg:
/hold
wait for 2.5.4 release

Issue #1039, opened 2025-01-16
Reduce unittest runtime


cydrain:
The current runtime for unittest is around 2 hours, and this time needs to be reduced to less than 1 hour.
Check following file for detailed case runtime:
[ut_runtime.txt](https://github.com/user-attachments/files/18436672/ut_runtime.txt)

sparknack:
ack

PR #1041, opened 2025-01-17
sparse: reduce ut execution time

sparknack:
Reduce ut execution time, which is from 532.62 secs to 46.50 secs on my machine.


sparknack:
issue: #1039

alexanderguzhva:
@sparknack is it needed to run 2000 vectors if we have 3000?

sparknack:
> @sparknack is it needed to run 2000 vectors if we have 3000?

3000 is the vector dimension, not the total number of the vectors.

foxspy (approved):
/lgtm

PR #1050, opened 2025-02-05
sparse: early return to avoid mmapping empty file

sparknack:
The size of the temporary file for mmapping could be 0, which is not allowed by mmap syscall. So add early return to avoid it.

Also remove some typos.


foxspy (approved):
/lgtm

Issue #909, opened 2024-10-23
Optimize Raw Data Search: BF functions and Intermin Index


cqy123456:
Purpose: 
Knowhere provides Brute Force functions or a JIT intermin index for milvus streaming raw data search.
In order to minimize the additional memory overhead of Milvus raw data search and improve search performance, knowhere will make the following optimizations:

**1. bruteforce search with zero copy:**
- [x] More Simd support for fp16/bf16(batch_4); -> pr: https://github.com/zilliztech/knowhere/pull/980
- [x] remove all (fp16/bf16->fp32) conversion, and use fp16/bf16 directly to do brute force; -> pr: https://github.com/zilliztech/knowhere/pull/980

**2. choose SCANN base index as intermin index: **
- [x] scann iterator; -> pr: https://github.com/zilliztech/knowhere/pull/992
- [x] thread safe protection;-> pr: https://github.com/zilliztech/knowhere/pull/992
- [x] sampling training or faster train;->pr:https://github.com/zilliztech/knowhere/pull/1052

**3. a no memory(NM) version of  interim index:**
- [x]  a NM flat as storage;
- [x] multi data type support;
- [x] distances function with get_vector lamdba;
- [x] more refine type support (data view, fp16, bf16, uint8)

**4. milvus related issue:**

future works:
1. More data type support: binary/ sparse;
2.  ivf flat cc with nm;
3. lock-free verison of thread safe scann;
4. use simd to convert fp16/bf16 to fp32;
5. scann only support dim%2 ==0, we need to solve this problem.

liliu-z:
@cqy123456 are we still tracking this here?

PR #1052, opened 2025-02-06
Reduce the pq build time

cqy123456:
issue: https://github.com/zilliztech/knowhere/issues/909
For better support scann_dvr as a growing index, this pr optimize the build time of pq.
Use corhere 256M to test: nlist =128, dsub = 4, one build thread:
build time :
before: 5.12s
after: 1.9s

### DIFF
--- a/src/simd/distances_avx.cc
+++ b/src/simd/distances_avx.cc
@@ -754,5 +754,174 @@ bf16_vec_norm_L2sqr_avx(const knowhere::bf16* x, size_t d) {
     auto res = _mm256_reduce_add_ps(msum_0);
     return res;
 }
+
+namespace {
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+inline void
+fvec_L2sqr_ny_avx_impl(float* dis, const float* x, const float* y, size_t d, size_t ny) {
+    size_t i = 0;
+    for (; i + 3 < ny; i += 4) {
+        const float* __restrict y1 = y + d * i;
+        const float* __restrict y2 = y + d * (i + 1);
+        const float* __restrict y3 = y + d * (i + 2);
+        const float* __restrict y4 = y + d * (i + 3);
+        fvec_L2sqr_batch_4_avx(x, y1, y2, y3, y4, d, dis[i], dis[i + 1], dis[i + 2], dis[i + 3]);
+    }
+    while (i < ny) {
+        const float* __restrict y_i = y + d * i;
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@cqy123456 just in case, there's a corresponding code in https://github.com/facebookresearch/faiss/blob/main/faiss/utils/distances_simd.cpp and https://github.com/facebookresearch/faiss/tree/main/faiss/utils/transpose

cqy123456:
fvec_op_ny_D4 to test:
avx2 version : 2.5s
avx512 version: 3.2s

thirdparty/faiss/faiss/utils/distances.cpp
```diff
@@ -967,6 +967,36 @@ void all_inner_product(
...
+void exhaustive_L2sqr_nearest_imp(
+        const float* __restrict x,
+        const float* __restrict y,
+        size_t d,
+        size_t nx,
+        size_t ny,
+        float* vals,
+        int64_t* ids) {
+    const size_t ny_batch_size = 256;
+    auto sub_dis = std::make_unique<float[]>(std::min(ny_batch_size, ny));
```
alexanderguzhva, line 979: this memory allocation can be avoided by having a static array allocation
```C++
constexpr size_t ny_batch_size = 256;
float sub_dis[ny_batch_size];
```

src/simd/distances_avx.cc
```diff
@@ -754,5 +754,174 @@ bf16_vec_norm_L2sqr_avx(const knowhere::bf16* x, size_t d) {
...
+    }
+}
+
+size_t
+fvec_L2sqr_ny_nearest_avx(float* __restrict distances_tmp_buffer, const float* __restrict x, const float* __restrict y,
+                          size_t d, size_t ny) {
+    fvec_L2sqr_ny_avx(distances_tmp_buffer, x, y, d, ny);
+
+    size_t nearest_idx = 0;
+    float min_dis = HUGE_VALF;
```
alexanderguzhva, line 915: I'd use `std::numeric_limits<float>::max()` instead of `HUGE_VALF`

alexanderguzhva:
lgtm overall, just a couple of nits

src/simd/distances_avx.cc
cqy123456: updated

thirdparty/faiss/faiss/utils/distances.cpp
cqy123456: updated

foxspy (approved):
/lgtm

PR #1069, opened 2025-02-12
Optimize e2e script

cydrain:


cydrain:
Decide not to merge this PR.
Still use "cp" instead of "ln -s" in e2e script to avoid potential issue caused by disk.

/close

PR #1076, opened 2025-02-13
fix a compilation bug in `test_data_view_index.cc`

alexanderguzhva:
otherwise, the code won't compile using `clang`
/kind bug


foxspy (approved):
/lgtm

Issue #1085, opened 2025-02-19
sparse enhancement


sparknack:
- [ ] refactor the sparse implementation

PR #1086, opened 2025-02-19
sparse: new version

sparknack:
issue: #1085 

1. Refactor the inverted index structures leveraging the PISA framework.
2. Categorize the inverted index into two types according to the index containers: growable and flattened.
3. Directly perform quantization of float to fp16.
4. Incorporate a search_algo parameter to enable dynamic adjustment of the retrieval algorithm.


sparknack:
issue: #1085

src/index/sparse/inverted/flatten_inverted_index.h
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    // reset reader to the saved beginning
+    reader.seekg(saved_reader_loc);
+
+    // calculate raw index byte size
+    auto raw_index_ids_byte_sz = nnz * sizeof(uint32_t);
+    auto raw_index_vals_byte_sz = nnz * sizeof(QType);
+    auto raw_index_offsets_byte_sz = (this->dim_map_.size() + 1) * sizeof(size_t);
```
alexanderguzhva, line 247: `size_t` is platform-dependent type, which makes it not very suitable for the serialization. And it will be `uint64_t` in 99.99% cases. So, I'd rather change the type to `uint64_t`
But the change is not super-required, tbh

alexanderguzhva (commented):
mostly nits

CMakeLists.txt
```diff
@@ -97,6 +97,9 @@ find_package(OpenMP REQUIRED)
 find_package(folly REQUIRED)
 include_directories(${folly_INCLUDE_DIRS})
 
+find_package(Microsoft.GSL REQUIRED)
```
zhengbuqian, line 100: can we use boost span instead? https://live.boost.org/doc/libs/1_85_0/libs/core/doc/html/core/span.html

PR #1091, opened 2025-02-21
Run e2e with 2 threads to reduce runtime

cydrain:


cydrain:
https://jenkins.milvus.io:18080/blue/organizations/jenkins/knowhere%2Fkn2/detail/PR-1091/2/pipeline

It seems that running pytest in parallel causes the system to crash consistently.

Screenshot from 2025-02-21 17-54-18

PR #1098, opened 2025-02-26
Add Int8Vector SIMD APIs for avx512/avx2/sse/ref

cydrain:
Issue: #977

### DIFF
--- a/src/simd/distances_avx.cc
+++ b/src/simd/distances_avx.cc
@@ -877,6 +877,95 @@ bf16_vec_L2sqr_batch_4_avx(const knowhere::bf16* x, const knowhere::bf16* y0, co
 }
 
 ///////////////////////////////////////////////////////////////////////////////
+// int8
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+float
+int8_vec_inner_product_avx(const int8_t* x, const int8_t* y, size_t d) {
+    int32_t res = 0;
+    FAISS_PRAGMA_IMPRECISE_LOOP
+    for (size_t i = 0; i < d; i++) {
+        res += (int32_t)x[i] * (int32_t)y[i];
+    }
+    return (float)res;
+}
+FAISS_PRAGMA_IMPRECISE_FUNCTION_END
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+float
+int8_vec_L2sqr_avx(const int8_t* x, const int8_t* y, size_t d) {
+    int32_t res = 0;
+    FAISS_PRAGMA_IMPRECISE_LOOP
+    for (size_t i = 0; i < d; i++) {
+        const int32_t tmp = (int32_t)x[i] - (int32_t)y[i];
+        res += tmp * tmp;
+    }
+    return (float)res;
+}
+FAISS_PRAGMA_IMPRECISE_FUNCTION_END
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+float
+int8_vec_norm_L2sqr_avx(const int8_t* x, size_t d) {
+    int32_t res = 0;
+    FAISS_PRAGMA_IMPRECISE_LOOP
+    for (size_t i = 0; i < d; i++) {
+        res += (int32_t)x[i] * (int32_t)x[i];
+    }
+    return (float)res;
+}
+FAISS_PRAGMA_IMPRECISE_FUNCTION_END
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+void
+int8_vec_inner_product_batch_4_avx(const int8_t* x, const int8_t* y0, const int8_t* y1, const int8_t* y2,
+                                   const int8_t* y3, const size_t d, float& dis0, float& dis1, float& dis2,
+                                   float& dis3) {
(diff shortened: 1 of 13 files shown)


src/simd/distances_avx.cc
```diff
@@ -876,6 +876,95 @@ bf16_vec_L2sqr_batch_4_avx(const knowhere::bf16* x, const knowhere::bf16* y0, co
...
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+float
+int8_vec_inner_product_avx(const int8_t* x, const int8_t* y, size_t d) {
+    float res = 0;
+    FAISS_PRAGMA_IMPRECISE_LOOP
+    for (size_t i = 0; i < d; i++) {
+        res += (float)x[i] * (float)y[i];
+    }
+    return res;
```
alexanderguzhva, line 890: Please use the following pattern here and it all the further functions that were added: compute in ints and convert the final result into float. This is faster in terms of CPU ops / cycles.
```C++
FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
float
int8_vec_inner_product_avx(const int8_t* x, const int8_t* y, size_t d) {
    int32_t res = 0;
    FAISS_PRAGMA_IMPRECISE_LOOP
        res += (int32_t)x[i] * (int32_t)y[i];
    }
    return (float)res;
}
```

src/simd/distances_avx.cc
```diff
@@ -876,6 +876,95 @@ bf16_vec_L2sqr_batch_4_avx(const knowhere::bf16* x, const knowhere::bf16* y0, co
...
+        res += tmp * tmp;
+    }
+    return res;
+}
+FAISS_PRAGMA_IMPRECISE_FUNCTION_END
+
+FAISS_PRAGMA_IMPRECISE_FUNCTION_BEGIN
+float
+int8_vec_norm_L2sqr_avx(const int8_t* x, size_t d) {
+    double res = 0;
```
alexanderguzhva, line 910: same here, `int32_t res = 0;`

src/simd/distances_neon.cc
```diff
@@ -2113,6 +2115,37 @@ bf16_vec_L2sqr_batch_4_neon(const knowhere::bf16* x, const knowhere::bf16* y0, c
     dis3 = vaddvq_f32(res.val[3]);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// int8
+
+float
+int8_vec_inner_product_neon(const int8_t* x, const int8_t* y, size_t d) {
+    throw std::runtime_error("not implemented");
```
alexanderguzhva, line 2123: why, please copy-paste a standard code from the `ref` implementation while this one is not implemented properly.

alexanderguzhva (changes requested):

src/simd/distances_avx.cc
cydrain: ok, thx

src/simd/distances_neon.cc
cydrain: Hi Alex, I plan to implement a real neon function later

src/simd/distances_neon.cc
```diff
@@ -2113,6 +2113,85 @@ bf16_vec_L2sqr_batch_4_neon(const knowhere::bf16* x, const knowhere::bf16* y0, c
...
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// int8
+
+float
+int8_vec_inner_product_neon(const int8_t* x, const int8_t* y, size_t d) {
+    // TODO caiyd: use ref implementation temporarily
+    int32_t res = 0;
+    for (size_t i = 0; i < d; i++) {
```
foxspy, line 2123: Duplicated code here
{ 
   return int8_vec_inner_product_ref(x, y, d);
}

src/simd/distances_neon.cc
```diff
@@ -2113,6 +2113,85 @@ bf16_vec_L2sqr_batch_4_neon(const knowhere::bf16* x, const knowhere::bf16* y0, c
...
+        res += (int32_t)x[i] * (int32_t)y[i];
+    }
+    return (float)res;
+}
+
+float
+int8_vec_L2sqr_neon(const int8_t* x, const int8_t* y, size_t d) {
+    // TODO caiyd: use ref implementation temporarily
+    int32_t res = 0;
```
foxspy, line 2132: Same as above

src/simd/distances_neon.cc
```diff
@@ -2113,6 +2113,85 @@ bf16_vec_L2sqr_batch_4_neon(const knowhere::bf16* x, const knowhere::bf16* y0, c
...
+}
+
+float
+int8_vec_norm_L2sqr_neon(const int8_t* x, size_t d) {
+    // TODO caiyd: use ref implementation temporarily
+    int32_t res = 0;
+    for (size_t i = 0; i < d; i++) {
+        res += (int32_t)x[i] * (int32_t)x[i];
+    }
+    return (float)res;
```
Presburger, line 2147: implement a neon version.

alexanderguzhva (approved):

src/simd/distances_neon.cc
alexanderguzhva: no, the duplicated code is correct, bcz it will be compiled with a different compiler options

PR #1108, opened 2025-03-05
Fix CAGRA-HNSW serialization format

lowener:
I faced today a bug with using CAGRA-HNSW, that was introduced in the switch from Raft to cuVS.
The `serialize_to_hnsw` serialization function used in knowhere was different from the cuVS one.
I added a test to make sure that bugs in this function will be detected in the future.
issue: #1113

### DIFF
--- a/src/common/cuvs/proto/cuvs_index.cuh
+++ b/src/common/cuvs/proto/cuvs_index.cuh
@@ -268,6 +268,20 @@ struct cuvs_index {
                                      bool include_dataset = true) {
         auto const& underlying_index = index.get_vector_index();
         if constexpr (vector_index_kind == cuvs_index_kind::cagra) {
+            size_t metric_type;
+            if (underlying_index.metric() == cuvs::distance::DistanceType::L2Expanded) {
+                metric_type = 0;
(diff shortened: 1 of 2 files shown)


alexanderguzhva:
what about the backward compatibility? I see that the code affects the serialization only

Presburger:
@lowener Hello, since Knowhere uses a customized version of HNSWLIB, I previously wrote a serialization function different from RAFT's to be compatible with our HNSWLIB version. Does the current implementation ensure complete alignment in field parsing? The previous implementation is in this PR: https://github.com/zilliztech/knowhere/pull/661/files

lowener:
@Presburger Yes, you can check cuvs implementation of the serialization here: https://github.com/rapidsai/cuvs/blob/branch-25.04/cpp/src/neighbors/detail/cagra/cagra_serialize.cuh#L105.
The only difference that I had not noticed before is that the first 3 fields (metric type, data size and dim) were assumed to be known on cuvs side. However that's not the case in knowhere. So I am adding them back in this PR. The rest of the serialization is the same, the number of fields and their content match.

Issue #1122, opened 2025-03-25
TODO: Isolate growing index thread pool


foxspy:
Isolate the growing index thread pool and increase the thread priority higher than the normal build pool,  ensure that the growing index write will not be blocked by other index building tasks

PR #1126, opened 2025-03-27
enhance: reduce the queuing problem caused by the mixed task of growing index and sealed index

foxspy:
issue: #1122 
milvus issue: https://github.com/milvus-io/milvus/issues/40825

### DIFF
--- a/include/knowhere/comp/thread_pool.h
+++ b/include/knowhere/comp/thread_pool.h
@@ -330,4 +330,33 @@ WaitAllSuccess(std::vector<folly::Future<T>>& futures) {
     return Status::success;
 }
 
+// This class is used to wrap the thread pool and the inline executor
+// If use_pool is true, the function will be pushed to the thread pool
+// If use_pool is false, the function will be executed directly
+class ThreadPoolWrapper {
+ public:
+    ThreadPoolWrapper(const std::shared_ptr<ThreadPool>& pool, bool use_pool = true)
+        : pool_(pool), use_pool_(use_pool) {
+    }
+
+    template <typename Func, typename... Args>
+    auto
+    push(Func&& func, Args&&... args) {
+        if (use_pool_) {
+            return pool_->push(std::forward<Func>(func), std::forward<Args>(args)...);
+        } else {
+            // If the pool is not used, the function will be executed within the current thread directly
+            return folly::makeSemiFuture()
+                .via(&folly::InlineExecutor::instance())
+                .then([func = std::forward<Func>(func), &args...](auto&&) mutable {
+                    return func(std::forward<Args>(args)...);
+                });
+        }
+    }
+
+ private:
+    std::shared_ptr<ThreadPool> pool_;
+    bool use_pool_;
+};
+
 }  // namespace knowhere
--- a/include/knowhere/index/index.h
+++ b/include/knowhere/index/index.h
@@ -140,7 +140,7 @@ class Index {
     }
 
     Status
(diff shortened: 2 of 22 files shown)


include/knowhere/comp/thread_pool.h
```diff
@@ -330,4 +330,37 @@ WaitAllSuccess(std::vector<folly::Future<T>>& futures) {
     return Status::success;
 }
 
+// This class is used to wrap the thread pool and the inline executor
+// If use_pool is true, the function will be pushed to the thread pool
+// If use_pool is false, the function will be executed directly
+class ThreadPoolWrapper {
+ public:
+    ThreadPoolWrapper(std::shared_ptr<ThreadPool> pool, bool use_pool = true) : pool_(pool), use_pool_(use_pool) {
```
alexanderguzhva, line 338: `const std::shared_ptr<ThreadPool>&`

include/knowhere/comp/thread_pool.h
```diff
@@ -330,4 +330,37 @@ WaitAllSuccess(std::vector<folly::Future<T>>& futures) {
...
+            return folly::makeSemiFuture()
+                .via(&folly::InlineExecutor::instance())
+                .then([func = std::forward<Func>(func), &args...](auto&&) mutable {
+                    return func(std::forward<Args>(args)...);
+                });
+        }
+    }
+
+    static std::shared_ptr<ThreadPoolWrapper>
+    GetThreadPoolWrapper(std::shared_ptr<ThreadPool> pool, bool use_pool) {
```
alexanderguzhva, line 357: `const std::shared_ptr<ThreadPool>&`
Also, I don't understand why this static constructor is needed at all, tbh

include/knowhere/comp/thread_pool.h
```diff
@@ -330,4 +330,37 @@ WaitAllSuccess(std::vector<folly::Future<T>>& futures) {
     return Status::success;
 }
 
+// This class is used to wrap the thread pool and the inline executor
+// If use_pool is true, the function will be pushed to the thread pool
+// If use_pool is false, the function will be executed directly
+class ThreadPoolWrapper {
```
alexanderguzhva, line 336: I don't understand why this is organized in a form of a class. Technically, the scenarios in the code imply a definite static method, because every instantiated `ThreadPoolWrapper` gets created and makes just 1 call
```C++
auto wrapper = std::make_shared<ThreadPoolWrapper>(pool, use_pool);
auto task = wrapper->push(...);
task.wait();
```

Wouldn't it be better to have a static method?
```C++
auto task = whatever_name(pool, use_pool, Func, Args...);
task.wait();
```

foxspy: > I don't understand why this is organized in a form of a class. Technically, the scenarios in the code imply a definite static method, because every instantiated `ThreadPoolWrapper` gets created and makes just 1 call
> 
> ```c++
> auto wrapper = std::make_shared<ThreadPoolWrapper>(pool, use_pool);
> auto task = wrapper->push(...);
> task.wait();
> ```
> 
> Wouldn't it be better to have a static method?
> 
> ```c++
> auto task = whatever_name(pool, use_pool, Func, Args...);
> task.wait();
> ```

The main purpose here is to maintain the existing implementation style. Just simply add a wrapper to ensure that the original ThreadPool does not need to be modified. At the same time, the original place where ThreadPool is called only needs to simply replace ThreadPool with ThreadPoolWrapper.

PR #1148, opened 2025-04-14
Add fallback bf for hnsw when the result is not enough

chasingegg:
/kind improvement

### DIFF
--- a/src/index/hnsw/faiss_hnsw.cc
+++ b/src/index/hnsw/faiss_hnsw.cc
@@ -1189,6 +1189,18 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
             return expected<DataSetPtr>::Err(Status::invalid_args, "an input index seems to be unrelated to HNSW");
         }
 
+        // set up a bf wrapper as fallback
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
lgtm overall, let's wait for the ut

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -1247,6 +1260,22 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
                     faiss::idx_t* const __restrict local_ids = ids.get() + k * idx;
                     float* const __restrict local_distances = distances.get() + k * idx;
 
+                    // check if we need to perform a brute-force search bcz of the lack of results
+                    auto bf_search_needed = [&]() -> bool {
+                        size_t real_topk = 0;
+                        for (auto j = 0; j < k; ++j) {
+                            if (local_ids[j] < 0) {
+                                continue;
```
foxspy, line 1268: break?

alexanderguzhva: no, the code is similar to `std::count_if()` of negative elements.

foxspy (approved):
/lgtm

PR #1160, opened 2025-04-17
knowhere support riscv

lyd1992:
image
image

This update introduces RISC-V 64 architecture recognition in the build system by adding dedicated macro detection (__riscv) to the CMake configuration file, while improving error diagnostics to clearly indicate riscv64 as a supported platform. These modifications ensure proper architecture validation and prevent accidental builds on unsupported platforms, laying foundational support for future RISC-V-specific optimizations like SIMD implementations.


alexanderguzhva:
issue: #1161

PR #1179, opened 2025-05-06
Fix scann filter search performance and cleanup ensure_topk_full config

chasingegg:
/kind improvement

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -805,7 +805,7 @@ IvfIndexNode<DataType, IndexType>::Search(const DataSetPtr dataset, std::unique_
                     faiss::IVFSearchParameters base_search_params;
                     base_search_params.sel = id_selector;
                     base_search_params.nprobe = nprobe;
-                    base_search_params.ensure_topk_full = ivf_cfg.ensure_topk_full.value();
+                    base_search_params.ensure_topk_full = scann_cfg.ensure_topk_full.value();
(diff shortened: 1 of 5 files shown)


src/index/ivf/ivf.cc
```diff
@@ -781,16 +781,10 @@ IvfIndexNode<DataType, IndexType>::Search(const DataSetPtr dataset, std::unique_
...
-                    if (ivf_search_params.ensure_topk_full) {
-                        ivf_search_params.nprobe = index_->nlist;
-                        // use max_codes to early termination
-                        ivf_search_params.max_codes =
-                            (nprobe * 1.0 / index_->nlist) * (index_->ntotal - bitset.count());
-                    } else {
-                        ivf_search_params.nprobe = nprobe;
-                        ivf_search_params.max_codes = 0;
-                    }
+                    ivf_search_params.ensure_topk_full = true;
```
chasingegg, line 784: CC index always set ensure_topk_full as true

src/index/ivf/ivf_config.h
```diff
@@ -40,7 +40,7 @@ class IvfConfig : public BaseConfig {
             .description("whether to use elkan algorithm")
             .for_train();
         KNOWHERE_CONFIG_DECLARE_FIELD(ensure_topk_full)
-            .set_default(true)
+            .set_default(false)
```
chasingegg, line 43: make the default value as false, and let wrapper node like DataViewIndex to set the config

PR #1183, opened 2025-05-09
Add support for FP16/INT8 on GPU indexes

lowener:
This is a WIP.
TODO: Add support for CAGRA+HNSW int8


cmake/libs/libcuvs.cmake
```diff
@@ -16,7 +16,7 @@
 add_definitions(-DKNOWHERE_WITH_CUVS)
 set(CUVS_VERSION "${RAPIDS_VERSION}")
 set(CUVS_FORK "rapidsai")
-set(CUVS_PINNED_TAG "branch-25.02")
+set(CUVS_PINNED_TAG "branch-25.06")
```
alexanderguzhva, line 19: has cuvs 25.06 been released already? If not, then it is safe to use it already?

src/common/cuvs/proto/cuvs_index.hpp
```diff
@@ -164,7 +164,18 @@ struct cuvs_index {
 
         auto k = neighbors.extent(1);
         auto k_tmp = k + k_offset;
-        if (refine_ratio > 1.0f) {
+        bool do_refine_step = refine_ratio > 1.0f;
```
alexanderguzhva, line 167: is the refine index type is the same as the coarse index type? If not, then 1.0f `refine_ratio` value is acceptable: imagine coarse index being `bf16` and the fine index being `fp32`

src/common/cuvs/integration/cuvs_knowhere_index.cuh
```diff
@@ -532,51 +525,33 @@ struct cuvs_knowhere_index<IndexKind, DataType>::impl {
...
                 thrust::make_tuple(thrust::device_ptr<knowhere_indexing_type>(device_knowhere_ids.data_handle()),
                                    thrust::device_ptr<knowhere_distance_type>(device_distances.data_handle()))),
-            detail::check_valid_entry<thrust::tuple<knowhere_indexing_type, knowhere_distance_type>,
-                                      knowhere_distance_type, knowhere_indexing_type>{max_distance,
-                                                                                      knowhere_indexing_type(size())},
-            knowhere_indexing_type{-1});
-
-        if constexpr (index_kind == cuvs_proto::cuvs_index_kind::brute_force) {
-            if (k_tmp > k) {
-                for (auto i = 0; i < host_ids.extent(0); ++i) {
```
alexanderguzhva, line 562: why is this logic no longer needed?

cmake/libs/libcuvs.cmake
lowener: The official release for 25.06 is the 5th of June. So it will be safe to use from that day on.

src/common/cuvs/integration/cuvs_knowhere_index.cuh
lowener: This logic was introduced for post-filtering the brute-force algorithm. It now supports pre-filtering so that's why it is not required anymore.
I found also some comments left at the time it was introduced detailing it: https://github.com/zilliztech/knowhere/pull/286#discussion_r1424297593

src/common/cuvs/proto/cuvs_index.hpp
lowener: The refinement step will narrow the n_candidates neighbors selected by the ANN search  down to `k` neighbors. If the `refine_ratio` is equal to 1.0 then there is no need to narrow it down because `n_candidates == k`

alexanderguzhva:
issue #1201

alexanderguzhva:
issue: #1201

PR #1199, opened 2025-05-21
remove override range_search from hnsw / ivf, use iterator-based instead

alwayslove2013:


alexanderguzhva:
@alwayslove2013 Do I get it correct that we're deprecating the range search completely?

alwayslove2013:
@alexanderguzhva In knowhere, there are two approaches to `range_search`. One is an **internally implemented** range_search within a specific index (override, like `ivf `/ `faiss_hnsw`), while the other is a unified method from the parent class (`index_node`) that utilizes an **iterator**. This iterator continuously calls next() to collect the best results that meet the specified range.
https://github.com/zilliztech/knowhere/blob/8a705a0a39fecb8fb94f8edee9ab964cb0c22566/include/knowhere/index/index_node.h#L155-L171

It's a historical misalignment between milvus and knowhere. Previously, it was defined that knowhere would return **all results that fit within the range**. However, Milvus does not require all results and operates under a k limit (which we can name `range_search_k`). If the specified range is too large while range_search_k is relatively small, it can lead to significant inefficiencies. The iterator-based range_search can appropriately halt when sufficient results are collected.

chasingegg (approved):
/lgtm

PR #1200, opened 2025-05-21
remove override range_search from ivf, use iterator-based instead

alwayslove2013:
/kind improvement


alwayslove2013:
> is it good to review these diffs? It seems that they fail UTs

alwayslove2013:
@alexanderguzhva, it appears that there are some additional questions; I will address them first.

PR #1205, opened 2025-05-26
Fix hnsw with mv bug

chasingegg:
milvus issue: https://github.com/milvus-io/milvus/issues/42060

No problem in hnswpq/hnswprq, when the scalar info is less or equal to 1 category, fall back to one big index instead of isolated indexes

/kind improvement


foxspy (approved):
/lgtm

Issue #782, opened 2024-08-26
Support SVE


liliu-z:

alexanderguzhva:
https://github.com/zilliztech/knowhere/issues/693

Issue #1193, opened 2025-05-13
Benchmarks for fp16, bf16, and int8 data types fall back to fp32 during index creation


ThatikondV:
**Description:**

While running benchmarks on Knowhere using fp16, bf16, or int8 data types, the index creation is unexpectedly falling back to fp32. This causes the benchmarking results to not reflect the intended lower-precision formats.

- Attempted to create a FLAT index using FLAT_bf16.
- Log shows that the correct key FLAT_bf16 is used initially:

      use key FLAT_bf16 to create knowhere index FLAT with version 6

- But immediately afterward, the index is replaced with FLAT_fp32:

      replace index FLAT_bf16 with FLAT_fp32

It is happening for all algorithms and datatypes like fp16 and bf16.

When I am going through the codebase, I could see that these datatypes are getting mocked in operands.h file.

This prevents accurate benchmarking of fp16 and bf16 index performance and behavior. For example, the reported build time and performance metrics will reflect fp32 characteristics instead of the target precision.

Kindly look into this issue and suggest a way to correctly evaluate benchmarks for fp16, bf16, and int8 data types without fallback to fp32.

Please find the attached images that demonstrate the issue for better clarity.

![Image](https://github.com/user-attachments/assets/0f8d6edd-27c2-4793-b54d-cfffaea1ee85)
![Image](https://github.com/user-attachments/assets/f2413dd6-c3c7-44fe-951c-0eb3d2d9db79)

alexanderguzhva:
@ThatikondV let me see

ThatikondV:
Hi @alexanderguzhva, Is there any update on this ?

alexanderguzhva:
@ThatikondV it seems that internally we're using BruteForce facilities for this, https://github.com/zilliztech/knowhere/blob/main/src/common/comp/brute_force.cc

ThatikondV:
Hi @alexanderguzhva,

Thanks for the clarification. I’ve reviewed brute_force.cc — it indeed defines brute-force search paths for all data types, including fp16, bf16, and int8.

However, my original concern is specifically about index creation, not search. Based on logs like:

use key FLAT_bf16 to create knowhere index FLAT
replace index FLAT_bf16 with FLAT_fp32

it seems the system falls back to fp32 even before search begins, during the index construction stage.

Could you clarify:


- Where the fallback logic resides for index types?
- Does FAISS support lower precision data types for index creation or it is only limited to fp32 for index creation during build ?

[...]

PR #1212, opened 2025-06-10
Enable SVE Support for L2 Metric Computation in BF16 functions

ThatikondV:
**Description:**

This PR introduces SVE (Scalable Vector Extension) enablement for L2 metric computation in BF16 functions. These enhancements provide notable performance improvements over the existing NEON implementation, especially on ARM platforms with SVE capabilities.

**Changes in This PR:**

Added SVE optimizations for L2 metric computation in BF16 functions.
Refactored internal compute kernels to utilize SVE BF16 intrinsics.

**Benchmark Results:**

I have created performance test cases by considering a million vectors of dimension 256 as currently I am facing an issue in benchmarking bf16 data type. Issue: #1193
image

**Key observations:**

- All of the functions exhibit performance gains with SVE when compared with NEON.

/kind improvement
Fixes https://github.com/zilliztech/knowhere/issues/782


abhijain1204fujitsu:
@presburger, @alexanderguzhva , kindly support to review the PR

alexanderguzhva:
@abhijain1204fujitsu @ThatikondV taking a look...

src/simd/distances_sve.cc
```diff
@@ -506,6 +506,116 @@ int8_vec_L2sqr_batch_4_sve(const int8_t* x, const int8_t* y0, const int8_t* y1,
...
+            pg = svwhilelt_b16(i, d);
+
+        svbfloat16_t bf16_x = svld1_bf16(pg, reinterpret_cast<const __bf16*>(x + i));
+        svbfloat16_t bf16_y = svld1_bf16(pg, reinterpret_cast<const __bf16*>(y + i));
+
+        acc_squared_norm_x = svbfdot_f32(acc_squared_norm_x, bf16_x, bf16_x);
+        acc_squared_norm_y = svbfdot_f32(acc_squared_norm_y, bf16_y, bf16_y);
+        acc_dot_product = svbfdot_f32(acc_dot_product, bf16_x, bf16_y);
+
+        i += svcnth();
```
alexanderguzhva, line 529: the reference code converts bf16 into fp32, but the provided code uses mathematically identical, but practically different approach
```C++
float
bf16_vec_L2sqr_ref(const knowhere::bf16* x, const knowhere::bf16* y, size_t d) {
    float res = 0;
        const float tmp = (float)x[i] - (float)y[i];
        res += tmp * tmp;
    }
    return res;
}
```

Did you have a chance to make a comparison whether your candidate code may lead to an accuracy loss? 

Also, what is the performance comparison vs a straightforward `bf16`->`fp32`->`svsub_f32`->`svmla_f32`?

src/simd/distances_sve.cc
```diff
@@ -506,6 +506,116 @@ int8_vec_L2sqr_batch_4_sve(const int8_t* x, const int8_t* y0, const int8_t* y1,
...
+        acc = svbfdot_f32(acc, x_vec, x_vec);
+
+        i += svcnth();
+    }
+
+    return svaddv_f32(svptrue_b32(), acc);
+}
+
+void
+bf16_vec_L2sqr_batch_4_sve(const knowhere::bf16* x, const knowhere::bf16* y0, const knowhere::bf16* y1,
```
alexanderguzhva, line 561: same

ThatikondV: @alexanderguzhva,

Thanks for the review!

I’ve added a benchmark to compare the performance and accuracy of the SVE bf16 implementations:

**Benchmark Setup:**

Vectors: 1 million pairs of 256-dim bf16 vectors

**Versions Compared:**

- Version 1 (SVE): Uses svbfdot intrinsics directly

- Version 2 (SVE): Converts bf16 to fp32 before computation (reference method)

- NEON: Baseline implementation using NEON intrinsics

**Performance Comparision:**

I have compared the performance for the bf16_vec_L2sqr_sve function for all the 3 versions and I am getting a similar performance gain for the other functions as well.

Version 1 (SVE)  : 0.02917 s
Version 2 (SVE)  : 0.03953 s
NEON Version     : 0.07355 s

I have also done the accuracy comparison for the developed code, and there is no accuracy loss in both the versions compared to the reference version.

alexanderguzhva: let it be then :)

PR #1218, opened 2025-06-18
Enable SVE Support for IP Metric Computation in INT8 functions

mohakgoel1:
**Description:**

This PR introduces SVE (Scalable Vector Extension) enablement for IP metric computation in INT8 functions. These enhancements provide notable performance improvements over the existing NEON implementation, especially on ARM platforms with SVE capabilities.

**Changes in This PR:**

Added SVE optimizations for IP metric computation in INT8 functions.

**Benchmark Results:**

Below are the results from the custom performance test cases we have created using 500000 vectors of dimension 256.

image


alexanderguzhva:
issue: #1219

PR #1220, opened 2025-06-19
feat: add RVV SIMD versions of fvec_L2sqr, fvec_L1, and fvec_Linf

lyd1992:
This PR adds RISC-V Vector (RVV) SIMD optimized implementations for the following distance functions:

- fvec_L2sqr
- fvec_L1
- fvec_Linf

1. New RVV SIMD kernels for L2, L1, and Linf distance calculations on float vectors.
2. API and results are fully compatible with the existing reference (scalar) implementations.
3. Significant performance improvements observed across a range of vector dimensions.

Benchmark Results:
image
- All results are numerically consistent with the reference implementation (max absolute difference < 0.003).
- Speedup is measured as (REF time / RVV time) for 10,000 repeats.

### DIFF
--- a/src/simd/distances_rvv.cc
+++ b/src/simd/distances_rvv.cc
@@ -77,6 +77,144 @@ fvec_inner_product_rvv(const float* x, const float* y, size_t d) {
     return __riscv_vfmv_f_s_f32m1_f32(sum_scalar);
 }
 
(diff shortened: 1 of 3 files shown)


alexanderguzhva:
issue: #1222

PR #1223, opened 2025-06-20
feat(simd): add RVV-optimized vector operations

lyd1992:
Introduces RISC-V Vector (RVV) optimized implementations for critical vector functions:

- fvec_norm_L2sqr_rvv
- fvec_L2sqr_ny_rvv
- fvec_inner_products_ny_rvv
- fvec_madd_rvv

Performance highlights (2048-dim vectors,Speedup Ratio, ny=2, repeat=10000):
image

Performance Comparison of fvec_L2sqr_ny and fvec_inner_products_ny at d=2048 (repeat=10000):
image

/kind improvement


camel-cdr:
Why unroll? That's what LMUL is for. (applies to the other RVV PRs as well)

Issue #1226, opened 2025-06-23
Incorrect Results When using SCANN Iterator with cosine metric


cqy123456:
When using SCANN's iterator interface with cosine distance metric, the returned results are incorrect (inconsistent with brute-force cosine calculations) and low recall.

PR #1227, opened 2025-06-23
Fix SCANN iterator problem on Cosine

liliu-z:
issue: #1226

### DIFF
--- a/thirdparty/faiss/faiss/impl/index_read.cpp
+++ b/thirdparty/faiss/faiss/impl/index_read.cpp
@@ -1461,7 +1461,7 @@ Index* read_index(IOReader* f, int io_flags) {
         READ1(ivpq->qbs2);
         READ1(ivpq->is_cosine);
         if (ivpq->is_cosine) {
-            READVECTOR(ivpq->norms);
+            READVECTOR(ivpq->inverse_norms);
         }
         read_ProductQuantizer(&ivpq->pq, f);
         read_InvertedLists(ivpq, f, io_flags);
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
issue: #1226

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -1461,7 +1461,7 @@ Index* read_index(IOReader* f, int io_flags) {
         READ1(ivpq->qbs2);
         READ1(ivpq->is_cosine);
         if (ivpq->is_cosine) {
-            READVECTOR(ivpq->norms);
+            READVECTOR(ivpq->inverse_norms);
```
chasingegg, line 1464: any compatibility issue?

alexanderguzhva: I think that it is fine, bcz it does not work as expected anyways ? So, basically, there's no point of keeping what's broken. Alternatively, one may change the code to invert norms during load/save.
But it depends on whether SCANN is widely used. @liliu-z

alexanderguzhva: Actually, we have examples of both ways
See https://github.com/zilliztech/knowhere/blob/76abfdc656ae29719c3ba5330df9cb4a047982c4/thirdparty/faiss/faiss/impl/index_write.cpp#L579
See https://github.com/zilliztech/knowhere/blob/76abfdc656ae29719c3ba5330df9cb4a047982c4/thirdparty/faiss/faiss/impl/index_write.cpp#L614

liliu-z: I checked, this norm is only for SCANN

PR #1232, opened 2025-06-26
[SIMD][RVV] Optimize int8 distance functions: ivec_inner_product_rvv & ivec_L2sqr_rvv with RVV SIMD

lyd1992:
This PR introduces RVV SIMD optimized implementations for the following int8 distance functions:

- ivec_inner_product_rvv
- ivec_L2sqr_rvv
- Key changes

Key changes:
Utilized RISC-V Vector (RVV) intrinsics for efficient vectorized computation.
Correctly handled vector length, sign extension, and reduction to ensure accuracy.

Test Results:
Extensive correctness and performance tests were conducted across multiple dimensions.
All results matched the reference implementation (diff=0), and significant speedups were observed:
image

Summary:
Correctness: All results are bitwise identical to the reference implementation.
Performance: RVV SIMD achieves 2x~8x speedup over the scalar baseline, with greater gains at higher dimensions.

### DIFF
--- a/src/simd/distances_rvv.cc
+++ b/src/simd/distances_rvv.cc
@@ -500,6 +500,60 @@ fvec_L2sqr_batch_4_rvv(const float* x, const float* y0, const float* y1, const f
     dis3 = __riscv_vfmv_f_s_f32m1_f32(sum_scalar);
 }
 
+int32_t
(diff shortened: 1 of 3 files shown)


alexanderguzhva:
issue: #1234

Issue #1237, opened 2025-07-03
Fix the slow loading of sparse indexes in mmap mode


sparknack:
For now, because the saved format uses a forward index, loading a sparse index requires rebuilding the index. This works well for in-memory indexes, but when mmap mode is enabled, the massive and frequent disk I/O makes the loading phase so slow that it becomes unusable.

Thus, we need to add support for an inverted index format.

PR #1238, opened 2025-07-03
enhance: sparse: Add support for persistency of sparse indexes

sparknack:
issue: #1237 

For now, because the saved format uses a forward index, loading a sparse index requires rebuilding the index. This works well for in-memory indexes, but when mmap mode is enabled, the massive and frequent disk I/O makes the loading phase so slow that it becomes unusable.
Thus, add support for an inverted index format. Note that this requires a higher index version, which is 8.

### DIFF
--- a/include/knowhere/sparse_utils.h
+++ b/include/knowhere/sparse_utils.h
@@ -383,9 +383,9 @@ class GrowableVectorView {
         return reinterpret_cast<T*>(mmap_data_)[i];
     }
 
-    T*
+    const T*
     data() const {
-        return reinterpret_cast<T*>(mmap_data_);
+        return reinterpret_cast<const T*>(mmap_data_);
     }
 
     const T&
--- a/src/index/sparse/sparse_index_node.cc
+++ b/src/index/sparse/sparse_index_node.cc
@@ -270,7 +270,7 @@ class SparseInvertedIndexNode : public IndexNode {
         }
         MemoryIOWriter writer;
         if (version_use_raw_data()) {
-            RETURN_IF_ERROR(index_->Save(writer));
+            RETURN_IF_ERROR(index_->SerializeV0(writer));
         } else {
             RETURN_IF_ERROR(index_->Serialize(writer));
         }
@@ -298,7 +298,7 @@ class SparseInvertedIndexNode : public IndexNode {
         }
         index_ = index_or.value();
         if (version_use_raw_data()) {
-            return index_->Load(reader, 0, "");
+            return index_->DeserializeV0(reader, 0, "");
         } else {
             binary_ = binary;  // save the binary to avoid being freed
             return index_->Deserialize(reader);
@@ -336,7 +336,7 @@ class SparseInvertedIndexNode : public IndexNode {
         MemoryIOReader map_reader(reinterpret_cast<uint8_t*>(mapped_memory), map_size);
         if (version_use_raw_data()) {
             auto supplement_target_filename = filename + ".knowhere_sparse_index_supplement";
-            return index_->Load(map_reader, map_flags, supplement_target_filename);
+            return index_->DeserializeV0(map_reader, map_flags, supplement_target_filename);
         } else {
             mmap_guard_ = std::move(mmap_guard);
             return index_->Deserialize(map_reader);
--- a/src/index/sparse/sparse_inverted_index.h
+++ b/src/index/sparse/sparse_inverted_index.h
@@ -75,15 +75,15 @@ class BaseInvertedIndex {
     virtual ~BaseInvertedIndex() = default;
 
     virtual Status
-    Save(MemoryIOWriter& writer) = 0;
+    SerializeV0(MemoryIOWriter& writer) const = 0;
 
     // supplement_target_filename: when in mmap mode, we need an extra file to store the mmapped index data structure.
     // this file will be created during loading and deleted in the destructor.
     virtual Status
-    Load(MemoryIOReader& reader, int map_flags, const std::string& supplement_target_filename) = 0;
+    DeserializeV0(MemoryIOReader& reader, int map_flags, const std::string& supplement_target_filename) = 0;
 
     virtual Status
-    Serialize(MemoryIOWriter& writer) = 0;
+    Serialize(MemoryIOWriter& writer) const = 0;
 
     virtual Status
     Deserialize(MemoryIOReader& reader) = 0;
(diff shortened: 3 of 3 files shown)


include/knowhere/sparse_utils.h
```diff
@@ -383,6 +383,11 @@ class GrowableVectorView {
         return reinterpret_cast<T*>(mmap_data_)[i];
     }
 
+    T*
```
alexanderguzhva: `const T*`

src/index/sparse/sparse_index_node.cc
```diff
@@ -267,7 +269,11 @@ class SparseInvertedIndexNode : public IndexNode {
             return Status::empty_index;
         }
         MemoryIOWriter writer;
-        RETURN_IF_ERROR(index_->Save(writer));
+        if (version_use_raw_data()) {
+            RETURN_IF_ERROR(index_->Save(writer));
+        } else {
+            RETURN_IF_ERROR(index_->Serialize(writer));
```
alexanderguzhva, line 275: I would suggest a different logic here to make the code more readable
```C++
...
index_->Serialize(writer);
...

void Serialize(...) {
    if (version_use_raw_data()) {
        Serialize_One_Way(...);    // pick the right name
    } else {
        Serialie_Another_Way(...);    // pick the right name as well
    }
}
```

The same for the loading. Currently, it is hard to understand what's the difference between `Load` and `Deserialize`, you'll need to really read the code

```

File: src/index/sparse/sparse_index_node.cc [hunk 3]
```diff
@@ -291,7 +297,12 @@ class SparseInvertedIndexNode : public IndexNode {
             return index_or.error();
         }
         index_ = index_or.value();
-        return index_->Load(reader, 0, "");
+        if (version_use_raw_data()) {
+            return index_->Load(reader, 0, "");
+        } else {
+            binary_ = binary;  // save the binary to avoid being freed
+            return index_->Deserialize(reader);
```
alexanderguzhva, line 304: same

src/index/sparse/sparse_inverted_index.h
```diff
@@ -306,6 +333,301 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+
+    Status
+    Serialize(MemoryIOWriter& writer) override {
+        // Serialized format:
+        // 1. Index Header (36 bytes):
+        //    - index_format_version (uint32_t): Version of the index format, currently 1
+        //    - nr_rows (uint32_t): Number of rows in the index
+        //    - max_dim (uint32_t): Number of columns, or maximum dimension ID
+        //    - nr_inner_dims (uint32_t): Number of inner dimensions
+        //    - reserved (16 bytes): Reserved for future use
```
alexanderguzhva: you could add more reserved bytes to make a header to be 64 bytes aligned
Typically, it never hurts to reserve more

src/index/sparse/sparse_inverted_index.h
```diff
@@ -658,8 +1009,8 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
             return plist_vals_[loc_];
         }
 
-        const Vector<table_t>& plist_ids_;
-        const Vector<QType>& plist_vals_;
+        const boost::span<table_t>& plist_ids_;
```
alexanderguzhva: span is designed to be passed by value. Always.
```C++
const boost::span<table_t> plist_ids_;
```
also, if the data is supposed to be immutable, then please use `span<const table_t>`

src/index/sparse/sparse_inverted_index.h
sparknack: OK, but since cardinal already used the same header, let’s just keep it consistent for now?

src/index/sparse/sparse_index_node.cc
sparknack: OK, but since version is a member of SparseIndexNode, maybe the version check should still stay in `SparseIndexNode`. How about changing `index->Save()` to `index->SerializeV0()` instead?

src/index/sparse/sparse_index_node.cc
```diff
@@ -321,17 +332,15 @@ class SparseInvertedIndexNode : public IndexNode {
...
-
+        auto mmap_guard = std::make_unique<MmapGuard>(map_size, filename, mapped_memory);
         MemoryIOReader map_reader(reinterpret_cast<uint8_t*>(mapped_memory), map_size);
-        auto supplement_target_filename = filename + ".knowhere_sparse_index_supplement";
-        return index_->Load(map_reader, map_flags, supplement_target_filename);
+        if (version_use_raw_data()) {
+            auto supplement_target_filename = filename + ".knowhere_sparse_index_supplement";
+            return index_->Load(map_reader, map_flags, supplement_target_filename);
+        } else {
+            mmap_guard_ = std::move(mmap_guard);
```
zhengbuqian, line 341: can we create the mmap_guard in the else branch?

src/index/sparse/sparse_inverted_index.h
```diff
@@ -306,6 +333,301 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+        }
+#if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOHWERE_WITH_LIGHT)
+        // use a section to store some build stats for prometheus
+        nr_sections += 1;
+#endif
+        writer.write(&nr_sections, sizeof(uint32_t));
+
+        // since writer doesn't support seekp() for now, calculate all sizes of each sections first
+        std::vector<InvertedIndexSectionHeader> section_headers(nr_sections);
+        uint64_t used_offset = sizeof(InvertedIndexSectionHeader) * nr_sections + 36;
```
hhy3: put this magic number `36` into a variable

src/index/sparse/sparse_inverted_index.h
```diff
@@ -61,6 +82,12 @@ class BaseInvertedIndex {
     virtual Status
     Load(MemoryIOReader& reader, int map_flags, const std::string& supplement_target_filename) = 0;
 
+    virtual Status
+    Serialize(MemoryIOWriter& writer) = 0;
```
hhy3: mark const here

src/index/sparse/sparse_inverted_index.h
```diff
@@ -306,6 +333,301 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+        if (metric_type_ == SparseMetricType::METRIC_BM25) {
+            bm25_params_->row_sums_spans_ =
+                boost::span<float>(bm25_params_->row_sums.data(), bm25_params_->row_sums.size());
+        }
+
+        return Status::success;
+    }
+
+    Status
+    Serialize(MemoryIOWriter& writer) override {
```
hhy3: mark const

PR #1252, opened 2025-07-10
feat: Extend BitsetView for 1-hop ID mapping in hierarchical index_nodes

alwayslove2013:
Implements a 1-hop ID mapping mechanism to accelerate bitset checks in hierarchical index nodes. This avoids costly multi-level ID translations by composing nested ID maps into a single lookup table.

- `BitsetView` is extended to support an `out_id_map` and `offset`.
- `index_node` now exposes its internal ID map and accepts an external map from its parent, enabling the composition of multi-level re-layouts.
- `BitsetViewWithMappingIDSelector` is removed, and `BitsetViewIDSelector` is simplified to a basic wrapper.
- Adds comprehensive tests for the new external ID map functionality.

/kind feature

### DIFF
--- a/include/knowhere/bitsetview.h
+++ b/include/knowhere/bitsetview.h
@@ -23,8 +23,8 @@ class BitsetView {
     BitsetView() = default;
     ~BitsetView() = default;
 
-    BitsetView(const uint8_t* data, size_t num_bits, size_t filtered_out_num = 0)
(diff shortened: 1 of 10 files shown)


alexanderguzhva:
/lgtm
although, I wonder whether the performance remains the same

alexanderguzhva:
issue: #1258

alwayslove2013:
> /lgtm although, I wonder whether the performance remains the same

I will check the performance later.

PR #1263, opened 2025-07-15
Accelerating Data Processing with Prefetching

wjunLu:
Through SVE+PF optimizations, the following performance gains are achieved on 16U64G hardware configurations:

- Milvus-HNSW Algorithm
Benchmark: ANN-Benchmarks GIST dataset
Target Recall: ≥0.99
Performance Improvement: 20% higher QPS (Queries Per Second)

- Milvus-SCANN Algorithm
Benchmark: ANN-Benchmarks GIST dataset
Target Recall: ≥0.95
Performance Improvement: 20% higher QPS (Queries Per Second)

### DIFF
--- a/src/simd/distances_sve.cc
+++ b/src/simd/distances_sve.cc
@@ -30,6 +30,13 @@ fvec_L2sqr_sve(const float* x, const float* y, size_t d) {
         if (d - i < svcntw())
             pg = svwhilelt_b32(i, d);
 
+        // prefetch one vector
+        size_t prefetch_offset = i + svcntw();
+        if (prefetch_offset < d) {
+            __builtin_prefetch(x + prefetch_offset);
+            __builtin_prefetch(y + prefetch_offset);
+        }
+
         svfloat32_t a = svld1_f32(pg, x + i);
         svfloat32_t b = svld1_f32(pg, y + i);
         svfloat32_t diff = svsub_f32_m(pg, a, b);


alexanderguzhva:
@wjunLu A couple of questions
1) can a simple `__builtin_prefetch` for a GNU compiler be used instead?
2) a prefetch constant is a magic number. Any particular reasons why the provided magic number can chosen? Why not 2x lower or 2x higher?
Thanks.

wjunLu:
> @wjunLu A couple of questions
> 
> 1. can a simple `__builtin_prefetch` for a GNU compiler be used instead?
> 2. a prefetch constant is a magic number. Any particular reasons why the provided magic number can chosen? Why not 2x lower or 2x higher?
>    Thanks.

The current approach and `DEFAULT_PREFETCH_OFFSET=0x200` are both based on empirical values from ARMv8 testings, and there is no theoretical foundation to guarantee its absolute rationality. Of course, this is hardly convincing. 

As for `__builtin_prefetch`, it's a more flexible approach for prefetching, I can replace the current approach by it and using its default prefetch length. 

BTW. I saw many fields using `__builtin_prefetch` with different constants, too.

src/simd/distances_sve.cc
```diff
@@ -30,8 +30,8 @@ fvec_L2sqr_sve(const float* x, const float* y, size_t d) {
         if (d - i < svcntw())
             pg = svwhilelt_b32(i, d);
 
-        __builtin_prefetch(x + i, 0, 1);
-        __builtin_prefetch(y + i, 0, 1);
+        __builtin_prefetch(x + i, 0, 3);
```
alexanderguzhva: I think that the prefetch offset is gone

wjunLu: My fault!
I have added the new offset as length as one vector

alexanderguzhva:
@wjunLu would you please handle the DCO?
Thanks

wjunLu:
> @wjunLu would you please handle the DCO? Thanks

Done
Thanks for informing!

alexanderguzhva:
@wjunLu could you please clarify the following. Initially, your commit had a prefetch offset 0x200, which translates in 0x80 (?) fp32 values and which was claimed to be optimal. The current code has a prefetch offset, equal to the simd width of fp32s. Because the CPU pipeline length, it might be possible that automatic prefetchers (which definitely exist on x86, for example) will be able to handle this situation on its own?
I mean that I can accept the commit, just please make confirm that it is relatively optimal for dims like 768.
Thanks.

wjunLu:
@alexanderguzhva Sorry! - I missed your reply because of the weekend.

1. Yes, as you said that x86 indeed possesses exceptionally robust hardware automatic prefetching capabilities, it has no need to manually add prefetch processing.
 2. For ARM platforms, I know some series of CPUs have hardware prefetching capabilities too (e.g., Neoverse N1/N2). But at least to my knowledge(and based on our tests), ARM's hardware prefetching alone still fails to deliver optimal performance. Combining hardware prefetching with `__builtin_prefetch` can maximize efficiency as much as possible. 
 3. The selection of prefetch offsets lacks a theoretically deterministic optimum, current selected prefetch offset (1x vector length) prioritizes code maintainability over theoretical peak performance(and is more flexible than magic number like `0x200`). Empirical testing with 2x/4x offsets is recommended for latency-critical loops, though this may increase cache pollution in some ARM microarchitectures (e.g., Cortex-A55's limited prefetch queue depth).

alexanderguzhva:
issue: #1270

alexanderguzhva:
@wjunLu Thanks! Please feel free to check other distance functions, if you have the time and an opportunity.

wjunLu:
> @wjunLu Thanks! Please feel free to check other distance functions, if you have the time and an opportunity.

No problem, I will re-check them!

PR #1274, opened 2025-07-24
Fix ARM duplicate definitions of ivec_inner_product and ivec_L2sqr in hook.cc

lyd1992:
Fix ARM duplicate definitions of ivec_inner_product and ivec_L2sqr in hook.cc


alexanderguzhva:
issue: #1276

PR #1280, opened 2025-08-05
 Fix undeclared identifier error in RVV SIMD hook.cc

lyd1992:
During compilation, the build fails with an "undeclared identifier" error at src/simd/hook.cc:559 due to the nonexistent function fvec_inner_product_bf16_patch. 
image


alexanderguzhva:
issue: #1283

PR #1281, opened 2025-08-05
feat: Enable RVV support for knowhere_utils library

hansu2022:
This PR fixes the issue where the knowhere_utils library was compiled without the necessary RVV compiler options. The previous configuration prevented the RVV-optimized code path from being activated, even on hardware that supports it.
I've added the -march=rv64gcv_zvfhmin compiler option to knowhere_utils in cmake/libs/libfaiss.cmake. This ensures that hook.cc correctly recognizes and enables RVV-related optimizations during compilation.

Expected Outcome:
On a RISC-V platform with RVV support, simd_type will now correctly display as RVV.
image


alexanderguzhva:
issue: #1283

PR #1284, opened 2025-08-06
feat: Add RISC-V Vector Extension optimization for fvec_L2sqr_ny_transposed

lyd1992:
Implement RISC-V Vector Extension optimized version of fvec_L2sqr_ny_transposed function for computing L2 squared distances with transposed matrix storage.

Performance improvements:
- 1.69x-2.70x speedup for dimension 2048 (ny: 4-64)
- 1.65x-3.84x speedup for dimension 4096 (ny: 4-64)
- Best performance: 3.84x speedup (ny=64, d=4096)

image


alexanderguzhva:
issue: #1287

PR #1289, opened 2025-08-07
enhance: add cardinal v2 for knowhere

foxspy:


foxspy:
/hold 
proto code, still needs to be verified.

alexanderguzhva:
@foxspy is it good to review it?

cmake/libs/cardinal/v2/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
+project(knowhere CXX C)
+
+# Use short SHA1 as version
+set(CARDINAL_VERSION v2.6.1-v2)
```
alexanderguzhva, line 5: I'd call it 3.0, maybe, just to avoid the confusion

cmake/libs/libfaiss.cmake
```diff
@@ -188,8 +191,10 @@ if(__X86_64)
             -mavx512bw
             -mavx512vl
             -mpopcnt>)
+  target_include_directories(faiss_avx512 PRIVATE ${Boost_INCLUDE_DIRS})
 
   add_library(faiss STATIC ${FAISS_SRCS})
+  target_include_directories(faiss PRIVATE ${Boost_INCLUDE_DIRS})
```
alexanderguzhva, line 197: `faiss` needs `boost`?
The baseline version of `Faiss` did not need `boost`

cmake/libs/cardinal/v2/CMakeLists.txt
foxspy: It is currently still in the testing phase and will be replaced before the release.

cmake/libs/libfaiss.cmake
foxspy: knowhere/include/knowhere/dataset.h:25,
knowhere/include/knowhere/utils.h:21,
knowhere/thirdparty/faiss/faiss/IndexIVFFastScan.cpp:31

knowhere/include/knowhere/sparse_utils.h:18:10: fatal error: boost/iterator/iterator_facade.hpp: No such file or directory
   18 | #include <boost/iterator/iterator_facade.hpp>

header file dependency, but it’s not actually linked during compilation

PR #1291, opened 2025-08-07
add checks for a dimensionality of a query for BF, HNSW and IVF

alexanderguzhva:
/kind improvement
issue: #1290

### DIFF
--- a/src/common/comp/brute_force.cc
+++ b/src/common/comp/brute_force.cc
@@ -119,6 +119,17 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
     auto labels = std::make_unique<int64_t[]>(nq * topk);
     auto distances = std::make_unique<float[]>(nq * topk);
 
+    const std::string metric_str = cfg.metric_type.value();
+    if ((base_dataset->GetDim() != query_dataset->GetDim()) &&
+        (IsMetricType(metric_str, metric::COSINE) || IsMetricType(metric_str, metric::IP) ||
+         IsMetricType(metric_str, metric::L2))) {
+        const std::string msg_e =
+            fmt::format("dimensionalities of the base dataset ({}) and the query ({}) do not match",
+                        base_dataset->GetDim(), query_dataset->GetDim());
+        LOG_KNOWHERE_ERROR_ << msg_e;
+        return expected<DataSetPtr>::Err(Status::invalid_args, msg_e);
+    }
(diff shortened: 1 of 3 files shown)


alexanderguzhva:
@sparknack Hi, could you please confirm that the logic is valid for sparse vectors? Thanks.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -1473,6 +1482,16 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
         const auto* data = dataset->GetTensor();
 
         const auto hnsw_cfg = static_cast<const FaissHnswConfig&>(*cfg);
+
+        const std::string metric_str = hnsw_cfg.metric_type.value();
+        if ((dim != Dim()) && (IsMetricType(metric_str, metric::COSINE) || IsMetricType(metric_str, metric::IP) ||
+                               IsMetricType(metric_str, metric::L2))) {
+            const std::string msg_e =
+                fmt::format("dimensionalities of the base dataset ({}) and the query ({}) do not match", Dim(), dim);
```
foxspy, line 1490: put these repeated codes into a util function

foxspy: If I remember correctly, this should have already been intercepted in Milvus. What issue are you encountering here?

src/common/comp/brute_force.cc
```diff
@@ -119,6 +119,17 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
     auto labels = std::make_unique<int64_t[]>(nq * topk);
     auto distances = std::make_unique<float[]>(nq * topk);
 
+    const std::string metric_str = cfg.metric_type.value();
+    if ((base_dataset->GetDim() != query_dataset->GetDim()) &&
+        (IsMetricType(metric_str, metric::COSINE) || IsMetricType(metric_str, metric::IP) ||
```
sparknack, line 124: IP metrics could also work with sparse vectors.
Can we skip all the query dimension checks by identifying whether the `base_dataset` and `query_dataset` are sparse?

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: there was a problem in KIOXIA unit test that triggered a problem

PR #1293, opened 2025-08-13
add sparse_fp32 data type for sparse vertor

marcelo-cjl:
/kind improvement
issue: #1294

### DIFF
--- a/include/knowhere/operands.h
+++ b/include/knowhere/operands.h
@@ -146,6 +146,12 @@ struct bf16 {
     }
 };
 
+struct sparse_fp32 {
+    using type = float;
+};
+
+using sparsefp32 = sparse_fp32::type;
+
 template <typename T>
 bool
 typeCheck(uint64_t features) {
@@ -158,9 +164,11 @@ typeCheck(uint64_t features) {
     if constexpr (std::is_same_v<T, bf16>) {
         return features & knowhere::feature::BF16;
     }
-    // TODO : add sparse_fp32 data type
     if constexpr (std::is_same_v<T, fp32>) {
-        return (features & knowhere::feature::FLOAT32) || (features & knowhere::feature::SPARSE_FLOAT32);
+        return features & knowhere::feature::FLOAT32;
+    }
+    if constexpr (std::is_same_v<T, sparse_fp32>) {
+        return features & knowhere::feature::SPARSE_FLOAT32;
     }
     if constexpr (std::is_same_v<T, int8>) {
         return features & knowhere::feature::INT8;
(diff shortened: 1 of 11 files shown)


alexanderguzhva:
@marcelo-cjl what is the point of introducing this `sparse_fp32` structure? I understand when this is done for units (to distringuish meters from kilograms) or some index type, but what would be the justification for this particular case?

foxspy:
> @marcelo-cjl what is the point of introducing this `sparse_fp32` structure? I understand when this is done for units (to distringuish meters from kilograms) or some index type, but what would be the justification for this particular case?

"It's just some code optimization. Now float is used as the type for sparse, which makes it difficult to handle type judgments in some places."

include/knowhere/operands.h
```diff
@@ -146,6 +146,20 @@ struct bf16 {
...
+    sparse_fp32() = default;
+    sparse_fp32(const float& f) {
+        value = f;
+    };
+    operator float() const {
+        return value;
+    }
+
+ private:
+    float value;
```
foxspy: "Why is this value needed? I understand this is just a placeholder, and the actual representation of sparse will be more complex."

include/knowhere/operands.h
```diff
@@ -146,6 +146,20 @@ struct bf16 {
     }
 };
 
+struct sparse_fp32 {
+ public:
+    sparse_fp32() = default;
+    sparse_fp32(const float& f) {
```
hhy3: just pass by value, not const&

include/knowhere/operands.h
hhy3: initialize it

src/index/index_factory.cc
```diff
@@ -10,6 +10,7 @@
 // or implied. See the License for the specific language governing permissions and limitations under the License.
 
 #include "knowhere/index/index_factory.h"
+#include <cstdio>
```
hhy3: not used include?

include/knowhere/operands.h
marcelo-cjl: it has been deleted

include/knowhere/operands.h
marcelo-cjl: the content has been deleted

include/knowhere/operands.h
marcelo-cjl: the content has been deleted

src/index/index_factory.cc
marcelo-cjl: yes, the content has been deleted

marcelo-cjl:
kind/improvement
issue: #1294

marcelo-cjl:
@Mergifyio refresh

marcelo-cjl:
@mergifyio refresh

marcelo-cjl:
issue: #1293

PR #1282, opened 2025-08-05
feat: added support for AISAQ index - an all in storage index.

gland1:
See full description here:
https://github.com/zilliztech/knowhere/issues/1241

The index is added as part of thirdparty diskann library.

Signed-off-by:Gili Buzaglo <Gili.Buzaglo@il.kioxia.com>

Signed-off-by:Shimon Tsalmon <Shimon.Tsalmon@il.kioxia.com>

Signed-off-by:Hiroshi Murayama <hiroshi4.murayama@kioxia.com>

PR #1303, opened 2025-08-15
Fix benchmarks. Fix the compilation using clang.

alexanderguzhva:
Fix regressions of #1282 and #1289

/kind improvement
issue: #1302


foxspy (approved):
/lgtm

PR #1307, opened 2025-08-18
[Fix] Correct target for setting RISC-V compile options

lyd1992:
Changed target_compile_options from 'faiss' to 'knowhere_utils'
to resolve CMake configuration error in RISC-V builds.

Original error:
CMake Error at cmake/libs/libfaiss.cmake:140:
  Cannot specify compile options for target "faiss" which is not built by this project

The 'faiss' target is external while 'knowhere_utils' is the correct
local target needing RISC-V V extension flags (-march=rv64gcv_zvfhmin).
This fix enables proper vector optimization on RISC-V hardware.
image


lyd1992:
I sincerely apologize for this oversight on my part.
/kind improvement

alexanderguzhva:
issue: #1308

PR #1309, opened 2025-08-18
Add aisaq benchmark2

gland1:
We've added benchmarks for :
aisaq-p: aisaq performance version which consumes more diskspace than diskann but keeps high QPS and use near zero ram.
aisaq-s: aisaq scale version which consumes the same diskspace like diskann with near zero ram but with much lower QPS.

Fixed some compilation warnings and also optimized defaults for aisaq p.


alexanderguzhva:
waiting for UTs to finish

alexanderguzhva:
issue: #1311

PR #1312, opened 2025-08-20
fix: fix binary search crash in minhash_lsh index

cqy123456:
issue : https://github.com/zilliztech/knowhere/issues/1320

### DIFF
--- a/src/index/minhash/minhash_index_node.cc
+++ b/src/index/minhash/minhash_index_node.cc
@@ -308,9 +308,9 @@ MinHashLSHNode<DataType>::Search(const DataSetPtr dataset, std::unique_ptr<Confi
         } else {
             std::vector<folly::Future<folly::Unit>> futures;
             constexpr size_t batch_size = 64;
-            size_t run_time = (nq + batch_size - 1) / batch_size;
-            futures.reserve(nq);
-            for (size_t row = 0; row < run_time; ++row) {
+            size_t run_times = (nq + batch_size - 1) / batch_size;
+            futures.reserve(run_times);
+            for (size_t row = 0; row < run_times; ++row) {
                 futures.emplace_back(
                     search_pool_->push([&, beg = row * batch_size, end = std::min(int64_t((row + 1) * batch_size), nq),
                                         p_id_ptr = p_id.get(), p_dist_ptr = p_dist.get()]() {
--- a/src/index/minhash/minhash_lsh.h
+++ b/src/index/minhash/minhash_lsh.h
@@ -581,8 +581,8 @@ MinHashLSH::BatchSearch(const char* query, size_t nq, float* distances, idx_t* l
                                               this->mh_vec_elememt_size_, refine_k, topk, res_dis, res_ids);
                 return;
             }));
-            WaitAllSuccess(futures);
         }
+        WaitAllSuccess(futures);
     }
     return Status::success;
 }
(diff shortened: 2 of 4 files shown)


src/simd/distances_avx512.cc
```diff
@@ -914,70 +914,132 @@ rabitq_dp_popcnt_avx512(const uint8_t* q, const uint8_t* x, const size_t d, cons
...
+
     return -1;
 }
 int
 u64_binary_search_ge_avx512(const uint64_t* data, const size_t size, const uint64_t target) {
-    constexpr int SIMD_WIDTH = 8;
+    if (size == 0)
+        return -1;
+
+    if (size < 64) {
```
alexanderguzhva: the previous condition in the function above was `if (size <= 64)`. I just want to confirm that this is intentional

tests/ut/test_simd.cc
```diff
@@ -47,14 +47,159 @@ TEST_CASE("Test minhash function") {
     auto simd_type = knowhere::KnowhereConfig::SimdType::AVX512;
     knowhere::KnowhereConfig::SetSimdType(simd_type);
     constexpr size_t seed = 111;
-    SECTION("test binary search funtion ") {
-        size_t size = 1200;
-        auto x = GenRandomVector<uint64_t>(size, 1, seed);
+    SECTION("test binary search function with comprehensive corner cases") {
+        auto size =
+            GENERATE(as<size_t>{}, 1, 2, 4, 7, 8, 12, 14, 16, 21, 28, 32, 35, 42, 49, 56, 64, 111, 128, 256, 1000);
+        auto x = GenRandomVector<uint64_t>(size, 1, seed, 10000);
```
alexanderguzhva: please do a verification for several random vectors, just add several (5?) iterations of generating `x` and testing it

src/simd/distances_avx512.cc
cqy123456: You're right. This code is redundant. If the size is less than 32,  jump to the non-SIMD path

tests/ut/test_simd.cc
cqy123456: updated

foxspy:
analyzer should be fixed @cqy123456

src/index/minhash/minhash_index_node.cc
```diff
@@ -309,7 +309,7 @@ MinHashLSHNode<DataType>::Search(const DataSetPtr dataset, std::unique_ptr<Confi
             std::vector<folly::Future<folly::Unit>> futures;
             constexpr size_t batch_size = 64;
             size_t run_time = (nq + batch_size - 1) / batch_size;
```
foxspy: run_times

src/index/minhash/minhash_lsh.h
```diff
@@ -581,8 +581,8 @@ MinHashLSH::BatchSearch(const char* query, size_t nq, float* distances, idx_t* l
                                               this->mh_vec_elememt_size_, refine_k, topk, res_dis, res_ids);
                 return;
             }));
-            WaitAllSuccess(futures);
         }
+        WaitAllSuccess(futures);
```
foxspy, line 585: Was it running serially before...?

foxspy (approved):
/lgtm

src/index/minhash/minhash_lsh.h
alexanderguzhva: I hope it has not been running in N^2 mode, because the `futures` got never cleared and I hope that finished `future` object would not re-run again

alexanderguzhva: and it is great that the performance can be improved quite a lot, it seems!

PR #1326, opened 2025-08-25
enhance: add int8 fallback

foxspy:
issue: #1325 
/kind improvement


tests/ut/test_get_vector.cc
```diff
@@ -128,7 +128,6 @@ TEST_CASE("Test Float Get Vector By Ids", "[Float GetVectorByIds]") {
         knowhere::Json json;
         json[knowhere::meta::DIM] = dim;
         json[knowhere::meta::METRIC_TYPE] = metric;
-        json[knowhere::meta::RETRIEVE_FRIENDLY] = true;
```
chasingegg, line 131: why this is removed

chasingegg: This config is deprecated.

PR #1332, opened 2025-08-29
fix: fix sse compile

foxspy:


alexanderguzhva:
@foxspy is it good? Should `e2e-sse` pass?

PR #1337, opened 2025-09-01
Update estimate load resource interface

chasingegg:
/kind improvement

num_rows and dim are useful for estimating indexes that are partial load


include/knowhere/index/index_static.h
```diff
@@ -71,12 +71,15 @@ class IndexStaticFaced {
...
      * @param file_size sum of all index file/binary sizes (in GB)
+     * @param num_rows number of rows in the index
+     * @param dim dimension of the vector
      * @param params the union of the index build and load parameters
      * @return memory and disk usage for loading the index
      */
     static expected<Resource>
     EstimateLoadResource(const knowhere::IndexType& indexType, const knowhere::IndexVersion& version,
-                         const float file_size, const knowhere::Json& params);
+                         const float file_size, const int64_t num_rows, const int64_t dim,
```
alexanderguzhva: same question here: does it make sense to pass `file_size` as a `float` instead of `uint64_t`?

chasingegg: the value is in GB, rename the variable to make it clear

chasingegg: I will try to refactor the caller side in milvus to use byte size which is more accurate, currently the index_size is / 1024 /1024 /1024 first to get GB then pass to knowhere then * 1024 * 1024 * 1024 to get byte size does not make sense at all...

foxspy (approved):
/lgtm

PR #1347, opened 2025-09-16
hnsw binary support

alwayslove2013:
feat: add binary vector support for HNSW

- Add QT_1bit_direct quantizer type for binary vectors
- Add BinarySQDistanceComputerWrapper for binary distance computation
- Implement binary data format conversion (bin1) in HNSW index

/kind improvement

### DIFF
--- a/src/index/hnsw/faiss_hnsw.cc
+++ b/src/index/hnsw/faiss_hnsw.cc
@@ -490,6 +490,15 @@ convert_rows_to_fp32(const void* const __restrict src_in, float* const __restric
             }
         }
         return true;
+    } else if (src_data_format == DataFormatEnum::bin1) {
+        const knowhere::bin1* const src = reinterpret_cast<const knowhere::bin1*>(src_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows; i++) {
+            for (size_t j = 0; j < uint8_dim; ++j) {
+                dst[i * dim + j] = (float)(src[offsets[i] * uint8_dim + j]);
+            }
+        }
+        return true;
     } else {
         // unknown
         return false;
@@ -522,6 +531,23 @@ convert_rows_to_fp32(const void* const __restrict src_in, float* const __restric
         const knowhere::int8* const src = reinterpret_cast<const knowhere::int8*>(src_in);
         for (size_t i = 0; i < nrows * dim; i++) {
             dst[i] = (float)(src[i + start_row * dim]);
+        }
+        return true;
+    } else if (src_data_format == DataFormatEnum::bin1) {
+        // NOTE: This is a little bit weird conversion. The source (`src_in`) is a uint8_t byte stream,
+        // where each query_row has ((dim + 7) / 8) * 8 bits, and the total is nrows * ((dim + 7) / 8) * 8 bits.
+        // But the final format required is nrows * dim * 32 bits (float).
+        // There are actually two conversions happening here:
+        // 1. Each uint8_t value must be converted to float (in `BinarySQDistanceComputerWrapper::set_query`
+        //    and `ScalarQuantizer::compute_codes`), it will be converted back to uint8_t). [same as int8]
+        // 2. Each row must occupy dim * 32 bits of space, even if not all bits are filled;
+        //    this is required by the convention set in `ScalarQuantizer::compute_codes`.
+        const knowhere::bin1* const src = reinterpret_cast<const knowhere::bin1*>(src_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows; i++) {
+            for (size_t j = 0; j < uint8_dim; j++) {
+                dst[i * dim + j] = (float)(src[(start_row + i) * uint8_dim + j]);
+            }
         }
         return true;
     } else {
@@ -561,6 +587,16 @@ convert_rows_from_fp32(const float* const __restrict src, void* const __restrict
             dst[i + start_row * dim] = (knowhere::int8)src[i];
         }
         return true;
+    } else if (dst_data_format == DataFormatEnum::bin1) {
+        knowhere::bin1* const dst = reinterpret_cast<knowhere::bin1*>(dst_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows * uint8_dim; i++) {
+            KNOWHERE_THROW_IF_NOT_MSG(src[i] >= std::numeric_limits<knowhere::bin1>::min() &&
+                                          src[i] <= std::numeric_limits<knowhere::bin1>::max(),
+                                      "convert float to bin1(uint8_t) overflow");
+            dst[i + start_row * uint8_dim] = (knowhere::bin1)src[i];
+        }
+        return true;
     } else {
         // unknown
         return false;
@@ -578,6 +614,8 @@ convert_ds_to_float(const DataSetPtr& src, DataFormatEnum data_format) {
         return ConvertFromDataTypeIfNeeded<knowhere::bf16>(src);
     } else if (data_format == DataFormatEnum::int8) {
         return ConvertFromDataTypeIfNeeded<knowhere::int8>(src);
+    } else if (data_format == DataFormatEnum::bin1) {
+        return ConvertFromDataTypeIfNeeded<knowhere::bin1>(src);
     }
     return nullptr;
 }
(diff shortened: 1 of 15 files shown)


thirdparty/faiss/faiss/MetricType.h
```diff
@@ -52,8 +52,7 @@ using idx_t = int64_t;
 /// this function is used to distinguish between min and max indexes since
 /// we need to support similarity and dis-similarity metrics in a flexible way
 constexpr bool is_similarity_metric(MetricType metric_type) {
-    return ((metric_type == METRIC_INNER_PRODUCT) ||
-            (metric_type == METRIC_Jaccard));
+    return metric_type == METRIC_INNER_PRODUCT;
```
alwayslove2013, line 55: @alexanderguzhva, could you please help check the impact of this change? For knowhere, a smaller Jaccard value indicates higher similarity (closer distance).

alexanderguzhva:
@alwayslove2013  There's `faiss::IndexBinaryHSNW` class. Do I get it correct that it was decided not to use this class?

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -524,6 +533,23 @@ convert_rows_to_fp32(const void* const __restrict src_in, float* const __restric
             dst[i] = (float)(src[i + start_row * dim]);
         }
         return true;
+    } else if (src_data_format == DataFormatEnum::bin1) {
+        // NOTE: This is a little bit weird conversion. The source (`src_in`) is a uint8_t byte stream,
+        // where each query_row has (dim / 8) * 8 bits, and the total is nrows * (dim / 8) * 8 bits.
```
alexanderguzhva: `((dim + 7) / 8) * 8` bits?

alwayslove2013: You're correct, the code does use (dim + 7) / 8. I simplified the comment to dim / 8 for brevity and readability, but I agree it's confusing.
I value your opinion — if you think the full formula is clearer, I'm happy to change it back immediately.

alwayslove2013:
Yes, exactly.

> There's `faiss::IndexBinaryHSNW` class. Do I get it correct that it was decided not to use this class?

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -490,6 +490,15 @@ convert_rows_to_fp32(const void* const __restrict src_in, float* const __restric
             }
         }
         return true;
+    } else if (src_data_format == DataFormatEnum::bin1) {
+        const knowhere::bin1* const src = reinterpret_cast<const knowhere::bin1*>(src_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows; i++) {
+            for (size_t j = 0; j < uint8_dim; ++j) {
+                dst[i * dim + j] = (float)(src[offsets[i] * uint8_dim + j]);
```
alexanderguzhva, line 498: what does this code do exactly?
please correct me if I'm wrong, but I see the following. This code grabs bit-packed code in `uint8_t` format and puts it without unpacking into float data. This does not lose the precision. But the intention is not to unpack `uint8_t` values into `float` values containing 0 and 1.

src/index/hnsw/faiss_hnsw.cc
alexanderguzhva: I'd change it

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -561,6 +587,16 @@ convert_rows_from_fp32(const float* const __restrict src, void* const __restrict
...
         }
         return true;
+    } else if (dst_data_format == DataFormatEnum::bin1) {
+        knowhere::bin1* const dst = reinterpret_cast<knowhere::bin1*>(dst_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows * uint8_dim; i++) {
+            KNOWHERE_THROW_IF_NOT_MSG(src[i] >= std::numeric_limits<knowhere::bin1>::min() &&
+                                          src[i] <= std::numeric_limits<knowhere::bin1>::max(),
+                                      "convert float to bin1(uint8_t) overflow");
+            dst[i + start_row * uint8_dim] = (knowhere::bin1)src[i];
```
alexanderguzhva, line 597: same

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -524,6 +533,23 @@ convert_rows_to_fp32(const void* const __restrict src_in, float* const __restric
...
+        // There are actually two conversions happening here:
+        // 1. Each uint8_t value must be converted to float (in `BinarySQDistanceComputerWrapper::set_query`
+        //    and `ScalarQuantizer::compute_codes`), it will be converted back to uint8_t). [same as int8]
+        // 2. Each row must occupy dim * 32 bits of space, even if not all bits are filled;
+        //    this is required by the convention set in `ScalarQuantizer::compute_codes`.
+        const knowhere::bin1* const src = reinterpret_cast<const knowhere::bin1*>(src_in);
+        auto uint8_dim = (dim + 7) / 8;
+        for (size_t i = 0; i < nrows; i++) {
+            for (size_t j = 0; j < uint8_dim; j++) {
+                dst[i * dim + j] = (float)(src[(start_row + i) * uint8_dim + j]);
```
alexanderguzhva, line 549: the comment confirms my guess

thirdparty/faiss/faiss/MetricType.h
alexanderguzhva: as far as I remember, the Jaccard value in Knowhere and Faiss are opposite (like `x` and `1-x`).

thirdparty/faiss/faiss/impl/ScalarQuantizer.cpp
```diff
@@ -164,9 +170,21 @@ void ScalarQuantizer::decode(const uint8_t* codes, float* x, size_t n) const {
...
+    }
+    if (metric == METRIC_Hamming) {
+        assert(qtype == QT_1bit_direct);
+        return sq_get_hamming_distance_computer(metric, qtype, d, trained);
+    }
+    if (metric == METRIC_Jaccard) {
+        assert(qtype == QT_1bit_direct);
+        return sq_get_jaccard_distance_computer(metric, qtype, d, trained);
+    }
 }
```
alexanderguzhva, line 186: your code does not handle a default case, a C++ compiler will issue warnings, so I'd use
```C++
    if (metric == METRIC_Hamming) {
        assert(qtype == QT_1bit_direct);
        return sq_get_hamming_distance_computer(metric, qtype, d, trained);
    }
    if (metric == METRIC_Jaccard) {
        assert(qtype == QT_1bit_direct);
        return sq_get_jaccard_distance_computer(metric, qtype, d, trained);
    }
    return sq_get_distance_computer(metric, qtype, d, trained);
```

thirdparty/faiss/faiss/impl/ScalarQuantizerCodec.h
```diff
@@ -608,6 +638,46 @@ SQDistanceComputer* select_distance_computer(
...
+struct BinarySQDistanceComputerWrapper : SQDistanceComputer {
+    BinaryComputerType binary_computer;
+    size_t code_size;
+    std::vector<uint8_t> tmp;
+
+    BinarySQDistanceComputerWrapper(size_t code_size, const std::vector<float>&)
+            : code_size(code_size), tmp(code_size) {}
+
+    void set_query(const float* x) final {
+        for (size_t i = 0; i < code_size; ++i) {
```
alexanderguzhva, line 653: so, a code here implies that bit-packed `uint8_t` values will be passed via `float` values.

alexanderguzhva:
overall, looks good, I just want to confirm that I understand the code correctly

src/index/hnsw/faiss_hnsw.cc
alwayslove2013: OK, I will change it.

src/index/hnsw/faiss_hnsw.cc
alwayslove2013: Yes, this is quite awkward, and I also felt uncomfortable writing this part... However, to stay consistent with the logic of 'HNSW supporting int8 via SQ'.

thirdparty/faiss/faiss/impl/ScalarQuantizerCodec.h
alwayslove2013: yes

thirdparty/faiss/faiss/MetricType.h
foxspy: For the term `distance`, smaller values are better; for the term `similarity`, larger values are better. This needs to align with the operator used for the calculation. It's essential to verify whether the specific computation logic represents distance or similarity.

From the perspective of JaccardComputer, here represents the `distance`, not `similarity`

foxspy (approved):
/lgtm

PR #1352, opened 2025-09-23
feat: emb list support within index_node

alwayslove2013:
This PR refactors embedding list support by moving the implementation to the base IndexNode class, enabling most indexes to support emb_list functionality through inheritance.

## Key Changes
- Removed index types with "emb_list_" prefix
- Moved emb_list processing logic to base `IndexNode` class
- Most indexes can now support emb_list functionality through inheritance, if they implement the id map and `CalcDistByIDs`

### Current emb_list supported Index Types
- HNSW / HNSW_PQ / HNSW_PRQ / HNSW_SQ
  - data_type: fp32 / fp16 / bf16 / int8
  - metric_type: MAX_SIM_L2 / MAX_SIM_COSINE / MAX_SIM_IP
- HNSW
  - data_type: binary
  - metric_type: MAX_SIM_JACCARD / MAX_SIM_HAMMING


include/knowhere/config.h
```diff
@@ -644,6 +644,7 @@ class BaseConfig : public Config {
      * - A factor for top-k in the first ANNS round.
      */
     CFG_FLOAT retrieval_ann_ratio;
+    CFG_STRING emb_list_meta_file_name;
```
foxspy: emb_list_meta_file_path

alexanderguzhva:
@alwayslove2013 what about the problem with e2e-sse test?

foxspy:
> @alwayslove2013 what about the problem with e2e-sse test?

This PR is unrelated; the issue here is that after some dependency packages were switched from static linking to dynamic linking, the pyknowhere.so package did not fully include the dependencies. Additionally, the SSE CI build machine and the test machine are in two different environments, which causes the SSE CI to fail.

include/knowhere/index/index_node.h
```diff
@@ -470,7 +479,164 @@ class IndexNode : public Object {
...
+        return SetBaseIndexIDMap();
+    }
+
+    /**
+     * @brief Serializes the index to a binary set, including the emb_list meta if is emb_list.
+     *
+     * @param binset The BinarySet to store the serialized index.
+     * @return Status indicating success or failure of the serialization.
+     */
+    virtual Status
```
foxspy, line 564: This approach is not applicable to DiskANN.

include/knowhere/index/index_node.h
```diff
@@ -470,7 +479,164 @@ class IndexNode : public Object {
...
+            std::memcpy(&size, binary_ptr->data.get(), sizeof(size_t));
+            const auto total_bytes = binary_ptr->size;
+            const auto comp_size = total_bytes / sizeof(size_t) - 1;
+            if (comp_size != size) {
+                LOG_KNOWHERE_WARNING_ << "the computed size of emb_list offset is not equal to size from binary set";
+                return Status::emb_list_inner_error;
+            }
+            std::vector<size_t> offset(size);
+            std::memcpy(offset.data(), binary_ptr->data.get() + sizeof(size_t), size * sizeof(size_t));
+            emb_list_offset_ = std::make_unique<EmbListOffset>(std::move(offset));
```
foxspy, line 639: This approach is not applicable to DiskANN.

include/knowhere/index/index_node.h
alwayslove2013: Yes, DiskANN is not supported. Need further discussion.

include/knowhere/config.h
alwayslove2013: done.

PR #1354, opened 2025-10-09
fix compile error of gpu index

marcelo-cjl:
Fix error of incorrect inherited function parameters


marcelo-cjl:
/kind improvement
issue: #1355

PR #1366, opened 2025-10-22
pyknowhere support emb_list

alwayslove2013:
/kind improvement

### DIFF
--- a/include/knowhere/emb_list_utils.h
+++ b/include/knowhere/emb_list_utils.h
@@ -27,6 +27,7 @@ class EmbListOffset {
         assert(lims[idx] == 0);
         assert(rows > 0);
         while (lims[idx] < rows) {
+            assert(idx == 0 || lims[idx] >= lims[idx - 1]);
             offset.push_back(lims[idx]);
             idx++;
         }
--- a/python/knowhere/__init__.py
+++ b/python/knowhere/__init__.py
@@ -94,7 +94,27 @@ def ArrayToDataSet(arr):
             return swigknowhere.Array2DataSetBF16(arr)
     raise ValueError(
         """
-        ArrayToDataSet only support numpy array dtype float32,uint8,float16 and bfloat16.
+        ArrayToDataSet only support numpy array dtype float32, uint8, int8, float16 and bfloat16.
+        """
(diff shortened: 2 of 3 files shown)


python/knowhere/__init__.py
```diff
@@ -98,6 +98,25 @@ def ArrayToDataSet(arr):
...
+        return swigknowhere.Array2DataSetF(arr, offsets)
+    if arr.dtype == np.float16:
+        arr = arr.astype(np.float32)
+        return swigknowhere.Array2DataSetFP16(arr, offsets)
+    if arr.dtype == bfloat16:
+        arr = arr.astype(np.float32)
+        return swigknowhere.Array2DataSetBF16(arr, offsets)
+    raise ValueError(
+        """
+        ArrayToDataSetWithOffsets only support numpy array dtype float32,uint8,float16 and bfloat16.
```
alexanderguzhva: and int8?

python/knowhere/knowhere.i
```diff
@@ -313,18 +315,37 @@ GetNullBitSetView() {
     return nullptr;
 };
 
+void setOffsets(knowhere::DataSetPtr ds, int* offsets, int offset_size) {
+    size_t* offsets_ = new size_t[offset_size];
+    for (int i = 0; i < offset_size; ++i) {
+        offsets_[i] = (size_t)(offsets[i]);
```
alexanderguzhva, line 321: so, offsets are expected to be positive, right? Because there are no checks whether an offset is within a correct range

include/knowhere/emb_list_utils.h
```diff
@@ -27,6 +27,7 @@ class EmbListOffset {
         assert(lims[idx] == 0);
         assert(rows > 0);
         while (lims[idx] < rows) {
+            assert(idx == 0 || lims[idx] >= lims[idx - 1]);
             offset.push_back(lims[idx]);
             idx++;
         }
```
alwayslove2013, line 33: @alexanderguzhva will check when generating the EmbListOffset structure.
> so, offsets are expected to be positive, right? Because there are no checks whether an offset is within a correct range.

PR #1380, opened 2025-10-30
fix: correct the bitset configuration in the Faiss HNSW unit tests

foxspy:
/kind enhancement

### DIFF
--- a/include/knowhere/bitsetview.h
+++ b/include/knowhere/bitsetview.h
@@ -93,6 +93,7 @@ class BitsetView {
         id_offset_ = id_offset;
     }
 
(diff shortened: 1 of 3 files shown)


alexanderguzhva:
@foxspy is the failure in UTs signal that there's something wrong in threshold, or is it a problem in the fix itself?

foxspy:
> @foxspy is the failure in UTs signal that there's something wrong in threshold, or is it a problem in the fix itself?

It looks like there’s still a problem with this pr; I'm still investigating.

chasingegg (approved):
/lgtm

foxspy:
issue: #921

PR #1385, opened 2025-11-12
add emb list UT for tiered index

chasingegg:
/kind improvement


alexanderguzhva:
@chasingegg should we wait for the change in cardinal?

chasingegg:
> @chasingegg should we wait for the change in cardinal?

@alexanderguzhva This UT will not be ran in knowhere's CI, and I've tested it in local test, so this PR is ready to be merged. And we could trigger the UT in cardinal's CI to verify it.

foxspy (approved):
/lgtm

PR #1393, opened 2025-11-19
fix: build portable pyknowhere

foxspy:
issue: #1392 
/kind improvement

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -43,15 +43,13 @@ jobs:
           CIBW_SKIP: "*musllinux*" # faiss-cpu fails to build on musllinux
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_BEFORE_ALL_LINUX: "bash scripts/python_deps.sh && rm -rf build && mkdir build && cd build && conan install .. --build=missing -o with_diskann=True -s compiler.libcxx=libstdc++11 -s build_type=Release && conan build .. && cd -"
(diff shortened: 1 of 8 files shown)


.github/workflows/release-python.yml
```diff
@@ -43,15 +43,13 @@ jobs:
...
-          CIBW_BEFORE_BUILD: "pip3 install pytest numpy faiss-cpu bfloat16"
-          # CIBW_ENVIRONMENT_MACOS: >
-          #   _PYTHON_HOST_PLATFORM=macosx-10.15-${{ matrix.arch }}
-          # CIBW_BEFORE_BUILD: "bash scripts/python_deps.sh && pip3 install pytest numpy faiss-cpu"
-          # CIBW_CONFIG_SETTINGS: --build-option='-c clang'
-          # CIBW_REPAIR_WHEEL_COMMAND_MACOS: >
-          #   DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}
-          CIBW_TEST_REQUIRES: "pytest numpy faiss-cpu"
+          # Use portable wheel build script instead of plain setup.py
+          CIBW_BEFORE_BUILD: "pip3 install pytest numpy faiss-cpu bfloat16 auditwheel"
```
alexanderguzhva, line 47: `bfloat16` won't compile with numpy 2 on ARM, because `bfloat16==1.2.0` is available only for x86. For ARM, there's only `bfloat16==1.1.0`, which needs numpy 1. 

Scripts do reflect this, but do we really still rely on numpy 1?

foxspy: There are two independent reasons why pyknowhere must constrain numpy<2.0:
1. The SWIG interface files are configured to use NumPy 1.7 C API https://github.com/zilliztech/knowhere/blob/main/python/knowhere/knowhere.i#L27
2. Platform availability: as you mentioned — the impact comes from bfloat; upgraded to version 2.0.

It can be upgraded to 2.0 later, but this PR doesn’t include those changes.

PR #1396, opened 2025-11-24
FEAT: Support RVV-optimized Scalar Quantizer

hansu2022:
implements RISC-V Vector (RVV 1.0) optimizations for Scalar Quantizer (SQ) in Knowhere. By leveraging vector intrinsics and register grouping (`m4`), we significantly accelerate distance computations for HNSW+SQ indices.

### Key Changes

- **Optimization**: Implemented RVV-accelerated distance computers for standard SQ types (Int8, 4-bit, 6-bit, FP16, BF16).
    
- **Build System**: Added `-march=rv64gcv_zvfhmin` to support necessary FP16 intrinsics.
    
- **Integration**: Registered RVV implementations in `FaissHook` to enable automatic dispatch on supported hardware.

### Performance Benchmarks

We conducted benchmarks on the **Gist dataset** with an HNSW SQ index configuration (`efConstruction=256`, `M=16`). The comparison is between the scalar implementation (NoRVV) and this PR's optimization (RVV).

#### 1. Single Query Performance (Latency Sensitive)

|**efSearch**|**Recall@100**|**QPS (NoRVV)**|**QPS (RVV)**|**Speedup**|**Avg Latency (RVV)**|
|---|---|---|---|---|---|
|**16**|0.353|273.45|**711.24**|**2.60x**|1.41 ms|
|**32**|0.487|184.91|**515.46**|**2.79x**|1.94 ms|
|**64**|0.640|115.23|**339.33**|**2.94x**|2.95 ms|
|**128**|0.777|67.10|**207.04**|**3.08x**|4.83 ms|
|**256**|0.873|38.15|**115.47**|**3.03x**|8.66 ms|
|**512**|0.933|21.45|**65.41**|**3.05x**|15.29 ms|

#### 2. Batch Query Performance (Throughput Sensitive)

|**efSearch**|**Recall@100**|**QPS (NoRVV)**|**QPS (RVV)**|**Speedup**|
|---|---|---|---|---|
|**16**|0.353|10,143|**36,931**|**3.64x**|
|**32**|0.487|8,699|**21,837**|**2.51x**|
|**64**|0.640|6,334|**17,408**|**2.75x**|
|**128**|0.777|4,049|**11,792**|**2.91x**|
|**256**|0.873|2,322|**6,571**|**2.83x**|

**Conclusion:**

- Achieved **2.6x - 3.1x** speedup for single query latency is significantly reduced.
    
- Achieved **2.5x - 3.6x** speedup for batch queries.
    
- Recall rates remain identical, ensuring no accuracy loss.

/kind improvement

### DIFF
--- a/cmake/libs/libfaiss.cmake
+++ b/cmake/libs/libfaiss.cmake
@@ -177,6 +177,9 @@ if(__X86_64)
   knowhere_file_glob(GLOB FAISS_NEON_SRCS thirdparty/faiss/faiss/impl/*neon.cpp)
   list(REMOVE_ITEM FAISS_SRCS ${FAISS_NEON_SRCS})
 
+  knowhere_file_glob(GLOB FAISS_RVV_SRCS thirdparty/faiss/faiss/impl/*rvv.cpp)
+  list(REMOVE_ITEM FAISS_SRCS ${FAISS_RVV_SRCS})
+
   add_library(faiss_avx2 OBJECT ${FAISS_AVX2_SRCS})
   target_compile_options(faiss_avx2 PRIVATE $<$<COMPILE_LANGUAGE:CXX>: -msse4.2
                                             -mavx2 -mfma -mf16c -mpopcnt>)
@@ -221,6 +224,9 @@ if(__AARCH64)
   knowhere_file_glob(GLOB FAISS_AVX_SRCS thirdparty/faiss/faiss/impl/*avx.cpp)
   list(REMOVE_ITEM FAISS_SRCS ${FAISS_AVX_SRCS})
 
+  knowhere_file_glob(GLOB FAISS_RVV_SRCS thirdparty/faiss/faiss/impl/*rvv.cpp)
+  list(REMOVE_ITEM FAISS_SRCS ${FAISS_RVV_SRCS})
+
   add_library(faiss STATIC ${FAISS_SRCS})
   target_include_directories(faiss PRIVATE ${Boost_INCLUDE_DIRS})
 
@@ -253,7 +259,7 @@ if(__RISCV64)
   target_compile_options(
     faiss
     PRIVATE $<$<COMPILE_LANGUAGE:CXX>:
-            -march=rv64gcv
+            -march=rv64gcv_zvfhmin
             -mabi=lp64d
             -Wno-sign-compare
             -Wno-unused-variable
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@hansu2022 looks good overall, but I see that there is a change in a needed arch type `rv64gcv_zvfhmin`. Correct me if I'm wrong, because I'm not familiar with RISC-V to that degree, but it seems that `zvhfmin` does not look too common nowadays. If I am correct, then would it be possible to add an alternative hook for the code that uses it? 
I would be happy to provide any further assistance, if needed. 
Thanks.

alexanderguzhva:
@hansu2022 Alternatively, we may agree to put all the changes without `zhvmin` and you'll publish a separate commit with a proper `FP16` handling. Please let me know if this works.

hansu2022:
> @hansu2022 looks good overall, but I see that there is a change in a needed arch type `rv64gcv_zvfhmin`. Correct me if I'm wrong, because I'm not familiar with RISC-V to that degree, but it seems that `zvhfmin` does not look too common nowadays. If I am correct, then would it be possible to add an alternative hook for the code that uses it? I would be happy to provide any further assistance, if needed. Thanks.

@alexanderguzhva Thanks for the review!
You are correct that zvfhmin is not common on legacy hardware. However, I included it to ensure the code is future-proof, based on the latest industry standards:
1. RVA23 Standard (Mandatory): The recently ratified [RVA23 Profile](https://www.google.com/url?sa=E&q=https%3A%2F%2Friscv.org%2Fblog%2Frisc-v-announces-ratification-of-the-rva23-profile-standard%2F) explicitly makes zvfhmin mandatory for 64-bit application processors. It is the baseline for the next generation of RISC-V SoCs.
2. Toolchain Readiness: Compiler support is already mature. GCC 14.2+ and LLVM 17+ fully support this extension.
3. Hardware Roadmap: Vendors (e.g., SiFive, Rivos) are currently developing RVA23-compliant chips, with commercial availability expected around late 2025.

Since the ecosystem is moving towards RVA23 as the standard, I believe supporting it now is beneficial. Given this context, are you okay with keeping it in this PR?
(To address your concern about compatibility without dropping this future-standard optimization, I can implement a dispatch hook (runtime check).)

alexanderguzhva:
@hansu2022 Actually, it seems that we already have `zhfvmin` in our codebase, check https://github.com/zilliztech/knowhere/blob/e5aabf572be7079bd8b8639447e64b357dbe3883/cmake/libs/libfaiss.cmake#L141
So, I accept the PR

Issue #1398, opened 2025-12-03
vector index search interface supports cancellation


foxspy:
For tasks that have timed out or are no longer needed, the API should supports cancellation to terminate invalid tasks early.Retry

PR #1399, opened 2025-12-03
enhance: add search task cancel check

foxspy:
issue: #1398 
/kind improvement

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -14,11 +14,9 @@
 
 #include <functional>
 #include <queue>
-#include <unordered_set>
 #include <utility>
 #include <vector>
 
-#include "common/OpContext.h"
 #include "knowhere/binaryset.h"
 #include "knowhere/bitsetview.h"
 #include "knowhere/config.h"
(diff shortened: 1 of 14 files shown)


include/knowhere/index/index_node.h
```diff
@@ -316,24 +321,29 @@ class IndexNode : public Object {
...
+            } else {
+                for (size_t i = 0; i < nq; i++) {
+                    futs.emplace_back(ThreadPool::GetGlobalSearchThreadPool()->push([&, idx = i]() {
+                        ThreadPool::ScopedSearchOmpSetter setter(1);
+                        task_with_unordered_iterator(idx);
+                    }));
+                }
             }
+            WaitAllSuccess(futs);
+        } catch (const std::exception& e) {
```
alexanderguzhva: do I get it correct that the new code is expected to catch an exception from a thread cancellation? If so, then why the most general `std::exception` is caught, not a specific cancellation exception?

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -1658,74 +1660,80 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
...
+            }
 
-                    if (hnsw_cfg.range_filter.value() != defaultRangeFilter) {
-                        FilterRangeSearchResultForOneNq(result_dist_array[idx], result_id_array[idx],
-                                                        is_similarity_metric, radius, range_filter);
-                    }
-                }));
+            // wait for the completion
+            WaitAllSuccess(futs);
+        } catch (const std::exception& e) {
```
alexanderguzhva, line 1733: same question about an exception

include/knowhere/index/index_node.h
foxspy: No, I think since this interface already returns an error code, it shouldn’t also throw an exception. Throwing exceptions directly wasn’t appropriate in the earlier implementation.

PR #1403, opened 2025-12-12
feat: add emb_list support for DiskANN

alwayslove2013:
/kind improvement

### DIFF
--- a/src/index/diskann/diskann.cc
+++ b/src/index/diskann/diskann.cc
@@ -12,6 +12,8 @@
 #include "knowhere/feder/DiskANN.h"
 
 #include <cstdint>
+#include <fstream>
+#include <limits>
 
 #include "diskann/aux_utils.h"
 #include "diskann/linux_aligned_file_reader.h"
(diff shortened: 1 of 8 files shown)


src/index/diskann/diskann.cc
```diff
@@ -225,6 +274,53 @@ namespace knowhere {
...
+Status
+ReadEmbListOffsetFromFile(const std::string& file_path, std::vector<size_t>& offsets) {
+    std::ifstream in_file(file_path, std::ios::binary);
+    if (!in_file) {
+        LOG_KNOWHERE_ERROR_ << "Failed to open emb_list offset file for reading: " << file_path;
+        return Status::emb_list_inner_error;
+    }
+
+    size_t size = 0;
+    in_file.read(reinterpret_cast<char*>(&size), sizeof(size_t));
```
alexanderguzhva, line 286: reading & writing `size_t` is not a good idea, in general, because it is a platform-dependent constant
I'd stick with `uint64_t`

PR #1409, opened 2025-12-24
feat: support brute_force search and iterator with chunked dataset

alwayslove2013:
/kind improvement

support chunked dataset for BruteForce
- [Search] dense vector
  - metric: L2 / COSINE / IP
  - datatype: fp32 / fp16 / bf16 / int 8
- [Search] binary vector search 
  - metric: HAMMING / JACCARD
  - datatype: bin1
- [Search] emb list search
  - metric: MAX_SIM_L2 / MAX_SIM_COSINE / MAX_SIM_IP
  - datatype: fp32 / fp16 / bf16 / int 8
  - metric: MAX_SIM_HAMMING / MAX_SIM_JACCARD
  - datatype: bin1
  
- [AnnIterator] dense vector
  - metric: L2 / COSINE / IP
  - datatype: fp32 / fp16 / bf16 / int 8
- [AnnIterator] binary vector 
  - metric: HAMMING / JACCARD
  - datatype: bin1

### DIFF
--- a/include/knowhere/dataset.h
+++ b/include/knowhere/dataset.h
@@ -28,7 +28,7 @@ namespace knowhere {
 
 class DataSet : public std::enable_shared_from_this<const DataSet> {
  public:
-    typedef std::variant<const float*, const size_t*, const int64_t*, const void*, int64_t, std::string, std::any> Var;
+    using Var = std::variant<const float*, const size_t*, const int64_t*, const void*, int64_t, std::string, std::any>;
     DataSet() = default;
     ~DataSet() {
         if (!is_owner) {
@@ -58,6 +58,11 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
                 if (ptr != nullptr) {
                     if (is_sparse) {
                         delete[](sparse::SparseRow<float>*)(*ptr);
+                    } else if (is_chunk) {
+                        for (auto i = 0; i < num_chunk; i += 1) {
+                            delete[]((char**)(*ptr))[i];
+                        }
+                        delete[](char**)(*ptr);
                     } else {
                         delete[](char*)(*ptr);
                     }
@@ -287,6 +292,30 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
         this->is_sparse = is_sparse;
     }
 
+    bool
+    GetIsChunk() const {
+        std::unique_lock lock(mutex_);
+        return this->is_chunk;
+    }
+
+    void
+    SetIsChunk(bool is_chunk) {
+        std::unique_lock lock(mutex_);
+        this->is_chunk = is_chunk;
+    }
+
+    int64_t
+    GetNumChunk() const {
+        std::unique_lock lock(mutex_);
+        return this->num_chunk;
+    }
+
+    void
+    SetNumChunk(int64_t num_chunk) {
+        std::unique_lock lock(mutex_);
+        this->num_chunk = num_chunk;
+    }
+
     int64_t
     GetTensorBeginId() const {
         std::shared_lock lock(mutex_);
@@ -322,6 +351,8 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
     std::map<std::string, Var> data_;
     bool is_owner = true;
     bool is_sparse = false;
+    bool is_chunk = false;
+    int64_t num_chunk = 1;
 };
 using DataSetPtr = std::shared_ptr<DataSet>;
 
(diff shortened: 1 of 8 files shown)


include/knowhere/dataset.h
```diff
@@ -58,6 +58,12 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
                 if (ptr != nullptr) {
                     if (is_sparse) {
                         delete[](sparse::SparseRow<float>*)(*ptr);
+                    } else if (is_chunk) {
+                        auto num_chunk = GetRows();
```
foxspy: Please try to avoid changing the semantics of GetRows. You can add a GetChunkNum to express that instead.

include/knowhere/dataset.h
```diff
@@ -58,6 +58,12 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
...
+                        for (auto i = 0; i < num_chunk; i += 1) {
+                            delete[]((float**)(*ptr))[i];
```
foxspy: This could be float / bf16 / fp16 / int8 types here.

include/knowhere/utils.h
```diff
@@ -144,37 +144,63 @@ data_type_conversion(const DataSet& src, const std::optional<int64_t> start = st
...
     }
 
-    // map
-    auto* des_data = new OutType[out_dim * count_rows];
+    auto des = std::make_shared<DataSet>();
+    bool is_chunk = src.GetIsChunk();
+    if (is_chunk) {
+        auto* src_data = (const InType**)src.GetTensor();
+        auto* des_data = new OutType*[count_rows];
+        auto chunk_lims = src.Get<const size_t*>(knowhere::meta::EMB_LIST_OFFSET);
```
foxspy, line 157: Please add validation checks—otherwise chunk_lims could cause an NPE.

include/knowhere/dataset.h
```diff
@@ -287,6 +293,18 @@ class DataSet : public std::enable_shared_from_this<const DataSet> {
         this->is_sparse = is_sparse;
     }
 
+    bool
```
foxspy, line 295: Is there a concurrent read/write scenario here? Why need to add a lock?

include/knowhere/utils.h
```diff
@@ -144,37 +144,63 @@ data_type_conversion(const DataSet& src, const std::optional<int64_t> start = st
...
+                    des_data[i][j * out_dim + d] = (OutType)src_data[start_row + i][j * in_dim + d];
+                }
+            }
+        }
+        des->SetTensor(des_data);
+        des->SetIsChunk(true);
+        des->SetIsOwner(true);
+        des->SetRows(count_rows);
+        des->SetDim(out_dim);
+        des->SetIsOwner(true);
```
foxspy: SetIsOwner twice

thirdparty/faiss/faiss/utils/binary_distances.cpp
```diff
@@ -788,7 +788,7 @@ void all_hamming_distances_template(
                     output_i[j] = hc.compute(y_j);
                 } else {
                     // Set to a large value for non-selected vectors
-                    output_i[j] = std::numeric_limits<int>::max();
+                    output_i[j] = std::numeric_limits<float>::infinity();
```
alexanderguzhva: no.
Faiss uses `std::numeric_limits<int>::max()` as a default `big number`, not infinity.

include/knowhere/dataset.h
alwayslove2013: In my view, num_rows should represent the length of the tensor array. However, if you insist, I can fix num_rows as num_vectors and add a chunk_num parameter only when is_chunk is enabled.

include/knowhere/dataset.h
alwayslove2013: how about using `char*`?

include/knowhere/dataset.h
alwayslove2013: just follow `is_sparse` hhhh
https://github.com/zilliztech/knowhere/blob/f4b3e312f0fe8f255c80ab787f241fe7d26b4c68/include/knowhere/dataset.h#L279-L282

thirdparty/faiss/faiss/utils/binary_distances.cpp
alwayslove2013: My mistakes. The Hamming distance calculation result should be `int32_t`. I will revise all related parts.

include/knowhere/dataset.h
foxspy: Yes—I think we should keep the existing semantics unchanged as much as possible. Any new logic can be implemented by adding new interfaces/APIs.

alwayslove2013: updated.

PR #1414, opened 2025-12-27
Add AVX512 to sparse inverted index (IP)

lyang24:
Implements AVX512-optimized compute_all_distances() for sparse inverted index search for ip distance.

   Implementation details:

   1. AVX512 path (when hardware supports it):
      - 16-wide SIMD vectorization using _mm512_i32gather_ps() and _mm512_i32scatter_ps()
      - 2x loop unrolling (processes 32 elements per iteration) to hide gather latency
      - Static asserts for type safety (32-bit doc IDs, float values)
      - Supports IP metrics only.
      - IP: fully vectorized multiply-add with gather/scatter

   2. Scalar fallback (matches original code structure):
      - Simple double loop over query terms and posting lists
      - Metric type check inside inner loop (no optimizations added)
      - Works on all x86-64 CPUs and ARM/Apple Silicon

   3. Runtime dispatcher:
      - CPU capability detection using __builtin_cpu_supports()
      - Automatically selects AVX512 or scalar based on hardware

   Design decisions:

   - No AVX2: AVX2 gather is too slow (12-20 cycles vs 4 for scalar loads)
     and lacks hardware scatter, making it slower than scalar code

   - No manual prefetch: Random doc IDs in posting lists cannot be predicted
     by software prefetching, and manual prefetch pollutes cache and wastes
     memory bandwidth. Hardware prefetchers handle random access better.

   - Scalar fallback unchanged: Matches original implementation exactly to
     ensure correctness and avoid micro-optimizations that may not help
     
   - As shown on benchmark the avx512 bm25 is underperforming the baseline 
   thus it will processed with the scalar route

### DIFF
--- a/benchmark/benchmark_sparse_simd.cpp
+++ b/benchmark/benchmark_sparse_simd.cpp
@@ -1,0 +1,361 @@
+// Copyright (C) 2019-2023 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+// This benchmark is x86_64-specific due to AVX512 intrinsics
+// It should only be built on x86_64 systems (guarded in CMakeLists.txt)
+#if !defined(__x86_64__) && !defined(_M_X64) && !defined(__amd64__)
+#error "This benchmark requires x86_64 architecture. It should not be built on ARM/other platforms."
+#endif
+
+#include <boost/core/span.hpp>
+#include <chrono>
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <random>
+#include <unordered_set>
+#include <vector>
+
+#include "knowhere/sparse_utils.h"
+#include "simd/instruction_set.h"
+#include "simd/sparse_simd.h"
+
+using namespace knowhere::sparse;
+
+// Generate synthetic sparse data with realistic posting list distributions
+struct SparseDataset {
+    size_t n_docs;
+    size_t n_queries;
+    size_t vocab_size;
+    std::vector<std::vector<table_t>> posting_list_ids;
+    std::vector<std::vector<float>> posting_list_vals;
+    std::vector<std::pair<size_t, float>> query;
+
+    SparseDataset(size_t n_docs, size_t n_queries, size_t vocab_size, size_t avg_query_terms,
+                  size_t avg_posting_list_len, bool force_heavy_terms = false)
+        : n_docs(n_docs), n_queries(n_queries), vocab_size(vocab_size) {
+        std::mt19937 rng(42);
+        std::uniform_real_distribution<float> val_dist(0.1f, 1.0f);
+        std::uniform_int_distribution<size_t> term_dist(0, vocab_size - 1);
+        std::uniform_int_distribution<size_t> query_len_dist(avg_query_terms / 2, avg_query_terms * 2);
+
+        // Initialize posting lists
+        posting_list_ids.resize(vocab_size);
+        posting_list_vals.resize(vocab_size);
+
+        // Generate posting lists with zipf-like distribution
+        // Create a more realistic distribution where some terms are very frequent
+        std::vector<size_t> target_lengths(vocab_size);
+
+        // Use Zipf distribution: rank r has frequency proportional to 1/r^alpha
+        double alpha = 1.0;  // Zipf parameter
+        double sum = 0.0;
+        for (size_t r = 1; r <= vocab_size; ++r) {
+            sum += 1.0 / std::pow(r, alpha);
+        }
+
+        // Scale so average = avg_posting_list_len
+        double scale = avg_posting_list_len * vocab_size / sum;
+
+        for (size_t term_id = 0; term_id < vocab_size; ++term_id) {
+            size_t rank = term_id + 1;
+            double freq = scale / std::pow(rank, alpha);
+            target_lengths[term_id] = std::min(n_docs, std::max(size_t(1), static_cast<size_t>(freq)));
+        }
+
+        // Generate posting lists
+        for (size_t term_id = 0; term_id < vocab_size; ++term_id) {
+            size_t target_len = target_lengths[term_id];
+
+            std::vector<table_t> doc_ids;
+            std::uniform_int_distribution<table_t> doc_dist(0, n_docs - 1);
+
+            // Generate unique random doc IDs
+            std::unordered_set<table_t> seen;
+            while (doc_ids.size() < target_len) {
+                table_t doc_id = doc_dist(rng);
+                if (seen.insert(doc_id).second) {
+                    doc_ids.push_back(doc_id);
+                }
+            }
+
+            // Sort for cache-friendly access
+            std::sort(doc_ids.begin(), doc_ids.end());
+
+            posting_list_ids[term_id] = std::move(doc_ids);
+            posting_list_vals[term_id].resize(posting_list_ids[term_id].size());
+
+            for (size_t i = 0; i < posting_list_vals[term_id].size(); ++i) {
+                posting_list_vals[term_id][i] = val_dist(rng);
+            }
+        }
+
+        // Generate query
+        if (force_heavy_terms) {
+            // Force query to include heavy (frequent) terms with long posting lists
+            // This ensures SIMD actually gets exercised
+            size_t heavy_terms = std::min(size_t(10), vocab_size);
+            for (size_t i = 0; i < heavy_terms; ++i) {
+                query.push_back({i, val_dist(rng)});
+            }
+            // Add some random terms too
+            size_t random_terms = avg_query_terms - heavy_terms;
+            for (size_t i = 0; i < random_terms; ++i) {
+                size_t term_id = term_dist(rng);
+                query.push_back({term_id, val_dist(rng)});
+            }
+        } else {
+            // Random query generation
+            size_t query_len = query_len_dist(rng);
+            for (size_t i = 0; i < query_len; ++i) {
+                size_t term_id = term_dist(rng);
+                float weight = val_dist(rng);
+                query.push_back({term_id, weight});
+            }
+        }
+    }
+};
+
+// Timing utilities
+class Timer {
+    std::chrono::high_resolution_clock::time_point start_;
+
+ public:
+    Timer() : start_(std::chrono::high_resolution_clock::now()) {
+    }
+
+    double
+    elapsed_ms() const {
+        auto end = std::chrono::high_resolution_clock::now();
+        return std::chrono::duration<double, std::milli>(end - start_).count();
+    }
+
+    void
+    reset() {
+        start_ = std::chrono::high_resolution_clock::now();
+    }
+};
+
+// Benchmark runner
+void
+run_benchmark(const char* name, const SparseDataset& dataset) {
+    using QType = float;  // This benchmark uses float values
+
+    printf("\n=== %s ===\n", name);
+    printf("Dataset: %zu docs, %zu vocab, query length: %zu\n", dataset.n_docs, dataset.vocab_size,
+           dataset.query.size());
+
+    // Calculate posting list length statistics
+    size_t total_postings = 0;
+    size_t non_empty = 0;
+    size_t min_len = SIZE_MAX;
+    size_t max_len = 0;
+    std::vector<size_t> lengths;
+    for (const auto& plist : dataset.posting_list_ids) {
+        if (!plist.empty()) {
+            size_t len = plist.size();
+            total_postings += len;
+            non_empty++;
+            min_len = std::min(min_len, len);
+            max_len = std::max(max_len, len);
+            lengths.push_back(len);
+        }
+    }
+    std::sort(lengths.begin(), lengths.end());
+    size_t median_len = lengths.empty() ? 0 : lengths[lengths.size() / 2];
+
+    printf("Posting list stats: avg=%.1f, median=%zu, min=%zu, max=%zu\n",
+           non_empty > 0 ? (double)total_postings / non_empty : 0.0, median_len, min_len, max_len);
+
+    // Show top-10 heaviest terms (what queries should hit for SIMD benefit)
+    printf("Top-10 heaviest terms: ");
+    for (size_t i = 0; i < std::min(size_t(10), lengths.size()); ++i) {
+        printf("%zu ", lengths[lengths.size() - 1 - i]);
+    }
+    printf("\n");
+
+    // Prepare data structures
+    std::vector<boost::span<const table_t>> ids_spans;
+    std::vector<boost::span<const float>> vals_spans;
+    for (size_t i = 0; i < dataset.vocab_size; ++i) {
+        ids_spans.emplace_back(dataset.posting_list_ids[i]);
+        vals_spans.emplace_back(dataset.posting_list_vals[i]);
+    }
+
+    const int warmup_runs = 5;
+    const int bench_runs = 50;
+
+    // Check CPU capabilities
+#if defined(__x86_64__) || defined(_M_X64)
+    auto& inst_set = faiss::InstructionSet::GetInstance();
+    printf("CPU Capabilities: AVX512F=%d, AVX2=%d\n", inst_set.AVX512F(), inst_set.AVX2());
+#else
+    printf("CPU Capabilities: ARM/Apple Silicon (no SIMD)\n");
+#endif
+
+    std::vector<float> result_scalar, result_avx512;
+
+#ifdef __AVX512F__
+    {
+        for (int i = 0; i < warmup_runs; ++i) {
+            result_avx512.assign(dataset.n_docs, 0.0f);
+            for (const auto& [dim_idx, q_weight] : dataset.query) {
+                const auto& plist_ids = ids_spans[dim_idx];
+                const auto& plist_vals = vals_spans[dim_idx];
+
+                accumulate_posting_list_contribution_ip_dispatch<QType>(plist_ids.data(), plist_vals.data(),
+                                                                        plist_ids.size(), static_cast<float>(q_weight),
+                                                                        result_avx512.data());
+            }
+        }
+    }
+#endif
+
+    // Benchmark scalar
+    printf("\n[Scalar Fallback]\n");
+    Timer timer;
+    for (int i = 0; i < bench_runs; ++i) {
+        result_scalar.assign(dataset.n_docs, 0.0f);
+        for (size_t q_idx = 0; q_idx < dataset.query.size(); ++q_idx) {
+            const auto& plist_ids = ids_spans[dataset.query[q_idx].first];
+            const auto& plist_vals = vals_spans[dataset.query[q_idx].first];
+            const float q_weight = dataset.query[q_idx].second;
+
+            for (size_t j = 0; j < plist_ids.size(); ++j) {
+                const auto doc_id = plist_ids[j];
+                result_scalar[doc_id] += q_weight * plist_vals[j];
+            }
+        }
+    }
+    double scalar_time = timer.elapsed_ms() / bench_runs;
+    printf("  Time: %.3f ms\n", scalar_time);
+
+    // Count non-zero results for verification
+    size_t scalar_nonzero = 0;
+    for (float score : result_scalar) {
+        if (score > 1e-6f)
+            scalar_nonzero++;
+    }
+    printf("  Non-zero scores: %zu / %zu\n", scalar_nonzero, result_scalar.size());
+
+#ifdef __AVX512F__
+    {
+        printf("\n[SIMD Dispatcher (AVX512 if available)]\n");
+
+        timer.reset();
+        for (int i = 0; i < bench_runs; ++i) {
+            result_avx512.assign(dataset.n_docs, 0.0f);
+            for (const auto& [dim_idx, q_weight] : dataset.query) {
+                const auto& plist_ids = ids_spans[dim_idx];
+                const auto& plist_vals = vals_spans[dim_idx];
+
+                accumulate_posting_list_contribution_ip_dispatch<QType>(plist_ids.data(), plist_vals.data(),
+                                                                        plist_ids.size(), static_cast<float>(q_weight),
+                                                                        result_avx512.data());
+            }
+        }
+
+        double avx512_time = timer.elapsed_ms() / bench_runs;
+        printf("  Time: %.3f ms\n", avx512_time);
+        printf("  Using: %s\n", inst_set.AVX512F() ? "AVX512" : "Scalar");
+
+        size_t avx512_nonzero = 0;
+        for (float score : result_avx512) {
+            if (score > 1e-6f)
+                avx512_nonzero++;
+        }
+        printf("  Non-zero scores: %zu / %zu\n", avx512_nonzero, result_avx512.size());
+
+        double max_diff = 0.0;
+        double avg_diff = 0.0;
+        size_t diff_count = 0;
+        for (size_t i = 0; i < result_scalar.size(); ++i) {
+            double diff = std::abs(result_scalar[i] - result_avx512[i]);
+            if (diff > 1e-4) {
+                avg_diff += diff;
+                diff_count++;
+                max_diff = std::max(max_diff, diff);
+            }
+        }
+        if (diff_count > 0) {
+            avg_diff /= diff_count;
+        }
+
+        printf("\n[Verification]\n");
+        printf("  Max difference: %.6f\n", max_diff);
+        printf("  Avg difference: %.6f (over %zu elements)\n", avg_diff, diff_count);
+        printf("  Correctness: %s\n", (max_diff < 1e-3) ? "PASS" : "FAIL");
+
+        printf("\n[Performance]\n");
+        double speedup = scalar_time / avx512_time;
+        printf("  Speedup: %.2fx\n", speedup);
+        printf("  Scalar:  %.3f ms (baseline)\n", scalar_time);
+        printf("  AVX512:  %.3f ms (%.1f%% of baseline)\n", avx512_time, 100.0 * avx512_time / scalar_time);
+    }
+#else
+    printf("\n[AVX512 not compiled in (requires -mavx512f)]\n");
+#endif
+
+    printf("==========================================\n");
+}
+
+int
+main() {
+    printf("╔══════════════════════════════════════════════════════════════════╗\n");
+    printf("║  Sparse Inverted Index SIMD Benchmark                           ║\n");
+    printf("╚══════════════════════════════════════════════════════════════════╝\n");
+
+    // Test configurations
+    struct BenchConfig {
+        const char* name;
+        size_t n_docs;
+        size_t vocab_size;
+        size_t avg_query_terms;
+        size_t avg_posting_list_len;
+        bool force_heavy_terms;
+    };
+
+    std::vector<BenchConfig> configs = {
+        // Ultra-sparse: posting lists shorter than SIMD width (16)
+        {"Ultra-sparse IP (random query, avg=8)", 50000, 2000, 15, 8, false},
+        {"Ultra-sparse IP (heavy terms, avg=8)", 50000, 2000, 15, 8, true},
+
+        // Sparse: posting lists around SIMD width (16-32)
+        {"Sparse IP (random query, avg=32)", 100000, 5000, 20, 32, false},
+        {"Sparse IP (heavy terms, avg=32)", 100000, 5000, 20, 32, true},
+
+        // Medium density: posting lists 2-8x SIMD width (64-128)
+        {"Medium IP (random query, avg=128)", 500000, 8000, 25, 128, false},
+        {"Medium IP (heavy terms, avg=128)", 500000, 8000, 25, 128, true},
+
+        // Dense: posting lists 16-32x SIMD width (256-512)
+        {"Dense IP (random query, avg=512)", 1000000, 10000, 30, 512, false},
+        {"Dense IP (heavy terms, avg=512)", 1000000, 10000, 30, 512, true},
+
+        // Very dense: posting lists 64-128x SIMD width (1024-2048)
+        {"Very Dense IP (heavy terms, avg=2048)", 1000000, 10000, 30, 2048, true},
+
+        // Real-world-like: MSMARCO/Wikipedia scale
+        {"Real-world IP (avg=256, heavy head)", 1000000, 10000, 25, 256, true},
+    };
+
+    for (const auto& config : configs) {
+        SparseDataset dataset(config.n_docs, 1, config.vocab_size, config.avg_query_terms, config.avg_posting_list_len,
+                              config.force_heavy_terms);
+        run_benchmark(config.name, dataset);
+    }
+
+    printf("\n╔══════════════════════════════════════════════════════════════════╗\n");
+    printf("║  Benchmark completed                                             ║\n");
+    printf("╚══════════════════════════════════════════════════════════════════╝\n");
+
+    return 0;
+}
--- a/src/index/sparse/sparse_inverted_index.h
+++ b/src/index/sparse/sparse_inverted_index.h
@@ -34,6 +34,8 @@
 #include "knowhere/prometheus_client.h"
 #include "knowhere/sparse_utils.h"
 #include "knowhere/utils.h"
+#include "simd/instruction_set.h"
+#include "simd/sparse_simd.h"
 
 namespace knowhere::sparse {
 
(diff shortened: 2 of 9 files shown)


src/index/sparse/sparse_inverted_index.h
```diff
@@ -118,6 +128,43 @@ class BaseInvertedIndex {
     n_cols() const = 0;
 };
 
+// CPU capability detection for SIMD optimizations
+#if defined(__x86_64__) || defined(_M_X64)
+namespace simd {
```
hhy3: consider move these simd codes to src/simd folder

src/index/sparse/sparse_inverted_index.h
```diff
@@ -960,17 +1007,188 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+
+                    // Compute contribution
+                    __m256 contribution = _mm256_mul_ps(q_weight_vec, vals);
+
+                    // AVX2 lacks scatter, so we extract and write manually
+                    alignas(32) float contrib_array[8];
+                    _mm256_store_ps(contrib_array, contribution);
+
+                    for (size_t k = 0; k < SIMD_WIDTH; ++k) {
+                        scores[plist_ids[j + k]] += contrib_array[k];
```
sparknack: You can use `_mm256_i32gather_epi32` with AVX2.

src/index/sparse/sparse_inverted_index.h
lyang24: thank you moved all smid related code over

src/index/sparse/sparse_inverted_index.h
lyang24: good catch fixed

lyang24:
to maintainers - what is the best acceptance test/ bench suite for this? i think there might be little perf benefit is doc list is short

src/index/sparse/sparse_inverted_index.h
```diff
@@ -118,6 +119,178 @@ class BaseInvertedIndex {
     n_cols() const = 0;
 };
 
+// CPU feature detection at runtime
```
alexanderguzhva: please use `src/simd/instruction_set.h` instead

alexanderguzhva (commented):
any benchmarks / unit tests?

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,301 @@
...
+                alignas(32) uint32_t id_array[8];
+
+                _mm256_store_ps(current_scores_array, current_scores);
+                _mm256_store_ps(contrib_array, contribution);
+                _mm256_store_ps(doc_len_ratios_array, doc_len_ratios_vec);
+                _mm256_store_si256(reinterpret_cast<__m256i*>(id_array), doc_ids);
+
+                // Apply computer function and scatter
+                for (size_t k = 0; k < SIMD_WIDTH; ++k) {
+                    scores[id_array[k]] = current_scores_array[k] + computer(contrib_array[k], doc_len_ratios_array[k]);
```
alexanderguzhva: no `_mm256_scatter_ps`?

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,301 @@
...
+                __m256 new_scores = _mm256_add_ps(current_scores, contribution);
+
+                // Scatter back (AVX2 lacks scatter, so use scalar loop)
+                alignas(32) float new_scores_array[8];
+                alignas(32) uint32_t id_array[8];
+                _mm256_store_ps(new_scores_array, new_scores);
+                _mm256_store_si256(reinterpret_cast<__m256i*>(id_array), doc_ids);
+
+                for (size_t k = 0; k < SIMD_WIDTH; ++k) {
+                    scores[id_array[k]] = new_scores_array[k];
```
alexanderguzhva: please add `_mm256_scatter_ps` here

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,301 @@
...
+                alignas(64) float contrib_array[16];
+                alignas(64) float doc_len_ratios_array[16];
+                alignas(64) uint32_t id_array[16];
+
+                _mm512_store_ps(contrib_array, contribution);
+                _mm512_store_ps(doc_len_ratios_array, doc_len_ratios_vec);
+                _mm512_store_si512(reinterpret_cast<__m512i*>(id_array), doc_ids);
+
+                for (size_t k = 0; k < SIMD_WIDTH; ++k) {
+                    scores[id_array[k]] += computer(contrib_array[k], doc_len_ratios_array[k]);
```
alexanderguzhva: no `_mm512_scatter_ps` here?

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,301 @@
...
+            const auto& plist_vals = inverted_index_vals_spans[q_vec[i].first];
+            const float q_weight = q_vec[i].second;
+
+            size_t j = 0;
+
+            // SIMD loop with TRUE SCATTER
+            for (; j + SIMD_WIDTH <= plist_ids.size(); j += SIMD_WIDTH) {
+                // Prefetch
+                for (size_t k = 0; k < SIMD_WIDTH && j + PREFETCH_DISTANCE + k < plist_ids.size(); ++k) {
+                    _mm_prefetch(reinterpret_cast<const char*>(&scores[plist_ids[j + PREFETCH_DISTANCE + k]]), _MM_HINT_T1);
```
alexanderguzhva: why `_MM_HINT_T1` instead of `T0`?

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,301 @@
...
+            const auto& plist_vals = inverted_index_vals_spans[q_vec[i].first];
+            const float q_weight = q_vec[i].second;
+
+            size_t j = 0;
+
+            // SIMD loop
+            for (; j + SIMD_WIDTH <= plist_ids.size(); j += SIMD_WIDTH) {
+                // Prefetch ahead
+                for (size_t k = 0; k < SIMD_WIDTH && j + PREFETCH_DISTANCE + k < plist_ids.size(); ++k) {
+                    _mm_prefetch(reinterpret_cast<const char*>(&scores[plist_ids[j + PREFETCH_DISTANCE + k]]), _MM_HINT_T1);
```
alexanderguzhva: why `_MM_HINT_T1` instead of `T0`?
also, why is `_mm_prefetch()` used, while the earlier avx2 version uses `__builtin_prefetch()`? Would it be possible to use only 1 approach everywhere? :)

lyang24:
> any benchmarks / unit tests?

thank you for the meticulous review, will fix the suggestins add a comprehensive bench in the next few weeks.

lyang24:
> any benchmarks / unit tests?

I did some initial benchmark prefetching, and avx2 seems to hurt perf and avx512 is a mixed bag. cc @alexanderguzhva does this align with expection? The benchmark is run on aws ec2 instances and code is in the repo.

`cd build/Release
  make benchmark_sparse_simd -j8
  
  ./benchmark/benchmark_sparse_simd
`

ai summarized report

AVX512 SIMD Benchmark Results - Complete Summary

  Test Environment

  - CPU: Intel c7i.2xlarge (Ice Lake, AVX512F supported)
  - Compiler: GCC with -mavx512f -mavx512dq
  - Dataset: Synthetic data with Zipf distribution (realistic sparse search workload)

  Query Type Explanation

  The benchmark tests two query generation strategies:

  Random Query: Query terms selected randomly from vocabulary
  - Most terms are infrequent (short posting lists of ~1-10 documents)
  - SIMD loop rarely triggers due to insufficient elements
  - Represents unrealistic worst-case scenario

  Heavy Terms Query: Query forced to include the top-10 most frequent terms
  - Heavy terms have long posting lists (hundreds to thousands of documents)
  - SIMD loop gets properly exercised with sufficient elements to amortize overhead
  - Represents realistic search queries (common words like "the", "search", "data")

  Example from benchmark output:
  Sparse IP (avg=32):
    Top-10 heaviest terms: 17593 8796 5864 4398 3518 2932 2513 2199 1954 1759
    Median posting length: 7
  - Random query: Likely hits median-length terms (~7 elements) → SIMD overhead dominates
  - Heavy query: Guaranteed to hit top-10 terms (~2000-17000 elements) → SIMD wins

  Complete Results Table

  | Test Case                | Avg Posting Length | Query Type  | Metric | Speedup | Verdict      |
  |--------------------------|--------------------|-------------|--------|---------|--------------|
  | Ultra-sparse IP (random) | 8                  | Random      | IP     | 0.70x   | ❌ Slower    |
  | Ultra-sparse IP (heavy)  | 8                  | Heavy terms | IP     | 1.35x   | ✅ Faster    |
  | Sparse IP (random)       | 32                 | Random      | IP     | 0.73x   | ❌ Slower    |
  | Sparse IP (heavy)        | 32                 | Heavy terms | IP     | 2.11x   | ✅ Faster    |
  | Sparse BM25 (heavy)      | 32                 | Heavy terms | BM25   | 0.77x   | ❌ Slower    |
  | Medium IP (random)       | 128                | Random      | IP     | 0.50x   | ❌ Slower    |
  | Medium IP (heavy)        | 128                | Heavy terms | IP     | 1.58x   | ✅ Faster    |
  | Medium BM25 (heavy)      | 128                | Heavy terms | BM25   | 0.78x   | ❌ Slower    |
  | Dense IP (random)        | 512                | Random      | IP     | 0.87x   | ❌ Slower    |
  | Dense IP (heavy)         | 512                | Heavy terms | IP     | 1.69x   | ✅ Faster    |
  | Dense BM25 (heavy)       | 512                | Heavy terms | BM25   | 0.77x   | ❌ Slower    |
  | Very Dense IP (heavy)    | 2048               | Heavy terms | IP     | 3.06x   | ✅ Best Case |
  | Very Dense BM25 (heavy)  | 2048               | Heavy terms | BM25   | 0.80x   | ❌ Slower    |
  | Real-world IP            | 256                | Heavy terms | IP     | 1.44x   | ✅ Faster    |
  | Real-world BM25          | 256                | Heavy terms | BM25   | 0.78x   | ❌ Slower    |

  Key Findings

  ✅ When AVX512 Helps (6/16 cases):

  - IP metric with heavy (frequent) terms: 1.35x - 3.06x speedup
  - Best performance: Very dense posting lists (avg=2048) → 3.06x speedup
  - Realistic workload: Real-world distribution (avg=256) → 1.44x speedup

  ❌ When AVX512 Hurts (10/16 cases):

  1. Random queries (0.50x-0.87x): Queries that miss heavy terms process short posting lists where SIMD overhead dominates
  2. BM25 metric (0.77x-0.80x): Per-element DocValueComputer function call overhead negates SIMD benefits across all densities

  Why the Dichotomy?

  AVX512 processes 16 floats per iteration. Performance depends on whether posting lists are long enough:

  - Heavy terms query hitting top-10 terms: Posting lists of 1000-17000 elements → SIMD loop runs 60-1000+ iterations → overhead amortized ✅
  - Random query hitting median terms: Posting lists of 1-10 elements → SIMD loop skipped, only scalar tail → pure overhead ❌

  Real-world queries typically include common terms (stop words, domain keywords), making the "heavy terms" scenario more representative.

  Conclusions

  1. AVX512 is beneficial for IP metric with frequent terms - the primary use case for sparse search
  2. Posting list length matters: Need avg > 32 elements for consistent speedup
  3. Query distribution matters: Only queries hitting heavy terms benefit from SIMD
  4. BM25 needs algorithmic changes: Current per-element function call approach incompatible with SIMD efficiency

src/simd/instruction_set.h
```diff
@@ -370,5 +372,44 @@ class InstructionSet {
 };
 
 }  // namespace faiss
+#endif  // __x86_64__ || _M_X64
+
+// Lightweight CPU feature detection for sparse vector SIMD
```
alexanderguzhva: No.

`InstructionSet` class has all the needed facilities to perform checks.
Please remove `SIMDCapabilities` completely, because it is totally redundant. 
If any features are needed, then I would modify `InstructionSet` class.

Also, AVX2 code is not very interesting in 2026, so it looks like a correct decision to get rid of it.

alexanderguzhva:
overall, lgtm. please fix the formatting and this `InstructionSet` issue.
@sparknack do you have any suggestions about this code?

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,205 @@
...
+template <typename QType>
+inline std::vector<float>
+compute_all_distances_avx512(size_t n_rows_internal, const std::vector<std::pair<size_t, float>>& q_vec,
+                             const std::vector<boost::span<const table_t>>& inverted_index_ids_spans,
+                             const std::vector<boost::span<const QType>>& inverted_index_vals_spans,
+                             const DocValueComputer<float>& computer, SparseMetricType metric_type,
+                             const boost::span<const float>* doc_len_ratios_spans_ptr) {
+    // Static asserts for type safety
+    static_assert(sizeof(table_t) == 4, "SIMD gather requires 32-bit doc IDs");
+    static_assert(std::is_same_v<QType, float>, "SIMD operations require float values");
```
sparknack: the element type of `inverted_index_vals_spans` should be `uint16_t` for BM25, so will this function still be called?

lyang24: yes this is a bug, from benchmark bm25 underperforms on avx512 - i will fix this by left out bm25 on regular scalar path.

src/simd/instruction_set.h
lyang24: yep used instruction set instead

lyang24:
from the research paper https://arxiv.org/html/2509.08395v2#S1 we can also look into  below as future improvement


  1. Window Partitioning
  - What: Process documents in chunks of λ=100K instead of all N documents at once
  - Why: 400KB score array fits in L3 cache vs 4MB that doesn't
  - Tradeoff: +10-20% memory overhead, harder to update incrementally

  2. Mass Ratio Pruning

  - What: Keep only dimensions that sum to α% of total "mass" (e.g., α=0.5 = keep 50%)
  - Why: Low-value dimensions contribute little to ranking
  - Tradeoff: -0.5% to -5% recall loss depending on aggressiveness

  3. Two-Phase Search

  - What:
    - Phase 1: Fast approximate search with pruned index → get top-γ candidates
    - Phase 2: Exact recomputation for γ candidates → return top-k
  - Why: Recovers accuracy lost from pruning
  - Tradeoff: 2× memory (must store both pruned index + original vectors)

alexanderguzhva:
@lyang24 let me see

alexanderguzhva:
@lyang24 https://arxiv.org/html/2509.08395v2#S1  ha, I know this paper :)

src/simd/sparse_simd_avx512.cc
```diff
@@ -0,0 +1,108 @@
...
+
+// ============================================================================
+// AVX512 SIMD Implementation (16-wide vectorization with hardware scatter)
+// ============================================================================
+template <typename QType>
+std::vector<float>
+compute_all_distances_avx512(size_t n_rows_internal, const std::vector<std::pair<size_t, float>>& q_vec,
+                             const std::vector<boost::span<const table_t>>& inverted_index_ids_spans,
+                             const std::vector<boost::span<const QType>>& inverted_index_vals_spans,
+                             const DocValueComputer<float>& computer, SparseMetricType metric_type,
```
sparknack: Maybe the `computer` parameter should be removed, because for future BM25 implementations, we’ll still need to use SIMD calculations directly instead of relying on a scalar computer.

src/index/sparse/sparse_inverted_index.h
```diff
@@ -959,19 +1017,20 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
         }
-        return scores;
+
+        // Use SIMD dispatcher from src/simd/sparse_simd.h
+        const boost::span<const float>* row_sums_ptr =
+            (metric_type_ == SparseMetricType::METRIC_BM25) ? &bm25_params_->row_sums_spans_ : nullptr;
+
+        return compute_all_distances_simd_dispatch<QType>(n_rows_internal_, q_vec_float, inverted_index_ids_spans_,
+                                                          inverted_index_vals_spans_, computer, metric_type_,
+                                                          row_sums_ptr);
```
sparknack: how about this approach?
```cpp
if (metric_type_ == SparseMetricType::METRIC_IP) {
    for (const auto& q_item : q_vec) {
            const auto& plist_ids = inverted_index_ids_spans[q_item.first];
            const auto& plist_vals = inverted_index_vals_spans[q_item.first];
            const float q_weight = q_item.second;
            accumulate_posting_list_contribution_ip_dispatch(plist_ids.data(), plist_vals.data(), plist_ids.size(), q_weight, scores.data());
    }
} else {
      for (const auto& q_item : q_vec) {
            const auto& plist_ids = inverted_index_ids_spans[q_item.first];
            const auto& plist_vals = inverted_index_vals_spans[q_item.first];
            const float q_weight = q_item.second;
            // accumulate_posting_list_contribution_bm25(...); or the original implementation
     }
}
```

This way, the scalar implementation avoids frequent metric_type_ checks.

Additionally, the SIMD logic becomes more modular, for example, sparse_simd.h can now avoid including `<boost/core/span.hpp>`, `vector` and `"knowhere/sparse_utils.h"`.

src/simd/sparse_simd_avx512.cc
lyang24: good call fixed

src/index/sparse/sparse_inverted_index.h
lyang24: great - i refactored the code

alexanderguzhva:
@lyang24 there are still compilation errors...

lyang24:
> @lyang24 there are still compilation errors...

its quite tricky the pre commit passed but i am not able to see the logs of e2e runs - i am guessing with each push. Is there a easy way i can validate before push? (i guess i have to build locally right).

alexanderguzhva:
@lyang24 well, I can help with letting you know about unusual errors, but I typically ensure that the code with my changes can be built locally & uts run before I push ones.

alexanderguzhva:
you may set up a compilation environment inside a docker container (ubuntu:22.04) if you don't want to spoil your primary development system with unneeded artifacts from conan.

alexanderguzhva:
@lyang24 the code seems to pass UTs. The formatting change is needed:
```
benchmark/benchmark_base.h
====================
--- original

+++ formatted

@@ -191,10 +191,10 @@

     void
     free_all() {
         if (xb_ != nullptr) {
-            delete[] (float*)xb_;
+            delete[](float*) xb_;
         }
         if (xq_ != nullptr) {
-            delete[] (float*)xq_;
+            delete[](float*) xq_;
         }
         if (gt_radius_ != nullptr) {
             delete[] gt_radius_;
```

@sparknack is this PR good to go?

lyang24:
> you may set up a compilation environment inside a docker container (ubuntu:22.04) if you don't want to spoil your primary development system with unneeded artifacts from conan.

Thank you Docker build is much easier and cleaner. One thing to note for future viewers doing this - is that we need to setup the docker container's network access to packages.zilliz.com so we can pull the folly lib

lyang24:
> @lyang24 the code seems to pass UTs. The formatting change is needed:
> 
> ```
> benchmark/benchmark_base.h
> ====================
> --- original
> 
> +++ formatted
> 
> @@ -191,10 +191,10 @@
> 
>      void
>      free_all() {
>          if (xb_ != nullptr) {
> -            delete[] (float*)xb_;
> +            delete[](float*) xb_;
>          }
>          if (xq_ != nullptr) {
> -            delete[] (float*)xq_;
> +            delete[](float*) xq_;
>          }
>          if (gt_radius_ != nullptr) {
>              delete[] gt_radius_;
> ```
> 
> @sparknack is this PR good to go?

Thank You, may i know what is the right clang format version to use in knowhere?
i am using 
```
clang-format --version
clang-format version 21.1.8
```
I think the there is a problem where pre commit hook do not ping the clang format version - the changes in this pr is not suppose to touch benchmark_base.h and this is what i saw :(

```
(conan311) lanqing@Lanqings-MacBook-Pro-2 knowhere % git checkout main benchmark/benchmark_base.h
(conan311) lanqing@Lanqings-MacBook-Pro-2 knowhere % git add  benchmark/benchmark_base.h
(conan311) lanqing@Lanqings-MacBook-Pro-2 knowhere % git commit --amend
check for added large files..............................................Passed
check for merge conflicts................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
clang-format.............................................................Failed
- hook id: clang-format
- exit code: 1

benchmark/benchmark_base.h
====================
--- original

+++ formatted

@@ -191,10 +191,10 @@

     void
     free_all() {
         if (xb_ != nullptr) {
-            delete[](float*) xb_;
+            delete[] (float*)xb_;
         }
         if (xq_ != nullptr) {
-            delete[](float*) xq_;
+            delete[] (float*)xq_;
         }
         if (gt_radius_ != nullptr) {
             delete[] gt_radius_;
```

alexanderguzhva:
@lyang24 my bad, it seems that I see this formatting problem as well. It turns out that our CI still uses an old clang-format. I also use `clang-21` and see the same problem

alexanderguzhva:
> > you may set up a compilation environment inside a docker container (ubuntu:22.04) if you don't want to spoil your primary development system with unneeded artifacts from conan.
> 
> Thank you Docker build is much easier and cleaner. One thing to note for future viewers doing this - is that we need to setup the docker container's network access to packages.zilliz.com so we can pull the folly lib

Technically speaking, the installation of `knowhere` components is done via `conan` package manager. Please check https://github.com/zilliztech/knowhere/blob/main/scripts/install_deps.sh as a rough list of needed commands to set it up.
The actual compilation on my machine is done via something like the following:
```bash
mkdir build
cd build
conan install .. --build=missing -o with_diskann=True -o with_ut=True -o with_benchmark=True -s compiler.libcxx=libstdc++11 -s build_type=Release --profile=clang21_release --install-folder=clang21 --output-folder=clang21
conan build .. --build-folder=clang21
```

lyang24:
> > > you may set up a compilation environment inside a docker container (ubuntu:22.04) if you don't want to spoil your primary development system with unneeded artifacts from conan.
> > 
> > 
> > Thank you Docker build is much easier and cleaner. One thing to note for future viewers doing this - is that we need to setup the docker container's network access to packages.zilliz.com so we can pull the folly lib
> 
> Technically speaking, the installation of `knowhere` components is done via `conan` package manager. Please check https://github.com/zilliztech/knowhere/blob/main/scripts/install_deps.sh as a rough list of needed commands to set it up. The actual compilation on my machine is done via something like the following:
> 
> ```shell
> mkdir build
> cd build
> conan install .. --build=missing -o with_diskann=True -o with_ut=True -o with_benchmark=True -s compiler.libcxx=libstdc++11 -s build_type=Release --profile=clang21_release --install-folder=clang21 --output-folder=clang21
> conan build .. --build-folder=clang21
> ```

thank you this is awesome

src/simd/sparse_simd.h
```diff
@@ -0,0 +1,42 @@
...
+#if defined(__x86_64__) || defined(_M_X64)
+void
+accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals, size_t list_size, float q_weight,
+                                  float* scores);
+#endif
+
+template <typename QType>
+inline void
+accumulate_posting_list_contribution_ip_dispatch(const uint32_t* doc_ids, const QType* doc_vals, size_t list_size,
+                                                 float q_weight, float* scores, const DocValueComputer<QType>& computer,
```
sparknack: Since the ip in `accumulate_posting_list_contribution_ip_dispatch()` already indicates this function is for IP computation, could we remove the `computer` parameter? 

For scalar calculations, maybe we could directly use `accumulate_posting_list_ip_scalar_ref()` instead.

.github/workflows/ut.yaml
```diff
@@ -48,6 +48,15 @@ jobs:
...
           && ./Release/tests/ut/knowhere_tests
+      - name: Run Sparse SIMD Benchmark
+        run: |
+          cd build
+          if [ -f ./Release/benchmark/benchmark_sparse_simd ]; then
+            echo "Running sparse SIMD benchmark..."
+            ./Release/benchmark/benchmark_sparse_simd
+          else
+            echo "Sparse SIMD benchmark not found, skipping..."
+          fi
```
sparknack: maybe we can remove this benchmark job from the unit tests check.

benchmarking isn’t essential for unit tests, tests/ut/test_sparse_simd.cc should be sufficient.

benchmark/benchmark_sparse_simd.cpp
```diff
@@ -0,0 +1,408 @@
...
+    for (size_t i = 0; i < dataset.vocab_size; ++i) {
+        ids_spans.emplace_back(dataset.posting_list_ids[i]);
+        vals_spans.emplace_back(dataset.posting_list_vals[i]);
+    }
+
+    boost::span<const float> doc_len_spans(dataset.doc_len_ratios);
+    const boost::span<const float>* doc_len_ptr =
+        (metric_type == SparseMetricType::METRIC_BM25) ? &doc_len_spans : nullptr;
+
+    SimpleBM25Computer computer;
```
sparknack: remove this unused variable.

benchmark/benchmark_sparse_simd.cpp
```diff
@@ -0,0 +1,408 @@
...
+        {"Medium BM25 (heavy terms, avg=128)", 500000, 8000, 25, 128, SparseMetricType::METRIC_BM25, true},
+
+        // Dense: posting lists 16-32x SIMD width (256-512)
+        // SIMD should show 3-4x speedup
+        {"Dense IP (random query, avg=512)", 1000000, 10000, 30, 512, SparseMetricType::METRIC_IP, false},
+        {"Dense IP (heavy terms, avg=512)", 1000000, 10000, 30, 512, SparseMetricType::METRIC_IP, true},
+        {"Dense BM25 (heavy terms, avg=512)", 1000000, 10000, 30, 512, SparseMetricType::METRIC_BM25, true},
+
+        // Very dense: posting lists 64-128x SIMD width (1024-2048)
+        // SIMD should show peak efficiency (4-5x speedup)
```
sparknack: Remove these performance comments if they do not reflect actual results.

benchmark/benchmark_sparse_simd.cpp
```diff
@@ -0,0 +1,408 @@
...
+        {"Sparse BM25 (heavy terms, avg=32)", 100000, 5000, 20, 32, SparseMetricType::METRIC_BM25, true},
+
+        // Medium density: posting lists 2-8x SIMD width (64-128)
+        // SIMD should show 2-3x speedup
+        {"Medium IP (random query, avg=128)", 500000, 8000, 25, 128, SparseMetricType::METRIC_IP, false},
+        {"Medium IP (heavy terms, avg=128)", 500000, 8000, 25, 128, SparseMetricType::METRIC_IP, true},
+        {"Medium BM25 (heavy terms, avg=128)", 500000, 8000, 25, 128, SparseMetricType::METRIC_BM25, true},
+
+        // Dense: posting lists 16-32x SIMD width (256-512)
+        // SIMD should show 3-4x speedup
```
sparknack: ditto.

benchmark/benchmark_sparse_simd.cpp
```diff
@@ -0,0 +1,408 @@
...
+        std::normal_distribution<float> len_dist(1.0f, 0.2f);
+        doc_len_ratios.resize(n_docs);
+        for (size_t i = 0; i < n_docs; ++i) {
+            doc_len_ratios[i] = std::max(0.5f, len_dist(rng));
+        }
+    }
+};
+
+// Simple BM25 computer for testing
+struct SimpleBM25Computer {
```
sparknack: remove this struct if it’s no longer in use.

benchmark/benchmark_sparse_simd.cpp
```diff
@@ -0,0 +1,408 @@
...
+        // Ultra-sparse: posting lists shorter than SIMD width (16)
+        // SIMD should be slower due to overhead without amortization
+        {"Ultra-sparse IP (random query, avg=8)", 50000, 2000, 15, 8, SparseMetricType::METRIC_IP, false},
+        {"Ultra-sparse IP (heavy terms, avg=8)", 50000, 2000, 15, 8, SparseMetricType::METRIC_IP, true},
+
+        // Sparse: posting lists around SIMD width (16-32)
+        // SIMD may break even or show modest gains
+        {"Sparse IP (random query, avg=32)", 100000, 5000, 20, 32, SparseMetricType::METRIC_IP, false},
+        {"Sparse IP (heavy terms, avg=32)", 100000, 5000, 20, 32, SparseMetricType::METRIC_IP, true},
+        {"Sparse BM25 (heavy terms, avg=32)", 100000, 5000, 20, 32, SparseMetricType::METRIC_BM25, true},
```
sparknack: Since BM25 SIMD isn’t supported at the moment, will these BM25 benchmarks show any performance comparison results?

src/simd/sparse_simd_avx512.cc
```diff
@@ -0,0 +1,77 @@
...
+
+        // Chunk 0: elements [j, j+16)
+        __m512 vals0 = _mm512_loadu_ps(&doc_vals[j]);
+        __m512i doc_ids0 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j]));
+
+        // Chunk 1: elements [j+16, j+32)
+        __m512 vals1 = _mm512_loadu_ps(&doc_vals[j + SIMD_WIDTH]);
+        __m512i doc_ids1 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j + SIMD_WIDTH]));
+
+        __m512 q_weight_vec = _mm512_set1_ps(q_weight);
```
sparknack: You can move this out of the for loop.

src/simd/sparse_simd_avx512.cc
```diff
@@ -0,0 +1,77 @@
...
+        __m512 current_scores1 = _mm512_i32gather_ps(doc_ids1, scores, sizeof(float));
+        __m512 new_scores1 = _mm512_add_ps(current_scores1, contribution1);
+        _mm512_i32scatter_ps(scores, doc_ids1, new_scores1, sizeof(float));
+    }
+
+    // Handle remaining 16-31 elements
+    for (; j + SIMD_WIDTH <= list_size; j += SIMD_WIDTH) {
+        __m512 vals = _mm512_loadu_ps(&doc_vals[j]);
+        __m512i doc_ids_vec = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j]));
+        __m512 q_weight_vec = _mm512_set1_ps(q_weight);
```
sparknack: ditto.

tests/ut/test_sparse_simd.cc
```diff
@@ -0,0 +1,342 @@
...
+
+        accumulate_posting_list_ip_avx512(test_data.doc_ids.data(), test_data.doc_vals.data(), test_data.doc_ids.size(),
+                                          q_weight, avx512_scores.data());
+
+        for (size_t i = 0; i < n_docs; ++i) {
+            REQUIRE(std::abs(avx512_scores[i]) < tolerance);
+        }
+    }
+
+    SECTION("Negative query weight") {
```
sparknack: negative query weight is not supported in sparse_inverted_index.h. instead, it will automatically be converted to a positive number.

src/simd/sparse_simd_avx512.cc
lyang24: good call q weight is a constant is wasteful to put this in loop

src/simd/sparse_simd_avx512.cc
```diff
@@ -0,0 +1,77 @@
...
+        __m512i doc_ids0 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j]));
+
+        // Chunk 1: elements [j+16, j+32)
+        __m512 vals1 = _mm512_loadu_ps(&doc_vals[j + SIMD_WIDTH]);
+        __m512i doc_ids1 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j + SIMD_WIDTH]));
+
+        __m512 q_weight_vec = _mm512_set1_ps(q_weight);
+
+        // Process chunk 0
+        __m512 contribution0 = _mm512_mul_ps(vals0, q_weight_vec);
```
lyang24: taking another look - seems like we can use fma here to save one instruction

__m512 new_scores0 = _mm512_fmadd_ps(vals0, q_weight_vec, current_scores0);

src/simd/sparse_simd_avx512.cc
```diff
@@ -0,0 +1,83 @@
...
+    for (; j + SIMD_WIDTH <= list_size; j += SIMD_WIDTH) {
+        __m512 vals = _mm512_loadu_ps(&doc_vals[j]);
+        __m512i doc_ids_vec = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j]));
+        __m512 current_scores = _mm512_i32gather_ps(doc_ids_vec, scores, sizeof(float));
+        __m512 new_scores = _mm512_fmadd_ps(vals, q_weight_vec, current_scores);
+        _mm512_i32scatter_ps(scores, doc_ids_vec, new_scores, sizeof(float));
+    }
+
+    // Masked tail: handle remaining 0-15 elements in SIMD mode instead of scalar
+    if (j < list_size) {
```
lyang24, line 73: i padded this instead of scalar

src/simd/sparse_simd.h
lyang24: fixed

.github/workflows/ut.yaml
lyang24: +1 fixed

lyang24:
this is the bench result form c7i.2xlarge on the newest code small improvement but no regressions

| Benchmark                       | Scalar   | SIMD     | Speedup |
|---------------------------------|----------|----------|---------|
| Ultra-sparse (random, avg=8)    | 0.004 ms | 0.003 ms | 1.34×   |
| Ultra-sparse (heavy, avg=8)     | 0.009 ms | 0.008 ms | 1.09×   |
| Sparse (random, avg=32)         | 0.010 ms | 0.007 ms | 1.34×   |
| Sparse (heavy, avg=32)          | 0.044 ms | 0.041 ms | 1.08×   |
| Medium (random, avg=128)        | 0.052 ms | 0.043 ms | 1.19×   |
| Medium (heavy, avg=128)         | 0.404 ms | 0.381 ms | 1.06×   |
| Dense (random, avg=512)         | 0.179 ms | 0.154 ms | 1.17×   |
| Dense (heavy, avg=512)          | 1.765 ms | 1.723 ms | 1.02×   |
| Very Dense (heavy, avg=2048)    | 3.370 ms | 3.032 ms | 1.11×   |
| Real-world (heavy, avg=256)     | 1.299 ms | 1.250 ms | 1.04×   |

alexanderguzhva:
@lyang24 excellent. Thanks a lot for the contribution! 
Let me ask @sparknack one more time to confirm that everything is good, and then we'll merge the PR.

sparknack:
Thanks so much for the continuous improvements! LGTM.

alexanderguzhva:
issue: #1433

PR #1418, opened 2026-01-01
feat: [NCS] Integrate NCS support into DiskANN

ronmarcus:
This commit updates DiskANN to support Near-Compute Storage (NCS), enabling index data to be served from a shared KV storage tier.

Key changes:
- Refactored `PQFlashIndex` to use an abstract `IndexReader` interface, replacing direct file I/O.
- Implemented `NCSReader` to fetch index nodes via the `milvus-common` NCS connector.
- Implemented `FileIndexReader` to maintain backward compatibility with local file storage.
- Added `NcsUpload` to the Index class, enabling the upload of local index files to the NCS tier.
- Updated `DiskANN` loading and searching logic to support key-based data retrieval (`ReadReq`).
- Added Python bindings and unit tests for NCS-enabled DiskANN.

This allows DiskANN to scale independently of local storage constraints by leveraging the NCS tier.

issue: milvus-io/milvus#45178

### DIFF
--- a/cmake/libs/libmilvus-common.cmake
+++ b/cmake/libs/libmilvus-common.cmake
@@ -1,6 +1,6 @@
 
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( MILVUS-COMMON-VERSION b0a5e9b ) 
+set( MILVUS-COMMON-VERSION 55e16501 ) 
 set( GIT_REPOSITORY  "https://github.com/ronmarcus/milvus-common.git" ) 
 
 message(STATUS "milvus-common repo: ${GIT_REPOSITORY}")
--- a/src/index/diskann/diskann.cc
+++ b/src/index/diskann/diskann.cc
@@ -238,11 +238,11 @@ class DiskANNIndexNode : public IndexNode {
             reader->read(reqs);
 
             std::vector<uint32_t> keys;
-            std::vector<milvus::SpanBytes> valueSpans;
+            std::vector<boost::span<uint8_t>> valueSpans;
             for(uint32_t j=0 ; j < batch_size_i ; j++){
                 auto key = i + j;
                 keys.push_back(key);
-                valueSpans.emplace_back(reqs[j].buf, max_node_len);
+                valueSpans.emplace_back(static_cast<uint8_t*>(reqs[j].buf), max_node_len);
             }
 
             auto putResults = connector->multiPut(keys, valueSpans);
--- a/tests/ut/test_diskann.cc
+++ b/tests/ut/test_diskann.cc
@@ -103,11 +103,11 @@ namespace {
  * @param ncs_kind The NCS backend type ("in_memory" or "redis")
  * @return json config with appropriate settings for the backend
  */
-json getNcsExtras(const std::string& ncs_kind) {
+nlohmann::json getNcsExtras(const std::string& ncs_kind) {
     if (ncs_kind == "redis") {
-        return json{{"redis_host", "localhost"}, {"redis_port", 6379}};
+        return nlohmann::json{{"redis_host", "localhost"}, {"redis_port", 6379}};
     }
-    return json::object();
+    return nlohmann::json::object();
 }
 
 /**
@@ -153,7 +153,7 @@ TEST_CASE("InMemoryNcsConnector - sanity", "[NcsTest]") {
         std::vector<uint8_t>(200, 0x22),  // 200 bytes
         std::vector<uint8_t>(300, 0x33)   // 300 bytes
     };
-    std::vector<SpanBytes> valueSpans;
+    std::vector<boost::span<uint8_t>> valueSpans;
     for (auto& value : values) {
         valueSpans.emplace_back(value.data(), value.size());
     }
@@ -167,10 +167,12 @@ TEST_CASE("InMemoryNcsConnector - sanity", "[NcsTest]") {
 
     // Prepare buffers for reading
     std::vector<std::vector<uint8_t>> readBuffers;
-    std::vector<SpanBytes> readSpans;
+    std::vector<boost::span<uint8_t>> readSpans;
     for (const auto& value : values) {
         readBuffers.emplace_back(value.size());
-        readSpans.emplace_back(readBuffers.back().data(), readBuffers.back().size());
+    }
+    for (auto& buf : readBuffers) {
+        readSpans.emplace_back(buf.data(), buf.size());
     }
 
     // Test multiGet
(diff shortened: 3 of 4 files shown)


alexanderguzhva:
@ronmarcus it seems that unit tests are failing. Could you please take a look whenever you have time? :)

ronmarcus:
> @ronmarcus it seems that unit tests are failing. Could you please take a look whenever you have time? :)

@alexanderguzhva thanks for pointing this out.
Fixed in commit 8347129b.

conanfile.py
```diff
@@ -101,6 +101,7 @@ def requirements(self):
         self.requires("glog/0.6.0")
         self.requires("nlohmann_json/3.11.2")
         self.requires("openssl/1.1.1t")
+        self.requires("hiredis/1.2.0")
```
alexanderguzhva: where is this used exactly?

cmake/libs/libmilvus-common.cmake
```diff
@@ -1,7 +1,7 @@
 
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( MILVUS-COMMON-VERSION b6629f7 )
-set( GIT_REPOSITORY  "https://github.com/zilliztech/milvus-common.git" )
+set( MILVUS-COMMON-VERSION b0a5e9b ) 
+set( GIT_REPOSITORY  "https://github.com/ronmarcus/milvus-common.git" )
```
alexanderguzhva, line 4: please fix this one appropriately to `https://github.com/zilliztech/milvus-common.git` or let us know about the proposed change there

conanfile.py
```diff
@@ -111,6 +112,7 @@ def requirements(self):
         self.requires("libcurl/8.2.1")
         self.requires("simde/0.8.2")
         self.requires("xxhash/0.8.3")
+        self.requires("hiredis/1.2.0")
```
alexanderguzhva, line 114: where is this used exactly?

ronmarcus: Hiredis is a client library for the Redis database. It is used by diskann when the Redis implementation of NCS is enabled. Specifically, Hiredis is used in the `RedisNcsConnector` and `RedisNcs` classes in milvus-common, which are used by diskann:
* diskann uses the `IndexReader` abstract class, which has two implementations, one of which is `NCSReader`. This uses the `NcsConnector` abstract class, with `RedisNcsConnector` as one implementation.
* In the unit tests under `test_diskann.cc`, the `Ncs` abstract class is used for NCS bucket management, with `RedisNcs` as one of its implementations.

conanfile.py
ronmarcus: This was a duplicated requirment, I removed it.
I answered below where Hiredis is used.

cmake/libs/libmilvus-common.cmake
ronmarcus: I have three related PRs across three repositories: `milvus-common`, `knowhere`, and `milvus`. The high-level idea behind them is described in the linked GitHub issue.

From the point of view of DiskANN, NCS (Near Compute Storage) is an alternative to a locally attached SSD for storing index data. Specifically, NCS can replace local storage when querying the index; the build process remains unchanged.
 
The proposed change in milvus-common provides the NCS infrastructure, including:
- `Ncs` class for NCS bucket management
- `NcsConnector` class for writing and reading to/from the NCS backend
 
NCS is used by both knowhere (to access the data) and milvus (the coordinator manages NCS). Thus, we added the NCS infrastructure to milvus-common.

tests/ut/test_diskann.cc
```diff
@@ -1393,3 +2004,174 @@ TEST_CASE("Test DiskANN Search Cancellation", "[diskann][cancellation]") {
...
+    // Mode-specific configuration
+    if (test_mode == std::string("connector")) {
+        build_config["max_degree"] = 56;
+        build_config["search_list_size"] = 100;
+        build_config["pq_code_budget_gb"] = 7.000000096013537e-06;
+        build_config["pq_code_budget_gb_ratio"] = 0.125;
+        build_config["search_cache_budget_gb"] = 6.000000212225132e-06;
+        build_config["search_cache_budget_gb_ratio"] = 0.10000000149011612;
+        build_config["build_dram_budget_gb"] = 503.04913330078125;
+        build_config["num_build_thread"] = 80;
```
alexanderguzhva: this triggers problems on our CI, bcz we use a github machine with 4 CPUs only and a proportional amount of RAM (16 GB, I think) :(

ronmarcus: I changed the test to use a typical config

src/index/diskann/diskann.cc
```diff
@@ -343,16 +449,18 @@ GetNecessaryFilenames(const std::string& prefix, const bool need_norm, const boo
                       const bool use_sample_warmup) {
     std::vector<std::string> filenames;
     auto pq_pivots_filename = diskann::get_pq_pivots_filename(prefix);
-    auto disk_index_filename = diskann::get_disk_index_filename(prefix);
+    auto disk_index_metadata_filename = diskann::get_disk_index_metadata_filename(prefix);
```
foxspy, line 452: It looks like this changes the file-naming scheme, which could cause serious compatibility issues. Please consider how the new code will load existing indexes, and how compatibility during upgrades will be ensured. Knowhere provides a version mechanism to guarantee compatibility—please make sure the PR has taken this into account.

src/index/diskann/diskann.cc
```diff
@@ -363,9 +471,8 @@ GetNecessaryFilenames(const std::string& prefix, const bool need_norm, const boo
 std::vector<std::string>
 GetOptionalFilenames(const std::string& prefix) {
     std::vector<std::string> filenames;
-    auto disk_index_filename = diskann::get_disk_index_filename(prefix);
-    filenames.push_back(diskann::get_disk_index_centroids_filename(disk_index_filename));
-    filenames.push_back(diskann::get_disk_index_medoids_filename(disk_index_filename));
+    filenames.push_back(diskann::get_disk_index_centroids_filename(prefix));
```
foxspy, line 474: same compatibility issue

src/index/diskann/diskann.cc
```diff
@@ -589,14 +700,18 @@ DiskANNIndexNode<DataType>::Deserialize(const BinarySet& binset, std::shared_ptr
     // set thread pool
     search_pool_ = ThreadPool::GetGlobalSearchThreadPool();
 
-    // load diskann pq code and meta info
-    std::shared_ptr<AlignedFileReader> reader = nullptr;
+    const milvus::NcsDescriptor* descriptor = nullptr;
```
foxspy, line 703: How do we guarantee consistency between upload and load? It looks like this relies on configuration, but if the configuration is wrong or missing, it could cause a serious load failure. We need to validate this consistency before loading.

PR #1406, opened 2025-12-18
Add sq4 for HNSW

CLiqing:
Add efficient SQ4 for HNSW

kind/enhancement

PR #1419, opened 2026-01-06
Fix SQ8's bug while SQ4U merged.

CLiqing:
Fix bug caused by https://github.com/zilliztech/knowhere/pull/1406


foxspy:
issue:  #1416 
/kind improvement

PR #1429, opened 2026-01-12
GPU support for building DISKANN based indexes. 

liorf95:
Aborted for now.

### DIFF
--- a/src/common/cuvs/integration/cuvs_knowhere_index.cuh
+++ b/src/common/cuvs/integration/cuvs_knowhere_index.cuh
@@ -430,7 +430,8 @@ struct cuvs_knowhere_index<IndexKind, DataType>::impl {
             auto device_data = raft::make_device_matrix<data_type, input_indexing_type>(res, row_count, feature_count);
             auto device_data_view = device_data.view();
             raft::copy(res, device_data_view, host_data);
-            raft::linalg::row_normalize<raft::linalg::NormType::L2Norm>(res, raft::make_const_mdspan(device_data_view), device_data_view);
+            raft::linalg::row_normalize<raft::linalg::NormType::L2Norm>(res, raft::make_const_mdspan(device_data_view),
+                                                                        device_data_view);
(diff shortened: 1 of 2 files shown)


foxspy:
@liorf95 Thank you for your contribution. Please rebase onto master so that the PR contains only your changes.

thirdparty/DiskANN/src/diskann_gpu.cpp
```diff
@@ -0,0 +1,251 @@
...
+		throw std::runtime_error("Failed to open file: " + fname);
+	}
+
+	// Read header
+	uint32_t N = 0;
+	uint32_t dim = 0;
+	datafile.read(reinterpret_cast<char*>(&N), sizeof(uint32_t));
+	datafile.read(reinterpret_cast<char*>(&dim), sizeof(uint32_t));
+
+	std::cout << "Read in file - N: " << N << ", dim: " << dim << std::endl;
```
foxspy: don’t use std::cout; use LOG_KNOWHERE_INFO_ instead

liorf95: Fixed.

cmake/libs/libcuvs.cmake
```diff
@@ -16,7 +16,7 @@
 add_definitions(-DKNOWHERE_WITH_CUVS)
 set(CUVS_VERSION "${RAPIDS_VERSION}")
 set(CUVS_FORK "rapidsai")
-set(CUVS_PINNED_TAG "branch-25.06")
+set(CUVS_PINNED_TAG "v25.10.00")
```
alexanderguzhva, line 19: may I ask why it's not v25.12 ?

liorf95: We have done many system tests with this cuvs version and confirmed proper support for diskann based indexes, while the latest one was just recently released. 
Thus, we would suggest starting with version 25.10 and upgrading at a later stage once we have better confidence in this newer version.

PR #1430, opened 2026-01-12
[first stage] Upgrade faiss to 1.13.2+ (commit 1aea3e934745f37)

alexanderguzhva:
This is the first patch of the following series. I expect it to be committed ASAP, because it allows starting the work on integrating some other Faiss parts, such as SVS, cuVS, etc. into Milvus. And these integration processes can be started independently from a fork-related refactoring process.

The code has been reorganized in the following way:
* `/thirdparty/faiss` has been replaced with a clone of the baseline Faiss version, 1.13.2+, commit 1aea3e934745f37e8d60cfff6c5900b175c42709
* all knowhere-specific Faiss fork pieces were moved to `/thirdparty/faiss/faiss/cppcontrib/knowhere` directory under a `C++` namespace `faiss::cppcontrib::knowhere`. 
* all knowhere-specific faiss integration code (such as `/src/simd/`) has been altered to use a `C++` namespace `faiss::cppcontrib::knowhere` 

It is expected to gradually remove duplicate components from a knowhere-fork a replace them with the baseline faiss components. If this is not possible, then a new component must be created and put into `cppcontrib/knowhere` directory.

Additional patches with the refactoring of the fork are expected.


alexanderguzhva:
issue: #1439

foxspy:
@alexanderguzhva  After this PR, does knowhere still need to bear the maintenance cost of keeping code in sync with Faiss? It currently looks like that’s the case.

alexanderguzhva:
@foxspy yes

foxspy:
> @foxspy yes

Could we maintain two versions of Faiss: one that directly references https://github.com/facebookresearch/faiss, and another one under thirdparty/faiss? The former would be integrated into Knowhere without any modifications, so we can avoid the cost of keeping it in sync.

alexanderguzhva:
@foxspy we cannot, because we still need to modify faiss a bit. For example, I needed to add extra metric types, which is impossible to do if the code is just referenced

PR #1441, opened 2026-01-17
using binary search for DAAT search

lyang24:


lyang24:
DAAT is more like a streaming algo this is a bad idea

PR #1447, opened 2026-01-20
enhance: add claude md for knowhere

foxspy:
issue: #1446

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,5 +70,3 @@ CMakeUserPresets.json
 # python
 .eggs
 *.egg-info
-
-CLAUDE.md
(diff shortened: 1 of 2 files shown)


foxspy:
> @foxspy looks good, but most `conan` commands are a kind of a `reference` ones. They require additional options if you want to compile using a different compiler, ARM SVE support, debugging version, asan, etc. Also, I'd upgrade conan to 1.66.0, because I had some problems somewhere on conan 1.61.0, which were gone after then upgrade.

These are some general project prompts. For personalized commands, you can add prompts to ${project_dir}/.claude/CLAUDE.local.md. This can serve as the default prompt for working with Claude, reducing the amount of context that needs to be passed. Also, CLAUDE.md should be continuously updated as the project evolves—any unreasonable parts can be gradually revised, modified, or replaced during development.

PR #1454, opened 2026-02-03
propose an query time improved version of max score

lyang24:
v2 focuses on optimizations at query time the inverted index structure is unchanged:
for ip path: we sequential access per posting list and added simd instructions.
for bm25 path: 
1.  Sorts terms by IDF-weighted priority: log(N/df) * max_score * q_weight instead of just max_score * q_weight, This puts rare, high-discriminative terms first as essential cursors.  
2.  if the cursor is already at or past the candidate, skip the seek(). (applied to both v1 and v2 impl in the experiements)

# Dataset & BM25 Preparation

All BM25 experiments use the same preprocessing pipeline and CSR format.

### Preprocessing pipeline

1. **Tokenization**
   - Lowercasing
   - English stopword removal  
   - Tokens mapped to integer IDs

3. **Term Frequency (TF) counting**
   - For each document, count occurrences of each token ID.

4. **CSR construction**
   - Rows = documents  
   - Columns = vocab IDs  
   - Values = raw TF counts (float32)

5. **Queries**
   - Tokenized the same way  
   - Query tokens remapped to the *corpus vocabulary IDs*

### CSR binary format

All runs: `top-10`, single-node EC2 r6i.2xlarge (8 vCPU, 64 GB RAM).

---

## MSMARCO Full (8.8M docs, 6,980 queries, BM25)

| Algorithm | Search (ms) | QPS | Peak RAM | Speedup |
|----------|------------:|----:|---------:|--------:|
| V1 | 92,303 | 75.6 | 4.5 GB | 1.0× |
| V2 | 38,240 | 182.5 | 4.5 GB | **2.4×** |

---

## SPLADE Full (8.8M docs, 6,980 queries, IP)

| Algorithm | Search (ms) | QPS | Peak RAM | Speedup |
|----------|------------:|----:|---------:|--------:|
| V1 | 1,124,693 | 6.2 | 19.0 GB | 1.0× |
| V2 | 627,796 | 11.1 | 19.1 GB | **1.8×** |

---

## HotpotQA (5.2M docs, 7,405 queries, BM25)

| Algorithm | Search (ms) | QPS | Recall@10 | Peak RAM | Speedup |
|----------|------------:|----:|----------:|---------:|--------:|
| V1 | 8,555 | 866 | 98.78% | 0.7 GB | 1.0× |
| V2 | 5,418 | 1,367 | 98.78% | 0.7 GB | **1.6×** |

---

## NQ (2.7M docs, 3,452 queries, BM25)

| Algorithm | Search (ms) | QPS | Recall@10 | Peak RAM | Speedup |
|----------|------------:|----:|----------:|---------:|--------:|
| V1 | 25,715 | 134 | 100% | 2.1 GB | 1.0× |
| V2 | 10,742 | 321 | 100% | 2.1 GB | **2.4×** |

---

### DIFF
--- a/src/simd/sparse_simd.h
+++ b/src/simd/sparse_simd.h
@@ -56,7 +56,7 @@ simd_seek_dispatch(const uint32_t* __restrict__ ids, size_t size, size_t start_p
 #if defined(__x86_64__) || defined(_M_X64)
     // Only use SIMD for larger seeks where setup overhead is amortized
     if (remaining >= SIMD_SEEK_THRESHOLD) {
-        if (faiss::InstructionSet::GetInstance().AVX512F()) {
+        if (faiss::cppcontrib::knowhere::InstructionSet::GetInstance().AVX512F()) {
             return simd_seek_avx512_impl(ids, size, start_pos, target);
         }
     }
@@ -92,7 +92,7 @@ accumulate_window_ip_dispatch(const uint32_t* doc_ids, const QType* doc_vals, si
                               float q_weight, float* scores, uint32_t window_start) {
 #if defined(__x86_64__) || defined(_M_X64)
     if constexpr (std::is_same_v<QType, float>) {
-        if (faiss::InstructionSet::GetInstance().AVX512F()) {
+        if (faiss::cppcontrib::knowhere::InstructionSet::GetInstance().AVX512F()) {
             accumulate_window_ip_avx512(doc_ids, doc_vals, list_start, list_end, q_weight, scores, window_start);
             return;
         }
@@ -122,7 +122,7 @@ extract_candidates_scalar(const float* scores, size_t window_size, float thresho
 inline size_t
 extract_candidates_dispatch(const float* scores, size_t window_size, float threshold, uint32_t* candidates) {
 #if defined(__x86_64__) || defined(_M_X64)
-    if (faiss::InstructionSet::GetInstance().AVX512F()) {
+    if (faiss::cppcontrib::knowhere::InstructionSet::GetInstance().AVX512F()) {
         return extract_candidates_avx512(scores, window_size, threshold, candidates);
     }
 #endif


src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +80,79 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+        // Load doc IDs and values for chunk 1
+        __m512i ids1 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(&doc_ids[j + SIMD_WIDTH]));
+        __m512 vals1 = _mm512_loadu_ps(&doc_vals[j + SIMD_WIDTH]);
+
+        // Convert to window-relative indices
+        __m512i local_ids0 = _mm512_sub_epi32(ids0, window_start_vec);
+        __m512i local_ids1 = _mm512_sub_epi32(ids1, window_start_vec);
+
+        // Gather-FMA-Scatter for chunk 0
+        __m512 current0 = _mm512_i32gather_ps(local_ids0, scores, sizeof(float));
```
alexanderguzhva: the baseline scalar version will work correctly if there are duplicate values among `local_ids`, while this version may produce incorrect results in this case. Is the uniqueness of `local_ids` values guaranteed / verified anywhere? 
This is not a blocker, but I'm mentioning Just in case.

alexanderguzhva:
overall, looks good

src/simd/sparse_simd_avx512.cc
lyang24: After a deeper look i think this is correct the simd loop cannot handle dups well - wrote a comment on the invariant in second commit about this, 

instead of handling this at simd level check for dups and fall back to scalar loop maybe expensive impact 
```
  // In SIMD loop, check for conflicts using AVX512CD                                                                                                                                                                                                
  __m512i conflicts = _mm512_conflict_epi32(local_ids);                                                                                                                                                                                              
  if (_mm512_test_epi32_mask(conflicts, conflicts)) {                                                                                                                                                                                                
      // Duplicates found - fall back to scalar for this batch                                                                                                                                                                                       
      for (int i = 0; i < 16; i++) { scores[local_ids[i]] += ...; }                                                                                                                                                                                  
  } else {                                                                                                                                                                                                                                           
      // Safe to use scatter                                                                                                                                                                                                                         
      _mm512_i32scatter_ps(...);                                                                                                                                                                                                                     
  } 
```

maybe  accept the limitation and write clearly about this in our documents is a better solution?

lyang24: actually i have a local commit that does the conflict handling and scalar fall back the result is still pretty good on msmacro bench probably cpu learned its a unlikely path, should i include the change?

https://github.com/zilliztech/knowhere/pull/1457/changes/56aa8e0b8d997a34d99dfd763fed9a41a82671d1

image

lyang24:
realized we can improve this further by adding SIMD for 16 scores at a time given the window is 64k that cut comparison instruction to 4,096.

image

src/simd/sparse_simd_avx512.cc
alexanderguzhva: I think I would keep the code that performs the conflict check, but make it disabled. Something like
```C++
    // by default, the code implies that there are no duplicate elements,
    //   and it does not work correctly by design if there are any.
    // Please change the `check_conflicts` variable to `true` to enable
    //   conflict checks.
    bool check_conflicts = false;

    bool any_conflicts = false;
    if (check_conflicts) { 
        any_conflicts = check_simd_mask(_mm512_conflict_epi32(...) ... );
    }

    if (any_conflicts) {
        // scalar version
    } else {
        // gather-scatter version
    }
``` 
In theory, you may even expose this `bool check_conflicts` flag to an external world, so that the behavior could be controlled. We all expect that there will be no duplicated, but who knows..

lyang24: thank you added this in the 4th commit

src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +113,168 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+        if constexpr (check_conflicts) {
+            if (!has_no_conflicts(local_ids0)) {
+                scalar_accumulate_window(doc_ids, doc_vals, j, j + SIMD_WIDTH, q_weight, scores, window_start);
+            } else {
+                __m512 current0 = _mm512_i32gather_ps(local_ids0, scores, sizeof(float));
+                __m512 new0 = _mm512_fmadd_ps(vals0, q_weight_vec, current0);
+                _mm512_i32scatter_ps(scores, local_ids0, new0, sizeof(float));
+            }
+        } else {
+            __m512 current0 = _mm512_i32gather_ps(local_ids0, scores, sizeof(float));
```
alexanderguzhva, line 169: there are duplicate lines 170-172 and 175-177. Please merge them into a single block. Fyi, `if constexpr` check for check_conflicts is not really needed, bcz modern compilers are smart enough, so you may use a plain `if`

src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +113,168 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+            if (!has_no_conflicts(local_ids1)) {
+                scalar_accumulate_window(doc_ids, doc_vals, j + SIMD_WIDTH, j + 2 * SIMD_WIDTH, q_weight, scores,
+                                         window_start);
+            } else {
+                __m512 current1 = _mm512_i32gather_ps(local_ids1, scores, sizeof(float));
+                __m512 new1 = _mm512_fmadd_ps(vals1, q_weight_vec, current1);
+                _mm512_i32scatter_ps(scores, local_ids1, new1, sizeof(float));
+            }
+        } else {
+            __m512 current1 = _mm512_i32gather_ps(local_ids1, scores, sizeof(float));
```
alexanderguzhva, line 179: same

src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +113,168 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+        } else {
+            __m512 current = _mm512_i32gather_ps(local_ids, scores, sizeof(float));
+            __m512 new_scores = _mm512_fmadd_ps(vals, q_weight_vec, current);
+            _mm512_i32scatter_ps(scores, local_ids, new_scores, sizeof(float));
+        }
+    }
+
+    // Masked tail - use scalar for simplicity (small number of elements)
+    if (j < list_end) {
+        scalar_accumulate_window(doc_ids, doc_vals, j, list_end, q_weight, scores, window_start);
```
alexanderguzhva: well, I leave it up to you, but I'd leave SIMD masked version here

src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +113,168 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+    size_t i = 0;
+
+    // Main SIMD loop - process 16 scores at a time
+    for (; i + SIMD_WIDTH <= window_size; i += SIMD_WIDTH) {
+        __m512 scores_vec = _mm512_loadu_ps(&scores[i]);
+
+        // Compare: mask bit set where score > threshold
+        __mmask16 mask = _mm512_cmp_ps_mask(scores_vec, threshold_vec, _CMP_GT_OQ);
+
+        if (mask != 0) {
```
alexanderguzhva: if the probability of `mask !=0` is very high, then the check is not really needed, because the code within `if` construction handles `mask == 0` case, if I'm not mistaken.

alexanderguzhva:
@sparknack could you please take a look as well?

lyang24:
> @sparknack could you please take a look as well?

there is a regression on BM25 on msmacro full data set, will make some adjustments

alexanderguzhva:
@lyang24 what is the hardware that you're testing on?

lyang24:
> @lyang24 what is the hardware that you're testing on?

i got data from https://storage.googleapis.com/ann-challenge-sparse-vectors/csr

c7i.2xlarge for 100k dataset and 1 mill dataset
c7i.8xlarge for the 8.8m dataset

src/simd/sparse_simd_avx512.cc
lyang24: nice fixed with masked tail

src/simd/sparse_simd_avx512.cc
lyang24: fixed

src/simd/sparse_simd_avx512.cc
lyang24: fixed

src/simd/sparse_simd_avx512.cc
lyang24: that is correct i removed the if block

lyang24:
> @sparknack could you please take a look as well?

I have addressed the comments. Previously i was testing bm25 on spade data which do not make sense, since then i have used bm25s lib to tokenize and remove stopwords then build the id based sparse vectors for bm25 dataset and run benches across MsMacro, nq, hotpotQA. I wonder if zilliz team can help me validate the results cc @alexanderguzhva @sparknack .

the changes are query time only
ip -> sequential access + smid
bm25 -> Sorts terms by IDF-weighted priority: log(N/df) * max_score * q_weight instead of just max_score * q_weight

if zilliz validate the results is it better to point this pr to the original maxscore instead? there is no breaking changes

alexanderguzhva:
@lyang24 well, it looks good to me, from a technical point of view. I'm waiting for a decision from @sparknack

sparknack:
If the performance is consistently better and there are no changes to the index file format, can we just replace DAAT_MAXSCORE with the v2 implementation?

src/index/sparse/sparse_inverted_index.h
```diff
@@ -1091,6 +1171,23 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+        for (auto q_dim : q_vec) {
+            auto& plist_ids = inverted_index_ids_spans_[q_dim.first];
+            auto& plist_vals = inverted_index_vals_spans_[q_dim.first];
+            cursors.emplace_back(plist_ids, plist_vals, n_rows_internal_,
+                                 max_score_in_dim_spans_[q_dim.first] * q_dim.second * dim_max_score_ratio,
+                                 q_dim.second, filter);
+        }
+        return cursors;
+    }
+
```
sparknack, line 1190: remove dead `SIMDCursor` and `make_simd_cursors` code

src/simd/sparse_seek_avx512.cc
```diff
@@ -0,0 +1,69 @@
...
+    // Broadcast target to all 16 lanes
+    __m512i target_vec = _mm512_set1_epi32(static_cast<int32_t>(target));
+
+    // SIMD loop: process 16 elements at a time
+    while (pos + AVX512_WIDTH <= size) {
+        __m512i id_vec = _mm512_load_si512(reinterpret_cast<const __m512i*>(&ids[pos]));
+
+        // AVX512 has native >= comparison
+        // _mm512_cmpge_epi32_mask returns a 16-bit mask where bit i is set if id[i] >= target
+        __mmask16 ge_mask = _mm512_cmpge_epi32_mask(id_vec, target_vec);
```
sparknack, line 48: `epi32` is for `int32_t`, should be `epu32`?

src/index/sparse/sparse_inverted_index.h
```diff
@@ -1261,6 +1484,118 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+        float bound_sum = 0.0f;
+        for (size_t i = posting_lists.size(); i > 0; --i) {
+            bound_sum += posting_lists[i - 1].max_score;
+            upper_bounds[i - 1] = bound_sum;
+        }
+
+        float threshold = heap.full() ? heap.top().val : 0;
+
+        // Allocate score buffer for all documents
+        std::vector<float> scores(n_rows_internal_, 0.0f);
```
sparknack, line 1532: `O(n)` memory usage is too high. in high-concurrency scenarios, this will lead to severe memory pressure and reduced cache efficiency.

src/index/sparse/sparse_inverted_index.h
```diff
@@ -1261,6 +1484,118 @@ class InvertedIndex : public BaseInvertedIndex<DType> {
...
+        float effective_threshold = std::max(0.0f, threshold - ne_upper_bound);
+        size_t num_candidates = sparse::extract_candidates_dispatch(scores.data(), n_rows_internal_,
+                                                                    effective_threshold, candidate_indices.data());
+
+        // Process extracted candidates
+        for (size_t c = 0; c < num_candidates; ++c) {
+            table_t doc_id = candidate_indices[c];
+            float score = scores[doc_id];
+
+            if (!filter.empty() && filter.test(doc_id))
```
sparknack, line 1568: should consider the filtering volume. When the bitset filtering rate is high, most of the SIMD computations will be wasted.

lyang24:
> If the performance is consistently better and there are no changes to the index file format, can we just replace DAAT_MAXSCORE with the v2 implementation?

yes i will break the changes down by ip smid, bm25 query time sort, and reduce seek and then address the comments in future prs.

PR #1457, opened 2026-02-04
Feature/conflict detection fallback

lyang24:


alexanderguzhva:
@lyang24 is it a test-related branch? According to my experience, a conflict detection code using Intel `AVX512CD` is VERY expensive. An alternative is to use a dedicated `vp2intersect` instructions on `Zen5`, which is too hardware constraining.

lyang24:
> @lyang24 is it a test-related branch? According to my experience, a conflict detection code using Intel `AVX512CD` is VERY expensive. An alternative is to use a dedicated `vp2intersect` instructions on `Zen5`, which is too hardware constraining.

yes this is a test branch, thank you for the valuable insight i think at this point the best approach is to document the limit of the v2 maxscore and ask users to send valid input.

Issue #1422, opened 2026-01-07
feat:  GPU support for building DISKANN based indexes.


liorf95:
The Index build process for large collections can take a long time and may consume 100% CPU; thus, reducing build duration time and freeing up CPU is crucial when large datasets are supported.
Although KNOWHERE has already introduced a GPU version for common indices such as IVF, IVF_PQ and Cagra, it does not include support for DISKANN based indices family (e.g., DISKANN, AISAQ).

While adding support for DISKANN based indices, we also want to introduce a new approach for GPU usage, which is more flexible and practical:
The index type will remain the same. However, it will be built with GPU assistance using the NVIDIA cuVS package if the data node server platform has a GPU.
This way, any nodes in the cluster can build indices for the index segments, not only those with a GPU.
In addition, the user won't need any more various configuration types as he still will be able to use the relevant existing one (e.g, - DISKANN or AISAQ).

When the GPU is available, it is used to build the Vamana graph, K-means calculations as part of PQ compression, sharding and AISAQ entry points.

For an Index built with segments that don’t fit in GPU memory, we will use the DISKANN Sharding algorithm, building the index with the GPU in parts
According to the available GPU memory.

alexanderguzhva:
@liorf95 please be aware that we're having an undergoing work of migrating our GPU code from using cuVS directly into using it for Faiss. Will it interfere with your plans? Of course, we can discuss this in more details

gland1:
As we link diskann lib directly with cuvs , not with knowhere code, I dont see a problem. Unless you have a problem with linking diskann with cuvs. The current code we'll provide doesnt use knowhere code for GPU access, it uses only APIs from cuvs.

alexanderguzhva:
@liorf95 Is the PR still available? we were about to review & test it

liorf95:
> [@liorf95](https://github.com/liorf95) Is the PR still available? we were about to review & test it

Currently, it is on hold - I was requested to conduct an internal review before uploading it publicly.
Sorry for the inconvenience about that.

PR #1460, opened 2026-02-08
feat:  GPU support for building DISKANN based indexes.

liorf95:
See issue #1422.


tests/ut/test_diskann.cc
```diff
@@ -64,7 +65,12 @@ constexpr uint32_t kNumRows = 1000;
 constexpr uint32_t kNumQueries = 10;
 constexpr uint32_t kDim = 128;
 constexpr uint32_t kLargeDim = 256;
-constexpr uint32_t kK = 10;
+constexpr uint32_t kK = 64;
+#ifdef KNOWHERE_WITH_CUVS
+constexpr uint32_t defaultMaxDegree = 64;
```
alexanderguzhva, line 71: why is there a difference for the value of `defaultMaxDegree` for CUVS and non-CUVS versions for the testing? Any particular reasons?

tests/ut/test_diskann.cc
```diff
@@ -64,7 +65,12 @@ constexpr uint32_t kNumRows = 1000;
 constexpr uint32_t kNumQueries = 10;
 constexpr uint32_t kDim = 128;
 constexpr uint32_t kLargeDim = 256;
-constexpr uint32_t kK = 10;
+constexpr uint32_t kK = 64;
+#ifdef KNOWHERE_WITH_CUVS
```
alexanderguzhva, line 69: what are the requirements for the GPU memory for a cuVS-based version of this test? It would be nice to have it as a comment in the beginning of the file, including a minimum sifficient GPU, if appropriate. Something like `This test is expected to run on a GPU with at least 2 GB of RAM, such as NVIDIA A10 or NVIDIA T4`

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -477,7 +480,19 @@ namespace diskann {
       double ram_budget, std::string mem_index_path, std::string medoids_file,
       std::string centroids_file) {
     size_t base_num, base_dim;
+    uint32_t shard_r = 2*R/3;
     diskann::get_bin_metadata(base_file, base_num, base_dim);
+#ifdef KNOWHERE_WITH_CUVS
+    raft::device_resources dev_resources;
+    if(compareMetric == diskann::L2 && is_gpu_available()) {
```
alexanderguzhva, line 492: and what is going to happen `if (compareMetric != diskann::L2)`?

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -477,7 +480,19 @@ namespace diskann {
...
+      size_t gpu_free_mem, gpu_total_mem;
+      gpu_get_mem_info(dev_resources, gpu_free_mem, gpu_total_mem);
+      LOG_KNOWHERE_INFO_ << "GPU has " <<  gpu_free_mem/(1024*1024*1024L) <<
+              " Gib free memory out of " <<  gpu_total_mem/(1024*1024*1024L) << " Gib total";
+      ram_budget = std::min<double>(ram_budget,(double)0.9*(gpu_free_mem/(1024*1024*1024)));
+      shard_r = std::max<uint32_t>((uint32_t)32,(uint32_t)R/2);
+    }
```
alexanderguzhva, line 502:
```C++
} else {
    // add to log why GPU is not going to be used
}
```

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -499,9 +514,42 @@ namespace diskann {
...
       std::unique_ptr<diskann::Index<T>> _pvamanaIndex =
           std::unique_ptr<diskann::Index<T>>(new diskann::Index<T>(
-              compareMetric, ip_prepared, base_dim, base_num, false, false));
-      _pvamanaIndex->build(base_file.c_str(), base_num, paras);
-      _pvamanaIndex->save(mem_index_path.c_str(), true);
+          compareMetric, ip_prepared, base_dim, base_num, false, false));
+      bool built_with_gpu=false;
+#ifdef KNOWHERE_WITH_CUVS
+      //currently cuvs vamana build only supports L2Expanded metric
+      if (compareMetric == diskann::L2 && is_gpu_available() &&
```
alexanderguzhva, line 529: this and the following calls imply that if the metric is right, the library is built with `#define KNOWHERE_WITH_CUVS` and `is_gpu_available()`, then the code will ALWAYS be built using a GPU.
Is it possible to introduce a configurable boolean flag that indicates whether a GPU is expected to be used by a user? Say, the default value is `true`. At least, this would be helpful for running unit tests, so that it would be possible to validate both GPU and non-GPU branches of the code on the same machine without tricks.

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -499,9 +514,42 @@ namespace diskann {
...
+          vamana_build_and_write<uint8_t>(dev_resources,
+                                           raft::make_const_mdspan(dataset.view()),
+                                           mem_index_path,
+                                           R,
+                                           L,
+                                           0.06,
+                                           1);
+        }
+        built_with_gpu=true;
+      }
```
alexanderguzhva, line 556:
```C++
} else {
     // add to log why GPU is not going to be used
}
```
Same comment to other possible `else` branches, if appropriate.

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -1762,7 +1841,22 @@ template<typename T>
       LOG(ERROR) << "Not building index. Please provide more RAM budget";
       return -1;
     }
-
+#ifdef KNOWHERE_WITH_CUVS
+    if(is_gpu_available()) {
+      if (R != 32 && R != 64 && R != 128) {
+        LOG_KNOWHERE_ERROR_ << "Invalid R value for cuvs - should be only 32 or 64 or 128";
```
alexanderguzhva:
```C++
LOG_KNOWHERE_ERROR_ << "Invalid R value (" << R << ") for cuvs - should be only 32 or 64 or 128";
```
Same comment applies for the following lines as well

thirdparty/DiskANN/include/diskann/diskann_gpu.h
```diff
@@ -0,0 +1,96 @@
...
+void vamana_build_and_write(raft::device_resources const& dev_resources,
+                            raft::device_matrix_view<const T, uint32_t> dataset,
+                            std::string out_fname,
+                            int degree,
+                            int visited_size,
+                            float max_fraction,
+                            int iters);
+bool is_gpu_available();
+
+void kmeans_gpu(
```
alexanderguzhva, line 61: personally, I would prefer `FAISS` based names here. So, that would be `train` (`kmeans_gpu`) and `assign` (`predict_gpu`).
Alternatively, please add comments on what these functions do, because it looks confusing for those who are not aware.

thirdparty/DiskANN/src/partition_and_pq.cpp
```diff
@@ -1061,16 +1233,33 @@ int partition_calc_kmeans(const std::string &data_file, const std::string &outpu
...
-    LOG_KNOWHERE_INFO_ << "Running LLoyds " << k << " clusters...";
-    kmeans::run_lloyds(train_data_float.get(), num_train, train_dim, centroids.data(), k, 10, nullptr, nullptr);
+    bool used_gpu = false;
+#ifdef KNOWHERE_WITH_CUVS
+    // GPU path
+    if(is_gpu_available()) {
+      raft::resources res;
+      LOG_KNOWHERE_INFO_ << "Running k-means with " << k << " clusters...using GPU!";
+      kmeans_gpu(res,train_data_float.get(), num_train, train_dim,
+              k, 10, centroids.data());
```
alexanderguzhva: would it be possible to extract `10` (number of iterations?) as a constexpr value nearly, please? Same for the CPU branch

thirdparty/DiskANN/include/diskann/diskann_gpu.h
```diff
@@ -0,0 +1,96 @@
...
+        size_t n_samples,
+        size_t dim,
+		size_t k,
+        const T* queries,
+        size_t n_queries,
+		int64_t* h_labels);
+
+void gpu_get_mem_info(raft::resources &dev_resources, size_t &gpu_free_mem,size_t &gpu_total_mem);
+
+#endif
```
alexanderguzhva, line 93: what about selecting the right GPU to deal with? It is expected to be done by a caller using `raft::resources` object, correct?

alexanderguzhva:
issue: #1422

tests/ut/test_diskann.cc
liorf95: The default value of max_degree for diskann based indexes is 56, so we did not want to modify UT for those basic scenarios.
However,  cuVS related limitations demand max_degree to be a power of 2 (e.g., 32, 64, 128, etc.)

thirdparty/DiskANN/src/aux_utils.cpp
liorf95: Currently, the cuvs Vamana build only supports the L2Expanded metric, so that the index will be built with cpu as usual.

thirdparty/DiskANN/src/aux_utils.cpp
gland1: > this and the following calls imply that if the metric is right, the library is built with `#define KNOWHERE_WITH_CUVS` and `is_gpu_available()`, then the code will ALWAYS be built using a GPU. Is it possible to introduce a configurable boolean flag that indicates whether a GPU is expected to be used by a user? Say, the default value is `true`. At least, this would be helpful for running unit tests, so that it would be possible to validate both GPU and non-GPU branches of the code on the same machine without tricks.

you can already control that with env varibale NVIDIA_VISIBLE_DEVICES - set it to none and the index will be built with CPU

thirdparty/DiskANN/include/diskann/diskann_gpu.h
gland1: > what about selecting the right GPU to deal with? It is expected to be done by a caller using `raft::resources` object, correct?

The current implementation doesnt handle multi GPU logic.IT is assumed that each process use one GPU. however with multi GPU platform in k8s with milvus you can assign each datanode a different GPU exclusively and accelerate the build further.

thirdparty/DiskANN/include/diskann/diskann_gpu.h
liorf95: We will add comments on what these functions do.

thirdparty/DiskANN/src/diskann_gpu.cpp
```diff
@@ -0,0 +1,284 @@
...
+			raft::make_const_mdspan(dataset.view()));
+
+	auto neighbors = raft::make_device_matrix<int64_t, int64_t>(dev_resources,
+			n_queries, k);
+	auto distances = raft::make_device_matrix<T, int64_t>(dev_resources, n_queries, k);
+
+	brute_force::search(dev_resources, search_params, index,
+			raft::make_const_mdspan(queries_matrix.view()), neighbors.view(), distances.view());
+	// Copy back
+	raft::copy(h_labels, neighbors.data_handle(), n_queries * k,
```
marcelo-cjl: same as raft::copy in read_bin_dataset above

thirdparty/DiskANN/src/partition_and_pq.cpp
```diff
@@ -539,6 +578,10 @@ int generate_pq_data_from_pivots(const std::string data_file,
           }));
     }
     knowhere::WaitAllSuccess(futures);
+#ifdef KNOWHERE_WITH_CUVS
+    // GPU path
+    raft::resources res;
```
marcelo-cjl: raft::resources is not thread-safe. It is created outside and then shared across multiple thread pool. Is there race here? Please correct me if I've misunderstood.

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -1762,7 +1850,24 @@ template<typename T>
...
+        return -1;
+      }
+      if (L != 32 && L != 64 && L != 128 && L != 256) {
```
hhy3: what is the specific reason to choose the seven magic numbers? can it be more flexible?

gland1: with cuvs 25.10 which is the version used, cuvs has validation for visted size which must be a power of 2.
This list is reasnoble considering this constraint as we see that higher L than 256 doesnt benfit too much

hhy3: maybe restrict it to 2^k with a larger boundary like L \in [32, 2048] instead of some magic number is better. maybe some larger or harder datasets or in higher recall requirement need higher parameter :)

alexanderguzhva:
issue: #1422 
/kind improvement

thirdparty/DiskANN/src/diskann_gpu.cpp
```diff
@@ -0,0 +1,284 @@
...
+	datafile.close();
+
+	// Create device matrix
+	auto dataset = raft::make_device_matrix<T, idxT>(dev_resources, N, dim);
+
+	// Copy to device using raft::copy
+	raft::copy(dataset.data_handle(), pinned_data, total,
+			raft::resource::get_cuda_stream(dev_resources));
+
+	cudaFreeHost(pinned_data);
```
liliu-z, line 89: `raft::copy` is async on the stream. `cudaFreeHost` here frees pinned memory while DMA is still reading from it — use-after-free. `kmeans_gpu` and `brute_force_gpu` have the same issue (async copy to host, then immediate return without sync). `predict_gpu` in this same file does it correctly. Add `raft::resource::sync_stream(dev_resources)` before freeing/returning in all three functions.

thirdparty/DiskANN/src/partition_and_pq.cpp
liliu-z, line 583: `raft::resources` is created once in the outer scope and captured by reference into multiple thread pool tasks. It holds a CUDA stream, cuBLAS handle, and workspace allocator — none are thread-safe. Concurrent `predict_gpu` calls will interleave operations on the same stream and corrupt allocator state. Either serialize the GPU path (like `generate_pq_pivots` already does for its GPU loop), or create a separate `raft::resources` per task.

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -1762,7 +1855,25 @@ template<typename T>
       LOG(ERROR) << "Not building index. Please provide more RAM budget";
       return -1;
     }
-
+#ifdef KNOWHERE_WITH_CUVS
+    if(is_gpu_available()) {
+      const int* deg_size = std::find(std::begin(DEGREE_SIZES), std::end(DEGREE_SIZES), R);
+      if (deg_size == std::end(DEGREE_SIZES)) {
+        LOG_KNOWHERE_ERROR_ << "Invalid R value for cuvs - should be power of 2 and maximum 256";
+        return -1;
```
liliu-z, line 1863: `R=56` (a common DiskANN default) will `return -1` here if a GPU is present. GPU acceleration is an optimization — it shouldn't break existing configs. Consider falling back to CPU build with a warning instead of failing, e.g. set a `use_gpu = false` flag when params don't meet cuVS constraints.

tests/ut/test_diskann.cc
```diff
@@ -64,7 +65,13 @@ constexpr uint32_t kNumRows = 1000;
 constexpr uint32_t kNumQueries = 10;
 constexpr uint32_t kDim = 128;
 constexpr uint32_t kLargeDim = 256;
-constexpr uint32_t kK = 10;
+constexpr uint32_t kK = 64;
```
liliu-z, line 68: `kK` changed from 10→64 and `search_list_size` from 36→64 globally, affecting CPU-only test paths too. Was this to accommodate GPU-built graphs needing wider search, or just to match the new `kK`? If the former, consider keeping separate params for GPU/CPU tests to avoid masking potential recall differences.

liorf95:
Since it was all already merged into main, I will handle all the above new comments on a new PR from a new forked branch.

liorf95:
All the above new review comments were addressed in the new PR #1481.

PR #1469, opened 2026-02-21
fix: move arm_sve.h include inside __ARM_FEATURE_SVE guard

xiaofan-luan:
## Summary
- `distances_sve.cc` and `distances_sve.h` included `<arm_sve.h>` unconditionally outside the `#if defined(__ARM_FEATURE_SVE)` block
- On platforms without SVE (e.g. Apple Silicon), the file is compiled without SVE flags, causing `arm_sve.h` to emit `#error "SVE support not enabled"`
- Move the include inside the existing `#ifdef` guard, consistent with how `hook.cc` and upstream faiss already handle it

issue: #1470

## Test plan
- [x] Verified local build on Apple Silicon (M-series) compiles successfully
- [ ] CI passes on ARM64 Linux (SVE enabled) — no functional change expected

### DIFF
--- a/src/simd/distances_sve.cc
+++ b/src/simd/distances_sve.cc
@@ -11,12 +11,11 @@
 
 #include "distances_sve.h"
 
-#include <arm_sve.h>
-
 #include <cmath>
 
 #include "faiss/impl/platform_macros.h"
 #if defined(__ARM_FEATURE_SVE)
+#include <arm_sve.h>
 namespace faiss {
 namespace cppcontrib {
 namespace knowhere {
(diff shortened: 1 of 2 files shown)


xiaofan-luan:
> ## Summary
> * `distances_sve.cc` and `distances_sve.h` included `<arm_sve.h>` unconditionally outside the `#if defined(__ARM_FEATURE_SVE)` block
> * On platforms without SVE (e.g. Apple Silicon), the file is compiled without SVE flags, causing `arm_sve.h` to emit `#error "SVE support not enabled"`
> * Move the include inside the existing `#ifdef` guard, consistent with how `hook.cc` and upstream faiss already handle it
> 
> issue: #1470
> 
> ## Test plan
> * [x]  Verified local build on Apple Silicon (M-series) compiles successfully
> * [ ]  CI passes on ARM64 Linux (SVE enabled) — no functional change expected
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

liliu-z (commented):
/lgtm
/approve

PR #1472, opened 2026-02-24
fix: add fmt::fmt-header-only alias for Cardinal compatibility

foxspy:
Cardinal v1/v2 link against fmt::fmt-header-only, but fmt 11.0.2 no longer provides this target. Add an INTERFACE IMPORTED alias that forwards to fmt::fmt.


foxspy:
issue: #1344

chasingegg (approved):
/lgtm

PR #1480, opened 2026-03-04
ci: centralize conan install flags into scripts/conan_build.sh

jamesgao-jpg:
## Summary
- Adds `scripts/conan_build.sh` — single source of truth for conan install flags
- Updates all 5 Jenkins Groovy pipelines and 4 GitHub Actions workflows to call it
- Eliminates flag duplication that caused #1479 to miss test-jobs/nightly.groovy

## Motivation
Every time a conan flag changes (e.g. `--update` in #1479, `-s compiler.cppstd=17` in #1474), 10+ files across 2 repos need updating. Files in test-jobs are always missed because there's no link between the repos. This script centralizes the flags so future changes only need one edit.

## Script interface
```
scripts/conan_build.sh [--gpu] [--cardinal] [--ut] [--asan] [--cpu] [--release] [--no-cppstd] [--no-build]
```
Base flags (always applied): `--update --build=missing -s compiler.libcxx=libstdc++11 -s compiler.cppstd=17 -o with_diskann=True`

## No behavioral changes
- `release.yaml` and `release-python.yml` use `--no-cppstd` to preserve their existing behavior (they historically built without `-s compiler.cppstd=17`)
- All other pipelines produce identical conan commands as before

## Test plan
- [x] `ut.yaml` CI passes on this PR (exercises `--cpu --ut --asan --release` and `--cpu --release`)
- [x] Manual review: each pipeline's old flags match the new script call

## Follow-up
- test-jobs PR to update `nightly.groovy` and `nightly-arm.groovy` (depends on this PR merging first)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1482, opened 2026-03-05
ci: add unified Makefile and centralize CI dependencies

### LINKED ISSUE #1483: ci: add unified Makefile and centralize CI dependencies
## Problem

Every CI pipeline (4 GHA workflows + 5 Groovy pipelines) duplicates its own inline `apt-get`/`pip3 install` commands and conan invocations. The same packages and flags appear across 9 files with minor variations, making it easy for versions to drift and hard to update consistently. There is no single source of truth for build flags or dependency versions.

Specific issues:
- Developers must remember different conan incantations for CPU, GPU, release, UT, etc.
- The same base packages (`libaio-dev`, `libcurl4-openssl-dev`, `conan==1.65.0`, etc.) are repeated across 9 pipeline files
- No shared definition for CI dependency versions — conan version has already drifted between files in the past

## Proposed Solution

1. **Top-level Makefile** as a unified developer CLI (matching the [Milvus Makefile](https://github.com/milvus-io/milvus/blob/master/Makefile) pattern), centralizing conan flags as Make variables.
2. **Composable `scripts/ci_deps.sh`** with sourced shell functions for dependency installation, pinning versions in one place.
3. **Update all CI pipelines** to use Makefile targets and `ci_deps.sh` functions.

## Related PR

- https://github.com/zilliztech/knowhere/pull/1482

jamesgao-jpg:
Closes #1483

## Motivation

Every CI pipeline (4 GHA workflows + 5 Groovy pipelines) duplicates its own inline `apt-get`/`pip3 install` commands and conan invocations. The same packages and flags appear across 9 files with minor variations, making it easy for versions to drift and hard to update consistently. There is no single source of truth for build flags or dependency versions.

This PR addresses two problems:
1. **No unified build interface** — developers must remember different conan incantations for CPU, GPU, release, UT, etc. Milvus solves this with a [top-level Makefile](https://github.com/milvus-io/milvus/blob/master/Makefile); knowhere should follow the same pattern for organizational consistency.
2. **Duplicated CI dependencies** — the same base packages are repeated across 9 pipeline files with no shared definition, leading to inconsistencies (e.g. different conan versions, missing packages in some pipelines).

## Summary

- **Add top-level `Makefile`** as a unified developer CLI (matching the Milvus Makefile pattern). Centralizes conan flags as Make variables (single source of truth) and uses composable flags for build variants:
  ```
  make                                            # CPU release (default)
  make WITH_GPU=True                              # GPU release
  make WITH_UT=True                               # CPU with unit tests
  make WITH_CARDINAL=True                         # Cardinal build
  make WITH_DEBUG=True                            # Debug build
  make CONAN_PROFILE=clang14                      # custom compiler via Conan profile
  make LIBCXX=libc++                              # override compiler.libcxx
  make test                                       # run unit tests
  make lint                                       # clang-tidy
  make format                                     # clang-format
  make wheel                                      # python wheel
  make codecov                                    # coverage report
  make clean                                      # cleanup
  ```
- **Add `scripts/install_deps.sh`** — single dependency installer script that installs all packages needed across CI pipelines in one shot. **Note: Since most pipelines have similar depdendency requirements, after discussing, we decide NOT to provide a more modular control flag option to control fine-grained dependency installation. Users should refer to install_deps.sh if they desire otherwise.**

- **Update `README.md`** — replace raw conan commands with Makefile targets, document all build flags, simplify the wheel section, update verified OS list (22.04 primary, 20.04 EOL with CMake caveat), and add `make pre-commit`/`make format`/`make lint` to the contributing section.

- **Update all CI pipelines** (4 GHA workflows + 5 Groovy pipelines) to use Makefile targets and `install_deps.sh`, replacing duplicated inline blocks:

  | Pipeline | Before | After |
  |----------|--------|-------|
  | `ut.yaml` (ut) | inline apt/pip + conan install + build + run | `bash scripts/install_deps.sh` + `make WITH_UT=True WITH_ASAN=True` + `make test` |
  | `ut.yaml` (swig-build) | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `make` |
  | `analyzer.yaml` | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `make WITH_UT=True` |
  | `release.yaml` | inline conan install + build | `make` |
  | `release-python.yml` | inline conan install + build | `make` |
  | `E2E2.groovy` | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `make` |
  | `E2E-arm.groovy` | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `make` |
  | `E2E2-SSE.groovy` | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `libopenblas-dev` override + `make` |
  | `E2E_GPU.groovy` | inline apt/pip + conan install + build | `bash scripts/install_deps.sh` + `build-essential` + `make WITH_GPU=True` |
  | `UT_GPU.groovy` | inline apt/pip + conan install + build + run | `bash scripts/install_deps.sh` + `git` + `make WITH_GPU=True WITH_UT=True` + `make test` |

### Notes
- Ubuntu 20.04 is listed as EOL in the README with a caveat that users must manually install CMake >= 3.26. The `install_deps.sh` script targets Ubuntu 22.04. We kept Ubuntu 20.04 in README for legacy reason.

### Relation to other PRs
- PR https://github.com/zilliztech/knowhere/pull/1480 takes a narrower approach (shell script for conan flags only). This PR supersedes it with a broader Makefile + dependency centralization approach.

## Test plan
- [x] Verify `ut.yaml` CI passes (unit tests + swig-build)
- [x] Verify `analyzer.yaml` CI passes (clang-tidy)
- [x] Verify E2E Groovy pipelines pass (E2E2, E2E-arm, E2E2-SSE, E2E_GPU, UT_GPU)
- [x] Verify `make build`, `make test`, `make WITH_GPU=True` work locally
- [x] Verify `CONAN_PROFILE` flag works with custom compiler (tested with clang-14 profile)
- [x] Verify macOS auto-detection via `uname -s` dry run
- [x] Verify compilation works on MacOS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ BUILD_DIR := $(PWD)/build
 SHELL := /bin/bash
 
 # Conan base flags (single source of truth for all pipelines)
-CONAN_BASE_FLAGS := --update --build=missing -o with_diskann=True
+CONAN_BASE_FLAGS := --update --build=missing
 
 # ---------- User-facing build flags ----------
 # Usage: make WITH_GPU=True WITH_UT=True WITH_ASAN=True WITH_DEBUG=True
@@ -41,16 +41,20 @@ else
 endif
 
 # ---------- Compose conan flags from user flags ----------
-CONAN_FLAGS := $(CONAN_BASE_FLAGS) -s compiler.libcxx=$(LIBCXX) -s build_type=$(BUILD_TYPE)
+CONAN_FLAGS := $(CONAN_BASE_FLAGS) -s compiler.libcxx=$(LIBCXX) -s build_type=$(BUILD_TYPE) -s compiler.cppstd=17
 
-# GPU builds use cuVS; CPU builds need liburing built from source.
-ifdef WITH_GPU
-    CONAN_FLAGS += -o with_cuvs=True
-else
-    CONAN_FLAGS += --build=liburing
+# DiskANN and liburing require libaio (Linux-only).
+ifneq ($(UNAME_S),Darwin)
+    CONAN_FLAGS += -o with_diskann=True
+    ifndef WITH_GPU
+        CONAN_FLAGS += --build=liburing
+    endif
 endif
 
-CONAN_FLAGS += -s compiler.cppstd=17
+# GPU builds use cuVS.
+ifdef WITH_GPU
+    CONAN_FLAGS += -o with_cuvs=True
+endif
 
 ifdef WITH_UT
     CONAN_FLAGS += -o with_ut=True


jamesgao-jpg:
issue: #1483 
/kind improvement

Makefile
```diff
@@ -0,0 +1,118 @@
...
+
+# ---------- Build targets ----------
+
+build: ## Build CPU (debug)
+	@echo "Building Knowhere (CPU, debug)..."
+	@mkdir -p $(BUILD_DIR) && cd $(BUILD_DIR) && \
+		conan install .. $(CONAN_CPU_FLAGS) && \
+		conan build ..
+
+build-release: ## Build CPU (release)
```
foxspy: use release as default

Makefile
```diff
@@ -0,0 +1,118 @@
...
+		conan install .. $(CONAN_CPU_FLAGS) -s build_type=Release && \
+		conan build ..
+
+build-ut: ## Build CPU with unit tests enabled (no run)
+	@echo "Building Knowhere (CPU, with unit tests)..."
+	@mkdir -p $(BUILD_DIR) && cd $(BUILD_DIR) && \
+		conan install .. $(CONAN_CPU_FLAGS) -o with_ut=True && \
+		conan build ..
+
+build-gpu: ## Build GPU (release)
```
foxspy: make with_gpu=True

Makefile
```diff
@@ -0,0 +1,118 @@
...
+		conan install .. $(CONAN_CPU_FLAGS) && \
+		conan build ..
+
+build-release: ## Build CPU (release)
+	@echo "Building Knowhere (CPU, release)..."
+	@mkdir -p $(BUILD_DIR) && cd $(BUILD_DIR) && \
+		conan install .. $(CONAN_CPU_FLAGS) -s build_type=Release && \
+		conan build ..
+
+build-ut: ## Build CPU with unit tests enabled (no run)
```
foxspy: make with_ut=True

foxspy:
@jamesgao-jpg some comments here
  ## Comment 1 — on Makefile (general)

**Decouple build options from targets — use composable variables instead of combinatorial targets**

The current design hardcodes option combinations into target names (`build-ut`, `build-release`, `build-gpu`, `ut-gpu`). This creates a combinatorial explosion — every new option (ASAN, coverage, cardinal...) doubles the number of targets, and arbitrary combinations (e.g. GPU + UT + ASAN) are impossible without adding yet another target.

Instead, keep a single `build` target and let users compose options freely via Make variables:

```makefile
# Composable - any combination works
make build WITH_UT=True WITH_CUVS=True WITH_ASAN=True

# Convenience presets for common combos
make gpu    # delegates to: make build WITH_CUVS=True
```

This is the standard pattern used by Linux kernel (`make ARCH=arm64`), Redis (`make USE_JEMALLOC=yes`), and Milvus itself (`USE_ASAN=ON make milvus`).

---

## Comment 2 — on Makefile (near variable definitions)

**Use `WITH_*` variable naming to stay consistent with existing conan options**

Knowhere's `conanfile.py` already defines options as `with_ut`, `with_diskann`, `with_cuvs`, `with_asan`, etc. The Makefile variables should mirror this naming directly:

```makefile
WITH_UT       ?= False
WITH_DISKANN  ?= True
WITH_CUVS     ?= False
WITH_ASAN     ?= False
BUILD_TYPE    ?= Release
```

This avoids a naming mismatch between the Makefile interface and the underlying build system, and makes the mapping self-evident.

---

## Comment 3 — on Makefile (near conan invocations)

**Define the conan base command once at the top, then append per-option flags conditionally**

Currently each target duplicates the full conan invocation with slightly different flags. Instead, define a single base command and let variables append to it:

```makefile
CONAN_BASE := conan install .. --update --build=missing \
    -s compiler.libcxx=$(LIBCXX) -s compiler.cppstd=17 \
    -s build_type=$(BUILD_TYPE)

# Each WITH_* appends its flag
CONAN_OPTIONS := \
    -o with_ut=$(WITH_UT) \
    -o with_diskann=$(WITH_DISKANN) \
    -o with_cuvs=$(WITH_CUVS) \
    -o with_asan=$(WITH_ASAN)

build:
	@mkdir -p $(BUILD_DIR) && cd $(BUILD_DIR) && \
		$(CONAN_BASE) $(CONAN_OPTIONS) && conan build ..
```

This eliminates the divergence between `CONAN_CPU_FLAGS` / `CONAN_GPU_FLAGS` and makes the Makefile a thin, declarative wrapper over conan — single source of truth, zero duplication.

jamesgao-jpg:
@foxspy Thanks for the review! Your suggestions have been addressed:

1. **Release is now the default** — `build_type=Release` unless `WITH_DEBUG=True` is specified.
2. **Flag-style interface** — named targets (`build-release`, `build-gpu`, `build-ut`) replaced with composable flags:
   ```
   make                              # CPU release (default)
   make WITH_GPU=True                # GPU release
   make WITH_UT=True                 # CPU with unit tests
   make WITH_UT=True WITH_ASAN=True  # CPU UT + ASAN
   make WITH_DEBUG=True              # CPU debug
   make WITH_MACOS=True              # macOS (libc++)
   make test                         # run unit tests
   ```
3. **New flags added** — `WITH_DEBUG` (switches build_type to Debug) and `WITH_MACOS` (switches compiler.libcxx to libc++) to support debug builds and macOS developers.

Regarding README.md — not updating it in this PR. The current README has accumulated significant tech debt:
- System Requirements lists Ubuntu 20.04 but CI runs on 22.04
- macOS build dependencies were never fully documented
- Python wheel section has an ALL CAPS header and manually lists deps that `ci_deps.sh` already handles
- Contributing section doesn't reference any Makefile targets (`make format`, `make lint`, etc.)
- `install_deps.sh` and `python_deps.sh` overlap heavily with the new `ci_deps.sh`

A separate PR to clean up the README and consolidate the dependency scripts would be more appropriate.

The focus of this PR is providing the unified Makefile + `ci_deps.sh` interface so other repos (test-jobs, cardinal) can follow up and use it consistently.

alexanderguzhva:
@jamesgao-jpg how complicated is it to allow passing a custom profile for `conan`? The reason is that I use `clang` or custom build `gcc` versions (such as `gcc-15`) for the compilation, and I'd like to have it handy, if possible

jamesgao-jpg:
Latest: Never mind, did that as well in this pr since that requires minimum change. Please check this commit:
https://github.com/zilliztech/knowhere/pull/1482/changes/78a40b93487a2d9b2c8ba8cec77eff12abada924

######################

@alexanderguzhva Thanks for the suggestion! It would be straightforward — just a `CONAN_PROFILE` Make variable that gets passed as `--profile` to the `conan install` call, so you could do something like `make CONAN_PROFILE=gcc15` or `make CONAN_PROFILE=clang`.

That said, this PR is already fairly large (Makefile + `ci_deps.sh` + 9 pipeline updates), so I'd prefer to keep it focused on centralizing what already exists, as the main purpose of this pr is to provide a uniform interface that other repos can refer to to mitigate the "Shotgun Surgery" problem with CI/CD we had before. Custom profile support would be a new capability that the original build procedure didn't have either.

I'll add it as an immediate follow-up PR once this one lands — should be a small, clean change. Would that work for you?

.github/workflows/release-python.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           CIBW_BUILD: "cp38-* cp39-* cp310-* cp311*"
           CIBW_SKIP: "*musllinux*" # faiss-cpu fails to build on musllinux
           CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_BEFORE_ALL_LINUX: "bash scripts/python_deps.sh && rm -rf build && mkdir build && cd build && conan install .. --update --build=missing -o with_diskann=True -s compiler.libcxx=libstdc++11 -s build_type=Release && conan build .. && cd -"
+          CIBW_BEFORE_ALL_LINUX: "bash scripts/python_deps.sh && make WITH_CPPSTD=False"
```
foxspy: What's the WITH_CPPSTD=False flag do?

scripts/ci_deps.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+CONAN_VERSION="1.65.0"
+CONAN_REMOTE_URL="https://milvus01.jfrog.io/artifactory/api/conan/default-conan-local"
+
+echo "[ci_deps] Installing dependencies..."
+
+${SUDO} apt-get update || true
+${SUDO} apt-get install -y \
+    libaio-dev \
```
foxspy: apt-get is not supported on macOS, and libaio cannot be found or linked on that platform. We need to ensure the scripts handle these package manager and dependency differences. Please ensure that the scripts are cross-platform compatible

scripts/ci_deps.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
foxspy: Please rename this file to install_deps.sh to better reflect its purpose.

scripts/ci_deps.sh
jamesgao-jpg: Originally ci_deps.sh was meant for dependency for CI pipelines only. In the latest commits, we updated it and removed unused scripts such as `install_deps.sh` and 'python_deps.sh' and moved their relevant section to `ci_deps.sh`, which becomes our new latest `install_deps.sh` that handles all dependency installation.

.github/workflows/release-python.yml
jamesgao-jpg: It was originally added so release.yaml and release-python.yaml could keep their previous behaviors of compiling with the release environment's compiler's default cpp version standard (as I was afraid the release versions could have mismatched ABI versions). 
After careful review, this is unnecessary and hence moved in the latest commits.

alexanderguzhva:
issue: #1483

PR #1486, opened 2026-03-09
get rid of C++ Variable Length Arrays in diskann

alexanderguzhva:
removes unexpected errors when compiling using `clang`.

Replace constructs like 
```C++
  uint32_t some_array[non_constant];
```
to either `std::unique_ptr<T[]>`, or `MaybeArray<T>` (which is used for avoid excessive allocations for tiny arrays)


alexanderguzhva:
issue: #1489
/kind improvement

foxspy (approved):
/lgtm

PR #1496, opened 2026-03-10
get rid of the `is_cosine` field

alexanderguzhva:
This is the change, targeted at lowering the burden of the `faiss::cppcontrib::knowhere`. 

The following the most important things are implemented:
* Remove the `is_cosine` field
  * All L2 norms are now stored in the memory in the `1 / ||x||` format 
* Simplify the hierarchy of `knowhere`'s custom components for `FAISS`

### DIFF
--- a/src/common/comp/brute_force.cc
+++ b/src/common/comp/brute_force.cc
@@ -572,8 +572,9 @@ BruteForce::SearchWithBuf(const DataSetPtr base_dataset, const DataSetPtr query_
                 ThreadPool::ScopedSearchOmpSetter setter(1);
                 auto cur_labels = labels + topk * index;
                 auto cur_distances = distances + topk * index;
-                RETURN_IF_ERROR(brute_force_dense_impl<DataType>(xq, index, xb, inv_norms.get(), cur_labels, cur_distances,
-                                                                 dim, nb, topk, faiss_metric_type, bitset, is_cosine));
+                RETURN_IF_ERROR(brute_force_dense_impl<DataType>(xq, index, xb, inv_norms.get(), cur_labels,
+                                                                 cur_distances, dim, nb, topk, faiss_metric_type,
+                                                                 bitset, is_cosine));
 
                 return Status::success;
             }));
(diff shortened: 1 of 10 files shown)


alexanderguzhva:
issue: #1501 
/kind improvement

alexanderguzhva:
issue: #1501 
/kind improvement

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/index_write.cpp
```diff
@@ -466,9 +470,6 @@ void write_index(const Index* idx, IOWriter* f, int io_flags) {
         WRITE1(h);
         write_index_header(idx, f);
         WRITEXBVECTOR(idxf->codes);
-        if (idx->is_cosine) {
```
foxspy, line 469: Will this introduce any compatibility issues?

alexanderguzhva: an excellent question. Let me see

alexanderguzhva: @foxspy A long answer is the following.

Any old version of the serialization code would serialize `IndexFlat` with the `Cosine` metric and then write the following norms via 
```C++
        if (idx->is_cosine) {
            WRITEVECTOR(idxf->code_norms);
        }
```
This will be correctly deserialized in a new version and the index type will be promoted to `IndexFlatCosine` internally. This `IndexFlatCosine` has been introduced in #746, which is an August 2024, and the change went live in `knowhere` version `2.3.8`. It seems that `knowhere` version `2.3.9` with this change went into Milvus `2.4.11` (https://github.com/milvus-io/milvus/pull/36067). So, any `IndexFlat` with the `Cosine` metric trained in  knowhere `2.3.8` (or newer) would not be read in `2.3.7` or earlier, bcz `2.3.7` knows nothing about `IndexFlatCosine` and its fourcc indicator, used for distinguishing indices.

Given this, no new problems will be introduced.

PR #1498, opened 2026-03-13
nit: Hoist bf_query_dataset construction out of candidate-doc loop

lyang24:
Move query-tensor pointer calculation and GenDataSet() call from the inner per-candidate loop to the outer per-query-emb-list loop in stage 2. These values are invariant across candidate docs, so building them once per query emb-list avoids repeated small-object allocation overhead.

### DIFF
--- a/src/index/index_node.cc
+++ b/src/index/index_node.cc
@@ -295,6 +295,11 @@ IndexNode::SearchEmbList(const DataSetPtr dataset, std::unique_ptr<Config> cfg, 
             el_ids_set.emplace(emb_list_offset_->get_el_id((size_t)stage1_ids[j]));
         }
 
+        // Generate query dataset once per query emb_list (invariant across candidate docs)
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@foxspy were there any problems with our CI recently?

alexanderguzhva:
@lyang24 waiting for CI issues to get resolved

jamesgao-jpg:
@lyang24 Hi, we checked the failure on CI and it seems that the problem arises because your current branch is missing a few refactor commits we had to knowhere CI. You may want to rebase from main first.

lyang24:
> @lyang24 Hi, we checked the failure on CI and it seems that the problem arises because your current branch is missing a few refactor commits we had to knowhere CI. You may want to rebase from main first.

thank you rebased

alexanderguzhva:
issue: #1513 
/kind improvement

PR #1500, opened 2026-03-13
enhance: implement GetEmbListByIds

SpadeA-Tang:
issue: https://github.com/zilliztech/knowhere/issues/1499

### DIFF
--- a/src/index/index_node.cc
+++ b/src/index/index_node.cc
@@ -450,7 +450,6 @@ IndexNode::GetEmbListByIds(const DataSetPtr dataset, milvus::OpContext* op_conte
     // index-type-specific implementations (HNSW, IVF, FLAT, etc. each store raw data differently),
     // whereas the current approach works generically across all index types via the GetVectorByIds interface.
     std::vector<size_t> out_offsets(num_el_ids + 1);
-    std::vector<int64_t> vec_ids;
     out_offsets[0] = 0;
     for (int64_t i = 0; i < num_el_ids; i++) {
         auto el_id = el_ids[i];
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
/kind improvement
issue: #1499

src/index/index_node.cc
```diff
@@ -429,6 +429,71 @@ IndexNode::AnnIteratorEmbListIfNeed(const DataSetPtr dataset, std::unique_ptr<Co
...
+    //
+    // TODO(perf): Vectors within each embedding list are contiguous in the index. However, the current
+    // implementation collects all these contiguous IDs into a flat array and passes them to GetVectorByIds,
+    // which internally calls reconstruct(id, ...) one vector at a time. This could be optimized by using
+    // reconstruct_n(start, len, ...) or direct memcpy from raw data storage, avoiding both the redundant
+    // ID array allocation and per-vector overhead. We don't do this yet because it would require
+    // index-type-specific implementations (HNSW, IVF, FLAT, etc. each store raw data differently),
+    // whereas the current approach works generically across all index types via the GetVectorByIds interface.
+    std::vector<size_t> out_offsets(num_el_ids + 1);
+    std::vector<int64_t> vec_ids;
```
zhengbuqian: nit: reserve for vec_ids.

maybe reserve `num_el_ids` or `2 * num_el_ids` if we don't want to do another looping to get the final size.

zhengbuqian (approved):

src/index/index_node.cc
SpadeA-Tang: loop may be better as the capacity here is highly unpredictable.

PR #1471, opened 2026-02-23
Add DSP (Dynamic Superblock Pruning) sparse index with avx 512SIMD optimizations

lyang24:
- DSP index with u8/u16 integer pruning and two-level block hierarchy (trade off between 1/4 of the forward index size and slight accuracy)  
- AVX-512 SIMD: gather/scatter IP accumulation, block UB scan, seek (the original paper uses avx2)

PR #1504, opened 2026-03-16
feat: add sparse DSP CC index

lyang24:


alexanderguzhva:
@sparknack could you please take a look as well? Thanks

cmake/libs/libfaiss.cmake
```diff
@@ -62,7 +62,7 @@ if(__X86_64)
                                               -mavx512bw -mpopcnt -mavx512vl)
   target_compile_options(utils_avx512icx PRIVATE -mfma -mf16c -mavx512f -mavx512dq
                                               -mavx512bw -mpopcnt -mavx512vl -mavx512vpopcntdq)
-  target_compile_options(sparse_simd_avx512 PRIVATE -mavx512f -mavx512dq)
+  target_compile_options(sparse_simd_avx512 PRIVATE -mavx512f -mavx512dq -mavx512bw)
```
alexanderguzhva: I think that you can safely add `-mavx512vl` and `-mavx512cd` as well, in order to help a compiler to generate a better code

src/index/sparse/sparse_dsp_config.h
```diff
@@ -0,0 +1,103 @@
...
+#ifndef SPARSE_DSP_CONFIG_H
+#define SPARSE_DSP_CONFIG_H
+
+#include "knowhere/config.h"
+
+namespace knowhere {
+
+// Search parameters for the DSP (Dynamic Superblock Pruning) index.
+//
+// Mode selection (dsp_mode):
```
alexanderguzhva: please add a comment that these values are mapped to the `DspSearchMode` enum class in the code.

src/index/sparse/sparse_dsp_config.h
```diff
@@ -0,0 +1,103 @@
...
+            .set_default(0)
+            .description(
+                "superblock selection mode: "
+                "0=dsp (dual-threshold mu/eta + optional top-gamma backstop), "
+                "1=lsp0 (top-gamma from ub>=theta, no mu/asc gate), "
+                "2=lsp1 (lsp0 + mu-overestimation gate: ub>theta/mu), "
+                "3=lsp2 (lsp1 + asc gate: ub>theta/mu || asc>theta/eta)")
+            .for_search();
+        KNOWHERE_CONFIG_DECLARE_FIELD(dsp_mu)
+            .set_range(0.1, 2.0)
```
alexanderguzhva: where do limitations come from? A range of recommended values from an article?

src/index/sparse/sparse_dsp_config.h
```diff
@@ -0,0 +1,103 @@
...
+            .set_default(1.0)
+            .description("superblock max-based threshold relaxation factor (used by dsp/lsp1/lsp2)")
+            .for_search();
+        KNOWHERE_CONFIG_DECLARE_FIELD(dsp_eta)
+            .set_range(0.1, 2.0)
```
alexanderguzhva: same

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+    DIM_MAP = 2,
+    ROW_SUMS = 3,
+    MAX_SCORES_PER_DIM = 4,
+    PROMETHEUS_BUILD_STATS = 5,
+    DSP_METADATA = 6
+};
+
+struct DspSectionHeader {
+    DspSectionType type;
+    uint64_t offset;
```
alexanderguzhva: please initialize fields with a default value

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+// Superblock selection modes for DSP search.
+enum class DspSearchMode : int {
+    DSP = 0,   // dual-threshold (mu, eta) + optional top-gamma backstop
+    LSP0 = 1,  // top-gamma from ub>=theta, no mu/asc gate (recommended for SPLADE)
+    LSP1 = 2,  // LSP/0 + mu-overestimation gate (ub>theta/mu)
+    LSP2 = 3,  // LSP/1 + asc gate (ub>theta/mu || asc>theta/eta)
+};
+
+struct DspSearchParams {
+    int refine_factor;
```
alexanderguzhva: please initialize fields with a default value

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+template <typename DType, typename QType, bool mmapped = false>
+class DspIndex : public DspIndexBase<DType> {
+ public:
+    template <typename U>
+    using Vector = std::conditional_t<mmapped, GrowableVectorView<U>, std::vector<U>>;
+
+    static constexpr uint32_t kSubblockSize = 8;
+    static constexpr uint32_t kSuperblockSize = 512;
+    static constexpr uint32_t kStride = kSuperblockSize / kSubblockSize;  // 64
+    static constexpr uint32_t kSimdWidth = 32;                            // AVX-512 processes 32 u16 values
```
alexanderguzhva, line 138: do I get it correct that these values are marked as `constexpr`, because they are just not going to be changed? For example, ARM SVE CPUs may have a non-constant `SIMD` width, would it be theoretically possible to alter the code in this case and remove `constexpr`? 
There's no need to change the code, it's just a question for the future.

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+
+// DSP (Dynamic Superblock Pruning) index for fast sparse vector search.
+//
+// DSP index structure:
+// - u8 quantized block max scores
+// - u16 upper bound accumulators with AVX-512 SIMD
+// - Counting sort (bucket sort) for block ordering by upper bound
+// - Forward index with two-pointer merge scoring
+// - Two-level hierarchy: superblocks for coarse pruning, subblocks for scoring
+template <typename DType, typename QType, bool mmapped = false>
```
alexanderguzhva, line 129: any specific reasons to put `mmapped` as a template parameter instead of introducing a `bool DspIndex::mmapped` field?

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+    // DSP Search
+    // ========================================================================
+    template <typename DocIdFilter>
+    void
+    search_dsp(const std::vector<std::pair<size_t, DType>>& q_vec, MaxMinHeap<float>& heap, size_t heap_capacity,
+               DocIdFilter& filter, const DocValueComputer<float>& computer, DspSearchMode mode, float mu, float eta,
+               int gamma, bool kth_init = true, float kth_alpha = 1.0f) const {
+        // ---- Step 0: Prepare sorted query ----
+        struct QueryTerm {
+            uint32_t inner_dim;
```
alexanderguzhva: initialize fields with default values

src/simd/sparse_simd_avx512.cc
```diff
@@ -9,15 +9,67 @@
...
+scan_block_ub_any_above_avx512_64(const uint16_t* block_ub, uint16_t threshold) {
+    const __m512i thresh_v = _mm512_set1_epi16(static_cast<int16_t>(threshold));
+    __m512i v0 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(block_ub));
+    __m512i v1 = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(block_ub + 32));
+    __mmask32 mask =
+        _mm512_cmp_epu16_mask(v0, thresh_v, _MM_CMPINT_NLE) | _mm512_cmp_epu16_mask(v1, thresh_v, _MM_CMPINT_NLE);
+    return mask != 0;
+}
+
+// Generic: loop over n elements in 32-element chunks. n must be a multiple of 32.
```
alexanderguzhva, line 48: what if `(n % 32 != 0)` ?
basically, I don't see how the caller of this function handles this situation and whether the `(n % 32 != 0)` situation can happen at all. A comment would be appreciated

src/simd/sparse_simd_avx512.cc
```diff
@@ -80,4 +132,67 @@ accumulate_posting_list_ip_avx512(const uint32_t* doc_ids, const float* doc_vals
...
+        __m512i bm16 = _mm512_cvtepu8_epi16(bm8);                                   \
+        __m512i prod = _mm512_mullo_epi16(bm16, qw_vec);                            \
+        __m512i cur = _mm512_loadu_si512(reinterpret_cast<const __m512i*>(ub_ptr)); \
+        cur = _mm512_adds_epu16(cur, prod);                                         \
+        _mm512_storeu_si512(reinterpret_cast<__m512i*>(ub_ptr), cur);               \
+    } while (0)
+
+// n == 32: single iteration, no loop
+void
+accumulate_block_ub_avx512_32(uint16_t* ub, const uint8_t* block_max, uint16_t query_weight) {
```
alexanderguzhva: I would add a `__restrict` qualifier to these pointers
`void accumulate_block_ub_avx512_32(uint16_t* __restrict ub, const uint8_t* __restrict block_max, uint16_t query_weight)`, just to ensure that compiler parallelizes certain SIMD intrinsics.

The same comment applies to the other similar functions down below

src/index/sparse/sparse_index_node.cc
```diff
@@ -650,6 +652,449 @@ class SparseInvertedIndexNodeCC : public SparseInvertedIndexNode<T, use_wand> {
...
+
+ public:
+    explicit SparseDspIndexNode(const int32_t& version, const Object& /*object*/)
+        : search_pool_(ThreadPool::GetGlobalSearchThreadPool()),
+          build_pool_(ThreadPool::GetGlobalBuildThreadPool()),
+          index_version_(version) {
+    }
+
+    ~SparseDspIndexNode() override {
+        delete index_;
```
sparknack, line 672: add `munmap` in `~SparseDspIndexNode()` and unmap the old mapping before re-mmap in `DeserializeFromFile()` (consider adopting `MmapGuard` like the existing sparse node)

src/index/sparse/sparse_dsp_config.h
```diff
@@ -0,0 +1,103 @@
...
+ public:
+    CFG_FLOAT drop_ratio_search;
+    CFG_INT refine_factor;
+    CFG_INT dsp_mode;
+    CFG_FLOAT dsp_mu;
+    CFG_FLOAT dsp_eta;
+    CFG_INT dsp_gamma;
+    CFG_BOOL dsp_kth_init;
+    CFG_FLOAT dsp_kth_alpha;
+    KNOHWERE_DECLARE_CONFIG(SparseDspConfig) {
```
sparknack, line 39: `KNOHWERE` → `KNOWHERE`?

src/index/sparse/sparse_index_node.cc
```diff
@@ -672,5 +1118,6 @@ KNOWHERE_SIMPLE_REGISTER_SPARSE_FLOAT_GLOBAL(SPARSE_INVERTED_INDEX_CC, SparseInv
                                              /*use_wand=*/false)
 KNOWHERE_SIMPLE_REGISTER_SPARSE_FLOAT_GLOBAL(SPARSE_WAND_CC, SparseInvertedIndexNodeCC, knowhere::feature::MMAP,
                                              /*use_wand=*/true)
+KNOWHERE_SIMPLE_REGISTER_SPARSE_FLOAT_GLOBAL(SPARSE_DSP_CC, SparseDspIndexNodeCC, knowhere::feature::MMAP)
```
sparknack, line 1136: register `SPARSE_DSP`?

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+            if (raw_rows[vec_id].size() > 0) {
+                writer.write(raw_rows[vec_id].data(), raw_rows[vec_id].size() * SparseRow<DType>::element_size());
+            }
+        }
+
+        return Status::success;
+    }
+
+    Status
+    DeserializeV0(MemoryIOReader& reader, int map_flags, const std::string& supplement_target_filename) override {
```
sparknack: V0 related code can be removed.

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+                    computer(inverted_index_vals_spans_[dim_it->second][it - plist_ids.begin()],
+                             metric_type_ == SparseMetricType::METRIC_BM25 ? bm25_params_->row_sums_spans_[vec_id] : 0);
+            }
+        }
+
+        return distance;
+    }
+
+    [[nodiscard]] size_t
+    size() const override {
```
sparknack, line 692: update the `size()` calculation

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+                        reader.read(&packed_size, sizeof(uint32_t));
+                        if (packed_size > 0) {
+                            bm.max_scores.resize(packed_size);
+                            reader.read(bm.max_scores.data(), sizeof(uint8_t), packed_size);
+                        }
+                    }
+
+                    uint32_t spb_total = 0;
+                    reader.read(&spb_total, sizeof(uint32_t));
+                    spb_dim_offsets_.resize(nr_dims + 1);
```
sparknack, line 574: if `spb_xxx_` and `fwd_xxx_` consume large amounts of memory, they should also use span to reduce memory overhead.

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1926 @@
...
+                    uint32_t index_encoding_type = 0;
+                    reader.read(&index_encoding_type, sizeof(uint32_t));
+                    if (index_encoding_type != 0) {
+                        return Status::invalid_serialized_index_type;
+                    }
+                    auto inverted_index_offsets_span = boost::span<const uint64_t>(
+                        reinterpret_cast<uint64_t*>(reader.data() + reader.tellg()), nr_inner_dims_ + 1);
+                    reader.advance(sizeof(uint64_t) * (nr_inner_dims_ + 1));
+                    inverted_index_ids_spans_.resize(nr_inner_dims_);
+                    inverted_index_vals_spans_.resize(nr_inner_dims_);
```
sparknack, line 485: seems `inverted_index_xxx_spans_` can be removed, since it is not used during the `Search()` phase. `GetAllDistances()` can also use the forward index instead.

src/index/sparse/sparse_index_node.cc
lyang24: https://github.com/zilliztech/knowhere/pull/1471#discussion_r2871970829
i though in previous comments cardinal will have compatibility issues so sparse_dsp is not needed

src/index/sparse/sparse_dsp_index.h
lyang24: they are constexpr both because they are fixed by the implementation today and because compile-time constants help simplify indexing and hot-path codegen.

  For a future target like ARM SVE, yes, some of this would likely need to change:

  - the SIMD-width-dependent part would need abstraction or runtime specialization
  - the layout constants could still remain fixed, unless we explicitly decide to redesign the index format too

src/index/sparse/sparse_dsp_index.h
lyang24: set default offest to 0

src/index/sparse/sparse_dsp_config.h
lyang24: fixed in comments

cmake/libs/libfaiss.cmake
lyang24: added

src/index/sparse/sparse_dsp_config.h
lyang24: that range is only acting as an implementation-side validation range, not a theoretical recommendation

src/index/sparse/sparse_dsp_config.h
lyang24: hat range is only acting as an implementation-side validation range, not a theoretical recommendation

src/index/sparse/sparse_dsp_config.h
lyang24: updated to paper's values
  - 0 < mu <= 1
  - 0 < eta <= 1
  - mu <= eta

src/index/sparse/sparse_dsp_index.h
lyang24: The main reason is to keep the mmap and non-mmap implementations as compile-time-specialized paths. They differ in storage ownership and a few control-flow decisions, and using the template bool lets us use if constexpr to remove the inactive path entirely rather than carrying a runtime branch/state through the implementation. If this
  abstraction becomes harder to maintain than it is worth, we can revisit a runtime field later.

src/index/sparse/sparse_dsp_index.h
lyang24: can we do this in a future pr?

src/index/sparse/sparse_dsp_index.h
lyang24: done

src/index/sparse/sparse_dsp_index.h
lyang24: done

src/index/sparse/sparse_index_node.cc
lyang24: done

src/simd/sparse_simd_avx512.cc
lyang24: added

src/simd/sparse_simd_avx512.cc
lyang24: added comment

src/index/sparse/sparse_dsp_index.h
lyang24: done

src/index/sparse/sparse_dsp_index.h
lyang24: big change will do it soon

lyang24: the change is big if performance regress its hard to track is it possible to do it as a follow up pr?

src/index/sparse/sparse_index_node.cc
sparknack: both `SPARSE_DSP` and `SPARSE_DSP_CC` are required in that comment, like `SPARSE_INVERTED_INDEX` and `SPARSE_INVERTED_INDEX_CC`.

lyang24: added back SPARSE_DSP

alexanderguzhva:
@lyang24 could you please rebase to the master? Thanks.

sparknack:
> @lyang24 could you please rebase to the master? Thanks.

@alexanderguzhva  master branch will still undergo many changes later, so please do not rebase for the time being. Merge into sparse_dsp_dev first, and we can cherry-pick the changes over later.

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1725 @@
...
+                    for (uint32_t spb = 0; spb < n_superblocks_; ++spb) {
+                        if (superblock_ub[spb] > mu_threshold || superblock_asc[spb] > eta_threshold) {
+                            mark_alive(spb);
+                        }
+                    }
+                    if (gamma > 0) {
+                        add_top_gamma(gamma, 0.0f);
+                    } else {
+                        uint32_t top2_ids[2] = {UINT32_MAX, UINT32_MAX};
+                        float top2_ub[2] = {0.0f, 0.0f};
```
sparknack: `top2` here may require further explanation: why is this necessary? and why specifically the number 2? also, is it secure?

src/index/sparse/sparse_index_node.cc
sparknack: `delete index_` should come before `mmap_guard_.reset()`?

src/index/sparse/sparse_index_node.cc
```diff
@@ -650,6 +651,462 @@ class SparseInvertedIndexNodeCC : public SparseInvertedIndexNode<T, use_wand> {
...
+        index_ = index.value();
+        binary_ = binary;
+        return index_->Deserialize(reader);
+    }
+
+    Status
+    DeserializeFromFile(const std::string& filename, std::shared_ptr<Config> config) override {
+        if (index_) {
+            LOG_KNOWHERE_WARNING_ << Type() << " already created, deleting old";
+            mmap_guard_.reset();  // unmap before deleting index that references mmap'd memory
```
sparknack: ditto.

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1725 @@
...
+            writer.write(bm25_params_->row_sums_spans_.data(), sizeof(float), n_rows_internal_);
+        }
+
+        // Write max scores per dim
+        if (max_score_in_dim_spans_.size() > 0) {
+            writer.write(max_score_in_dim_spans_.data(), sizeof(float), nr_inner_dims_);
+        }
+
+        // Write prometheus build stats
+#if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOHWERE_WITH_LIGHT)
```
sparknack: `KNOHWERE_WITH_LIGHT` -> `KNOWHERE_WITH_LIGHT`

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1725 @@
...
+        section_meta.emplace_back(DspSectionType::POSTING_LISTS, posting_lists_size);
+        section_meta.emplace_back(DspSectionType::DIM_MAP, sizeof(uint32_t) * nr_inner_dims_);
+
+        if (metric_type_ == SparseMetricType::METRIC_BM25) {
+            section_meta.emplace_back(DspSectionType::ROW_SUMS, sizeof(float) * n_rows_internal_);
+        }
+        if (max_score_in_dim_spans_.size() > 0) {
+            section_meta.emplace_back(DspSectionType::MAX_SCORES_PER_DIM, sizeof(float) * nr_inner_dims_);
+        }
+#if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOHWERE_WITH_LIGHT)
```
sparknack: `KNOHWERE_WITH_LIGHT` -> `KNOWHERE_WITH_LIGHT`

src/index/sparse/sparse_dsp_index.h
```diff
@@ -0,0 +1,1725 @@
...
+            std::sort(spb_by_ub.begin(), spb_by_ub.end(),
+                      [&](uint32_t a, uint32_t b) { return superblock_ub[a] > superblock_ub[b]; });
+
+            std::vector<uint8_t> spb_processed(n_superblocks_, 0);
+
+            const uint32_t batch_sizes[] = {64, 256};
+            const int n_batches = sizeof(batch_sizes) / sizeof(batch_sizes[0]);
+            uint32_t cursor = 0;
+
+            for (int batch_idx = 0; batch_idx <= n_batches && cursor < spb_by_ub.size(); ++batch_idx) {
```
sparknack, line 1637: how about pruning superblock one by one? does it have any performance problem?

src/index/sparse/sparse_dsp_index.h
lyang24: > “erroneously prune enough superblocks that there are not enough documents left to contribute k results”

  > “impose an extra condition guaranteeing top-γ superblocks"
  
  the paper had a warning on gamma = 0 which we set as default and i used a magic number 2 to carry the top 2 pruned super block for defense

src/index/sparse/sparse_dsp_index.h
lyang24: yes performance issues 
With filter_rate=0.8

  Same setup for both runs:

  - MSMARCO BM25
  - topk=10, nq=1000, warmup=1
  - random filter, seed=42
  - only change: batched bootstrap at src/index/sparse/sparse_dsp_index.h:1633 vs strict one-superblock-at-a-time in the filtered path at src/index/sparse/
    sparse_dsp_index.h:1621

  Results:

  - Batched: recall 0.9990, QPS 385.5, mean 2.59 ms, p99 7.49 ms
  - One-by-one: recall 0.9990, QPS 150.6, mean 6.64 ms, p99 13.95 ms

  So one-by-one is about 2.6x slower at the same recall and fill rate. That supports the batching design.

src/index/sparse/sparse_index_node.cc
lyang24: fixed

src/index/sparse/sparse_dsp_index.h
lyang24: fixed

src/index/sparse/sparse_dsp_index.h
lyang24: fixed

jamesgao-jpg:
Hi, when you have time, please rebase your pr from main branch to resolve CI failure problem :)

lyang24:
closed and cherry picked commits to main

Issue #1510, opened 2026-03-17
bug: bad optional access crash in GPU_CAGRA adapt_for_cpu search when ef not provided


marcelo-cjl:
## Bug Description

When using `GPU_CAGRA` index with `adapt_for_cpu=True`, search crashes with `std::bad_optional_access` if the user does not explicitly pass `ef` in search params. Every search call fails — not just the second one.

## Root Cause

In `src/index/gpu_cuvs/gpu_cuvs_cagra.cc:60`, `GpuCuvsCagraHybridIndexNode::Search()`:

```cpp
hnswlib::SearchParam param{(size_t)cagra_cfg.ef.value()};
```

The `ef` field is declared with `allow_empty_without_default()` in `gpu_cuvs_cagra_config.h:118`, so it's an empty `std::optional` when not provided. Calling `.value()` on it throws `std::bad_optional_access`.

CAGRA users have no reason to pass `ef` — it's an HNSW-specific parameter. The `adapt_for_cpu` path silently requires it.

## Steps to Reproduce

```python
from pymilvus import connections, Collection, CollectionSchema, FieldSchema, DataType, utility
import numpy as np

connections.connect(host="localhost", port=19530)

fields = [
    FieldSchema(name="id", dtype=DataType.INT64, is_primary=True, auto_id=True),
    FieldSchema(name="vec", dtype=DataType.FLOAT_VECTOR, dim=128),
]
col = Collection("repro_cagra", CollectionSchema(fields))

vectors = np.random.random((5000, 128)).astype(np.float32)
col.insert([vectors.tolist()])
col.flush()

col.create_index("vec", {
    "index_type": "GPU_CAGRA",
    "metric_type": "IP",
    "params": {"intermediate_graph_degree": 64, "graph_degree": 32, "adapt_for_cpu": True},
})
col.load()

# This crashes with "bad optional access":
col.search(data=vectors[:1].tolist(), anns_field="vec",
           param={"metric_type": "IP", "params": {}}, limit=50)
```

**Error:**
```
MilvusException: Operator::GetOutput failed for [Operator:PhyVectorSearchNode, plan node id: 12] : bad optional access
```

**Verified reproduction output:**
```
--- Search WITHOUT ef (should trigger bad optional access) ---
  Search #1: FAILED — bad optional access
  Search #2: FAILED — bad optional access
  Search #3: FAILED — bad optional access

--- Search WITH ef=64 (workaround) ---
[...]

PR #1511, opened 2026-03-17
fix: default ef in GPU_CAGRA CheckAndAdjust to prevent bad_optional_access

marcelo-cjl:
## Summary
- When using `GPU_CAGRA` with `adapt_for_cpu=true`, searching without an explicit `ef` parameter crashes with `std::bad_optional_access` because `GpuCuvsCagraConfig::CheckAndAdjust` did not provide a default for `ef`
- Add the same defaulting logic as `BaseHnswConfig`: `ef = max(k, 16)`, and validate `ef >= k` when explicitly provided
- Add regression test that searches with `adapt_for_cpu=true` without setting `ef`

Closes https://github.com/zilliztech/knowhere/issues/1510


alexanderguzhva:
issue: #1510 
/kind improvement

PR #1497, opened 2026-03-12
upgrade faiss to 1.14.1

alexanderguzhva:
* Upgrade faiss to 1.14.1+, commit so far is e5114c2368073d00316a866baa33e6595df3c396 (Mar 20 2026)
* deprecate enable_bf16_patch logic
* get rid of the forked `IndexPQ`, `ProductQuantizer`
* get rid of forked `IVFFastScan`-related facilities (such as `pq4`)
* get rid of the forked `simblib`
* get rid of the forked `IndexBinaryHNSW`
* get rid of the forked `partitioning_avx2`
* in process of merging dynamic dispatching mechanisms
* modernize a forked `IVFFastScan` indices to match the baseline
* upgrade the `libfaiss.cmake` file

Contains #1496 
Contains https://github.com/facebookresearch/faiss/pull/4894
Contains https://github.com/facebookresearch/faiss/pull/4926 , which also fixes some problems with the pqfastscan filtering code

PR #1515, opened 2026-03-21
Upgrade knowhere to use C++20

alexanderguzhva:
This is needed for the future versions of Faiss, cuVS, SVS
wait for #1497 before merging
/kind improvement


foxspy (approved):
/lgtm

PR #1516, opened 2026-03-22
update fmt to 11.2.0 in conanfile.py

alexanderguzhva:
otherwise, `clang-21` won't compile the code
wait for #1497 

/kind improvement


foxspy (approved):
/lgtm

PR #1518, opened 2026-03-23
Fix cuVS GPU memory access errors in DiskANN partitioning

### LINKED ISSUE #1519: cuVS GPU kmeans crashes when num_centers > num_train in DiskANN partitioning
## Description

When building DiskANN indexes with cuVS GPU acceleration, the GPU kmeans path crashes with a memory access error if the number of cluster centers exceeds the number of training points (`num_centers > num_train`). This can happen in `generate_pq_pivots`, `partition`, `partition_with_ram_budget`, and `partition_calc_kmeans`.

Additionally, cuVS's `robust_prune.cuh` contains an unused variable that triggers CUDA diagnostic 550, which becomes a build error due to cuVS's `-Werror=all-warnings` flag.

## Expected Behavior

- GPU kmeans should gracefully fall back to the CPU path when `num_centers > num_train`.
- GPU build should compile cleanly without spurious warnings from cuVS headers.

## Actual Behavior

- GPU kmeans crashes with a memory access error.
- GPU build fails due to `-Werror` promoting cuVS's unused-variable warning to an error.

Signed-off-by: jamesgao-jpg <james.gao@zilliz.com>

jamesgao-jpg:
## Summary

Fixes #1519

- **Guard cuVS GPU kmeans calls** against crashing when `num_centers > num_train` by adding a bounds check that falls back to the CPU path. This affects `generate_pq_pivots`, `partition`, `partition_with_ram_budget`, and `partition_calc_kmeans` in DiskANN.
- **Suppress CUDA diagnostic 550** (unused variable in cuVS `robust_prune.cuh`) to prevent it from becoming a build error via cuVS's `-Werror=all-warnings`.

## Test plan

- [x] GPU CI pipeline passes (DiskANN build + search with cuVS enabled)
- [x] CPU-only build unaffected

Signed-off-by: jamesgao-jpg <james.gao@zilliz.com>

### DIFF
--- a/thirdparty/DiskANN/src/partition_and_pq.cpp
+++ b/thirdparty/DiskANN/src/partition_and_pq.cpp
@@ -325,35 +325,40 @@ int generate_pq_pivots(const float *passed_train_data, size_t num_train,
 
   bool used_gpu = false;
 #ifdef KNOWHERE_WITH_CUVS
-  // GPU path — cuVS kmeans crashes when num_centers > num_train, fall back to CPU
-  if(is_gpu_available() && num_train >= num_centers) {
-    raft::resources res;
-    LOG_KNOWHERE_INFO_ << "Running pq with " << num_centers << " clusters, pq "<< num_pq_chunks<< " ...using GPU!";
-    for (size_t i = 0; i < num_pq_chunks; i++) {
-      size_t cur_chunk_size = chunk_offsets[i + 1] - chunk_offsets[i];
-      if (cur_chunk_size == 0)
-        continue;
-      std::unique_ptr<float[]> cur_pivot_data =
-        std::make_unique<float[]>(num_centers * cur_chunk_size);
-      std::unique_ptr<float[]> cur_data =
-        std::make_unique<float[]>(num_train * cur_chunk_size);
-      std::unique_ptr<uint32_t[]> closest_center =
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@jamesgao-jpg the code with the if-branch that checks `num_train >= num_centers` needs to explicitly add to the log the message that the computation switched from GPU to CPU, otherwise we won't be able to diagnose what's going on. Something like the following (not the real code, but rather a train of thoughts)
```C++
if (gpu_used()) {
    if (num_train >= num_centers) {
        LOG_WARNING << "switching to CPU, because of the `num_train >= num_centers` situation";
    } else {
        // go with GPU
    }
}
```

jamesgao-jpg:
> @jamesgao-jpg the code with the if-branch that checks `num_train >= num_centers` needs to explicitly add to the log the message that the computation switched from GPU to CPU, otherwise we won't be able to diagnose what's going on. Something like the following (not the real code, but rather a train of thoughts)
> 
> ```c++
> if (gpu_used()) {
>     if (num_train >= num_centers) {
>         LOG_WARNING << "switching to CPU, because of the `num_train >= num_centers` situation";
>     } else {
>         // go with GPU
>     }
> }
> ```

Fixed logic with added logging if switch to CPU route

alexanderguzhva:
issue: #1519
/kind improvement

PR #1524, opened 2026-03-24
sparse: add SINDI inverted index with SIMD-accelerated window-based search

sparknack:
issue: #1555

SINDI partitions posting lists into fixed-size windows (1024-65535 docs), storing only 16-bit local offsets instead of full 32-bit doc IDs. Search uses a window-level TAAT (Term-At-A-Time) strategy: for each window, scores are scatter-accumulated into a dense buffer via SIMD (AVX-512/AVX2/SVE), then batch-inserted into a top-k heap. Per-dimension window NNZ metadata enables O(1) window skipping. Supports both IP (fp16 quantized values) and BM25 (uint16 term frequencies) scoring.

Based on the SINDI algorithm for sparse vector search. Reference: https://arxiv.org/abs/2509.08395

### DIFF
--- a/src/index/sparse/sindi_inverted_index.h
+++ b/src/index/sparse/sindi_inverted_index.h
@@ -8,6 +8,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
 #include <memory>
 #include <numeric>
 #include <queue>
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@sparknack it seems that another rebase is needed...

alexanderguzhva:
/kind improvement
issue: #1555

PR #1533, opened 2026-03-25
fix: initialize nearest_idx to -1 in k=1 L2 fast path

CLiqing:
issue: https://github.com/zilliztech/knowhere/issues/1538
/kind improvement

When all L2 distances overflow to infinity (e.g., max-float vectors), exhaustive_L2sqr_nearest_imp returns nearest_idx=0 (initial value) because no distance satisfies `dis < HUGE_VALF`. This is inconsistent with HeapBlockResultHandler which initializes ids to -1 via heap_heapify, correctly signaling "no valid result".

Change nearest_idx from size_t(0) to int64_t(-1) to match the general path behavior.


alexanderguzhva:
looks good, waiting for a CI

alexanderguzhva:
@CLiqing please rebase on top of master to get rid of this `e2e(arm)` error

Issue #1534, opened 2026-03-25
HNSW RangeSearch bypasses brute-force fallback at high filter ratios


foxspy:
## Description

HNSW `RangeSearch` always delegates to the iterator-based path for all common data types (fp32, fp16, bf16, int8, bin1) because `is_ann_iterator_supported()` returns `true`. This means the brute-force fallback logic in `WhetherPerformBruteForceRangeSearch()` (threshold 0.97) is never reached — it is effectively dead code.

At high filter ratios (e.g., 99%), the iterator walks the HNSW graph and visits enormous numbers of filtered-out nodes, causing extreme performance degradation.

## Benchmark (cohere 1M, 768d, COSINE, range_filter=1.0, radius=0.85)

| Scenario | Method | QPS | Recall |
|----------|--------|-----|--------|
| 0% filter | HNSW iterator | 540 | 0.998 |
| **99% filter** | **HNSW iterator** | **18** | 1.0 |
| 99% filter | Brute force | 4037 | 1.0 |

**99% filter with iterator path is 42x slower than brute force.**

## Root Cause

In `faiss_hnsw.cc`, `BaseFaissRegularIndexHNSWNode::RangeSearch()`:

```cpp
// Line 1578-1581
if (is_ann_iterator_supported()) {
    return IndexNode::RangeSearch(dataset, std::move(cfg), bitset_, op_context);
}
// WhetherPerformBruteForceRangeSearch() at line 1633 is never reached
```

The check `is_ann_iterator_supported()` returns `true` for all common data types, causing an early return before the brute-force threshold check.

## Expected Behavior

`WhetherPerformBruteForceRangeSearch()` should be evaluated before falling back to the iterator path, so that high-filter-ratio range searches use the brute-force path (IndexBruteForceWrapper) instead of the iterator.

## Scope & Limitations

This fix only addresses `RangeSearch`. Other operations that directly use the HNSW iterator (e.g., groupby search) still suffer from the same high-filter-ratio performance issue.

The fundamental problem is that the HNSW iterator always traverses the graph even when most nodes are filtered out. Under high filter ratios, the iterator should fall back to bitset-based sequential scan instead of graph traversal. This broader optimization will be addressed separately in a follow-up.

PR #1536, opened 2026-03-25
fix: [2.6 cherry-pick] check brute-force threshold before iterator path in HNSW RangeSearch

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/1534

Cherry-pick of #1535 onto v2.6.10 for the 2.6 release line.

## Summary

- HNSW `RangeSearch` always used the iterator path for all common data types, bypassing the brute-force fallback (`WhetherPerformBruteForceRangeSearch`, threshold 0.97)
- At high filter ratios (e.g. 99%), the iterator visits too many filtered-out graph nodes, causing 42x slowdown vs brute force
- This fix checks `WhetherPerformBruteForceRangeSearch()` before the iterator path, falling through to the brute-force path when filter ratio >= 97%

## Benchmark (cohere 1M, 768d, COSINE, filter_ratio=0.99)

| Method | QPS | Recall |
|--------|-----|--------|
| HNSW RangeSearch (before fix) | 18 | 1.0 |
| **HNSW RangeSearch (after fix)** | **3166** | **1.0** |
| Brute force RangeSearch (reference) | 4037 | 1.0 |

### DIFF
--- a/src/index/hnsw/faiss_hnsw.cc
+++ b/src/index/hnsw/faiss_hnsw.cc
@@ -1565,8 +1565,19 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
     expected<DataSetPtr>
     RangeSearch(const DataSetPtr dataset, std::unique_ptr<Config> cfg, const BitsetView& bitset_,
                 milvus::OpContext* op_context) const override {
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@foxspy waiting for a rebase

Issue #1532, opened 2026-03-25
enhance: IVFRaBitQ - add opt-in FastScan backend for current 1-bit search path


lyang24:
`IVFRaBitQ` in `knowhere` currently uses the older per-vector scan path. This leaves a clear performance gap versus the newer block-based FastScan implementation already available in vendored Faiss.

  This issue proposes a small, mergeable first step: add an opt-in FastScan backend for the existing 1-bit `IVFRaBitQ` path, while keeping current behavior as the default.

alexanderguzhva:
@lyang24 it will be possible to expose after #1497 lands

PR #1542, opened 2026-03-26
feat: add IndexIVFRaBitQFastScan

lyang24:
Add rbq_use_fastscan config flag (default false) that enables the upstream faiss::IndexIVFRaBitQFastScan SIMD batch backend for 1-bit IVFRaBitQ. The legacy backend remains the default and is unchanged.

Explicitly unsupported on FastScan in this version:
- Per-search rbq_bits_query > 0 (rejected at search/range_search)
- AnnIterator (rejected at runtime)

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -16,6 +16,9 @@
 #include <faiss/cppcontrib/knowhere/IndexScaNN.h>
 
 #include "common/metric.h"
+#include "faiss/IndexIVFRaBitQ.h"
+#include "faiss/IndexIVFRaBitQFastScan.h"
+#include "faiss/IndexRefine.h"
 #include "faiss/VectorTransform.h"
 #include "faiss/cppcontrib/knowhere/IndexBinaryFlat.h"
 #include "faiss/cppcontrib/knowhere/IndexBinaryIVF.h"
@@ -26,10 +29,12 @@
 #include "faiss/cppcontrib/knowhere/IndexIVFRaBitQ.h"
 #include "faiss/cppcontrib/knowhere/IndexScalarQuantizer.h"
 #include "faiss/cppcontrib/knowhere/index_io.h"
+#include "faiss/index_io.h"
 #include "index/clustering_config.h"
 #include "index/data_view_dense_index/index_node_with_data_view_refiner.h"
 #include "index/ivf/ivf_config.h"
 #include "index/ivf/ivf_wrapper.h"
+#include "index/ivf/ivfrbq_fastscan_wrapper.h"
 #include "index/ivf/ivfrbq_wrapper.h"
 #include "io/memory_io.h"
 #include "knowhere/bitsetview_idselector.h"
@@ -71,7 +76,8 @@ class IvfIndexNode : public IndexNode {
                           std::is_same<IndexType, faiss::cppcontrib::knowhere::IndexBinaryIVF>::value ||
                           std::is_same<IndexType, faiss::cppcontrib::knowhere::IndexScaNN>::value ||
                           std::is_same<IndexType, faiss::cppcontrib::knowhere::IndexIVFScalarQuantizerCC>::value ||
-                          std::is_same<IndexType, IndexIVFRaBitQWrapper>::value,
+                          std::is_same<IndexType, IndexIVFRaBitQWrapper>::value ||
+                          std::is_same<IndexType, IndexIVFRaBitQFastScanWrapper>::value,
                       "not support");
         static_assert(std::is_same_v<DataType, fp32> || std::is_same_v<DataType, bin1>,
                       "IvfIndexNode only support float/binary");
@@ -178,7 +184,8 @@ class IvfIndexNode : public IndexNode {
         if constexpr (std::is_same<faiss::cppcontrib::knowhere::IndexBinaryIVF, IndexType>::value) {
             return true;
         }
-        if constexpr (std::is_same<IndexIVFRaBitQWrapper, IndexType>::value) {
+        if constexpr (std::is_same<IndexIVFRaBitQWrapper, IndexType>::value ||
+                      std::is_same<IndexIVFRaBitQFastScanWrapper, IndexType>::value) {
             return false;
         }
         return false;
@@ -249,6 +256,9 @@ class IvfIndexNode : public IndexNode {
         if constexpr (std::is_same<IndexIVFRaBitQWrapper, IndexType>::value) {
             return std::make_unique<IvfRaBitQConfig>();
         }
+        if constexpr (std::is_same<IndexIVFRaBitQFastScanWrapper, IndexType>::value) {
+            return std::make_unique<IvfRaBitQFastScanConfig>();
+        }
     };
 
     std::unique_ptr<BaseConfig>
@@ -304,6 +314,9 @@ class IvfIndexNode : public IndexNode {
         if constexpr (std::is_same<IndexType, IndexIVFRaBitQWrapper>::value) {
             return index_->size();
         }
+        if constexpr (std::is_same<IndexType, IndexIVFRaBitQFastScanWrapper>::value) {
+            return index_->size();
+        }
     };
     int64_t
     Count() const override {
@@ -338,6 +351,9 @@ class IvfIndexNode : public IndexNode {
         if constexpr (std::is_same<IndexType, IndexIVFRaBitQWrapper>::value) {
             return knowhere::IndexEnum::INDEX_FAISS_IVFRABITQ;
         }
+        if constexpr (std::is_same<IndexType, IndexIVFRaBitQFastScanWrapper>::value) {
+            return knowhere::IndexEnum::INDEX_FAISS_IVFRABITQ_FASTSCAN;
+        }
     };
 
  private:
@@ -359,7 +375,8 @@ class IvfIndexNode : public IndexNode {
         return std::is_same_v<IndexType, IndexIVFPQWrapper> || std::is_same_v<IndexType, IndexIVFSQWrapper> ||
                std::is_same_v<IndexType, faiss::cppcontrib::knowhere::IndexIVFScalarQuantizerCC> ||
                std::is_same_v<IndexType, faiss::cppcontrib::knowhere::IndexScaNN> ||
-               std::is_same_v<IndexType, IndexIVFRaBitQWrapper>;
+               std::is_same_v<IndexType, IndexIVFRaBitQWrapper> ||
+               std::is_same_v<IndexType, IndexIVFRaBitQFastScanWrapper>;
     }
 
  private:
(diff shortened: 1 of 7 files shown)


lyang24:
Result on cohere 1M, k=10, nlist=1024, IP, refine=false:

  | nprobe | legacy recall | fastscan recall | legacy avg ms | fastscan avg ms | speedup |
  |---|---:|---:|---:|---:|---:|
  | 16  | 0.538 | 0.572 | 1.749 | 1.602 | 1.09x |
  | 32  | 0.568 | 0.592 | 2.509 | 2.040 | 1.23x |
  | 64  | 0.596 | 0.608 | 3.967 | 3.142 | 1.26x |
  | 128 | 0.596 | 0.624 | 7.110 | 5.380 | 1.32x |
  | 256 | 0.604 | 0.626 | 13.928 | 10.145 | 1.37x |
  
  fastscan is will be faster on 1m and plus datasets - larger datasets amortize set up costs.
  but this recall is unusable, need to to impl refine. (maybe refine intergration could be a follow up pr?)

lyang24:
ok wired in refines the recalls are acceptable for ivf at 4k refine the gain is consistent

 | nprobe | legacy recall | fastscan recall | legacy avg ms | fastscan avg ms | speedup |
  |---|---:|---:|---:|---:|---:|
  | 16  | 0.724 | 0.762 | 1.845 | 1.639 | 1.13x |
  | 32  | 0.800 | 0.834 | 2.608 | 2.152 | 1.21x |
  | 64  | 0.872 | 0.886 | 4.070 | 3.255 | 1.25x |
  | 128 | 0.906 | 0.908 | 7.240 | 5.335 | 1.36x |
  | 256 | 0.930 | 0.928 | 14.093 | 9.917 | 1.42x |

jamesgao-jpg:
Hi, we just committed a fix to our CI testing pipeline with https://github.com/zilliztech/knowhere/pull/1540. I saw your current pipelines are failing due to that problem, so you may need to rebase to pick up that fix first :)

alexanderguzhva:
@lyang24 Thank you for your code! 
Would it be possible to introduce a new index type instead of adding a new flag `rbq_use_fastscan`? Similarly, a new class type, such a `IndexIVFRaBitQFastScanWrapper`. Maybe, separate unit test cases as well, in order to blend the vanilla RaBitQ and the FastScan one. It will make the code way cleaner

alexanderguzhva:
@lyang24 filtering in the baseline FastScan should be supported after the `https://github.com/facebookresearch/faiss/pull/4926`. So, please take a look if removing checks for `sel == nullptr` in the baseline `Faiss` code for IVFRaBitQFastScan helps and makes the filtering work. If so, I'll be able to write a PR into the baseline `Faiss` and/or propagate yours quickly. 
This could be done in a separate PR, if reasonable. Or I can do it on my own.

tests/ut/test_search.cc
```diff
@@ -1106,3 +1106,270 @@ TEST_CASE("Test RangeSearch Cancellation", "[range_search][cancellation]") {
         REQUIRE(results2.has_value());
     }
 }
+
+TEST_CASE("Test IVFRaBitQ FastScan Backend", "[ivfrabitq_fastscan]") {
+    const int64_t nb = 1000, nq = 10;
+    const int64_t dim = 128;
+    const int64_t topk = 10;
+
+    auto metric = GENERATE(as<std::string>{}, knowhere::metric::L2, knowhere::metric::COSINE);
```
alexanderguzhva, line 1115: how does a COSINE distance passes this unit test, given that the description of the PR says that COSINE is not supported? Please check whether it works or whether a test passes when it should not :)

lyang24: well ivf is a special case knowhere normalize cosine to ip when pass to faiss. The pr description is not accurate. i will modify it faiss does not support cosine but knowhere does
https://github.com/zilliztech/knowhere/blob/fd532fbaf35f9db5f8e5a1acc4dee41c21bca4a3/src/index/ivf/ivf.cc#L1000

lyang24:
> @lyang24 Thank you for your code! Would it be possible to introduce a new index type instead of adding a new flag `rbq_use_fastscan`? Similarly, a new class type, such a `IndexIVFRaBitQFastScanWrapper`. Maybe, separate unit test cases as well, in order to blend the vanilla RaBitQ and the FastScan one. It will make the code way cleaner

refactored

lyang24:
> This could be done in a separate PR, if reasonable. Or I can do it on my own.

look like removing the check still have some follow up work to do maybe a follow up pr would be better
- thread const IDSelector* sel through make_knn_scanner
- add sel to rabitq_ivf_make_knn_scanner[_impl]
- add sel to IVFRaBitQHeapHandler
- pass it into ResultHandlerCompare(..., sel)

alexanderguzhva:
> > This could be done in a separate PR, if reasonable. Or I can do it on my own.
> 
> look like removing the check still have some follow up work to do maybe a follow up pr would be better
> 
>     * thread const IDSelector* sel through make_knn_scanner
> 
>     * add sel to rabitq_ivf_make_knn_scanner[_impl]
> 
>     * add sel to IVFRaBitQHeapHandler
> 
>     * pass it into ResultHandlerCompare(..., sel)

@lyang24 agreed, this can be implemented in further PRs. But please
* restore various comments that were removed in the code, they were added for a reason
* remove the introduced `size_cache_` completely, because the `size()` function evaluation is not expected to be called often
* squash your commits into a single one. 

As soon as it is done, I'll be ready to accept this.

alexanderguzhva:
/hold
until requested changes are implemented

lyang24:
> > > This could be done in a separate PR, if reasonable. Or I can do it on my own.
> > 
> > 
> > look like removing the check still have some follow up work to do maybe a follow up pr would be better
> > ```
> > * thread const IDSelector* sel through make_knn_scanner
> > 
> > * add sel to rabitq_ivf_make_knn_scanner[_impl]
> > 
> > * add sel to IVFRaBitQHeapHandler
> > 
> > * pass it into ResultHandlerCompare(..., sel)
> > ```
> 
> @lyang24 agreed, this can be implemented in further PRs. But please
> 
> * restore various comments that were removed in the code, they were added for a reason
> * remove the introduced `size_cache_` completely, because the `size()` function evaluation is not expected to be called often
> * squash your commits into a single one.
> 
> As soon as it is done, I'll be ready to accept this.

all fixed :)

src/index/ivf/ivf.cc
```diff
@@ -938,6 +972,16 @@ IvfIndexNode<DataType, IndexType>::Search(const DataSetPtr dataset, std::unique_
         }
     }
 
+    if constexpr (std::is_same_v<IndexType, IndexIVFRaBitQFastScanWrapper>) {
+        if (!bitset.empty()) {
+            // Upstream IndexIVFRaBitQFastScan::make_knn_scanner still drops the
+            // IDSelector, so reject filtered search up front to avoid returning
+            // incorrect results.
+            return expected<DataSetPtr>::Err(Status::invalid_args,
```
foxspy: This could pose a significant issue as filtering is a core use case. If an index is created but cannot be queried with expressions, it will be difficult to justify to our users. We should consider adding bitset support; please let us know if you need any assistance, as we are happy to help.

foxspy (changes requested):

src/index/ivf/ivf.cc
alexanderguzhva: @lyang24 @foxspy please correct me if I'm wrong, but this is a temporary code until a support of filtering is added into ivfrabitqfastscan in Faiss, correct?

foxspy: @alexanderguzhva Yes, if filtering is not supported, we won't be able to integrate this index into Milvus. Since filtering is a fundamental requirement, its absence would lead to unexpected behavior and significant overhead in terms of user education and support.

lyang24: yea i am ok with putting this on hold when faiss side is ready and then integrate with filters

lyang24: oh the faiss github display confused me i though the dependent pr for filter is closed but closed actually means merged :) will impl filters by change vendored faiss code

lyang24:
cohere 100k Setup:

  - dataset: /home/ubuntu/data/cohere-wiki-1024-ip.hdf5
  - nb=100000, nq=200
  - k=1000, metric=IP
  - nlist=1024, nprobe=64, refine_k=1.0
  - filter sweep: 0.0, 0.2, 0.4, 0.6, 0.8, 0.9

  Result: filtering works on IVF_RABITQ_FASTSCAN, but fastscan is slower than regular IVF_RABITQ in this configuration, with effectively identical recall/fill
  at every point.

  Filter ratio, Latency and recall:

  - 0.0: regular 2.02 ms, fastscan 3.33 ms, recall both 0.654
  - 0.2: regular 1.97 ms, fastscan 3.36 ms, recall both 0.644
  - 0.4: regular 1.66 ms, fastscan 3.05 ms, recall both 0.624
  - 0.6: regular 1.83 ms, fastscan 3.31 ms, recall both 0.591
  - 0.8: regular 1.40 ms, fastscan 2.90 ms, recall both 0.491
  - 0.9: regular 1.49 ms, fastscan 2.99 ms, recall both 0.363, avg valid hits both 665

  Build time:

  - regular 22.23 s
  - fastscan 22.49 s

lyang24:
1 M cohere so filtering is slower for fast scan but the impl is correct

Ran the 1M sweep on 50.18.128.112 with the same harness, after changing scripts/bench_ivf_rabitq_fastscan_filter_sweep.py to build the two indexes
  sequentially so the remote host would not OOM.

  Setup:

  - dataset: /home/ubuntu/data/cohere-wiki-1024-ip.hdf5
  - nb=1,000,000, nq=200
  - k=1000, metric=IP
  - nlist=1024, nprobe=64, refine_k=1.0
  - filter ratios: 0.0, 0.2, 0.4, 0.6, 0.8, 0.9

  Results:

  - 0.0: regular 5.34 ms, fastscan 4.39 ms, recall both 0.693
  - 0.2: regular 4.88 ms, fastscan 4.64 ms, recall both 0.690
  - 0.4: regular 4.37 ms, fastscan 4.75 ms, recall both 0.691
  - 0.6: regular 3.71 ms, fastscan 4.65 ms, recall both 0.687
  - 0.8: regular 2.99 ms, fastscan 4.44 ms, recall both 0.674
  - 0.9: regular 2.56 ms, fastscan 4.22 ms, recall both 0.653

  Build time:

  - regular 109.05 s
  - fastscan 109.75 s

alexanderguzhva:
Hi @lyang24 I've seen your https://github.com/facebookresearch/faiss/pull/5026 . Let me upgrade the our faiss version for you could proceed

lyang24:
> Collaborator

thank you!

alexanderguzhva:
@lyang24 please rebase to master, #1558 has been merged. Thanks!

lyang24:
> @lyang24 please rebase to master, #1558 has been merged. Thanks!

this is done

alexanderguzhva:
@lyang24 looks good to me. What are the further steps?
@foxspy any concerns about the requested change?

lyang24:
> @lyang24 looks good to me. What are the further steps? @foxspy any concerns about the requested change?

for me the next step would be merge this - i had some ideas pending on this change to merge

foxspy (approved):
/lgtm

alexanderguzhva:
issue: #1532 
/kind improvement

PR #1556, opened 2026-04-03
fix: initialize nearest_idx to -1 in k=1 L2 fast path

CLiqing:
When all L2 distances overflow to infinity (e.g. max-float vectors), the loop in exhaustive_L2sqr_nearest_imp never updates nearest_idx, returning 0 instead of -1 (no valid result).

Backport of zilliztech/knowhere#1533 to 2.6.


liliu-z (commented):
/lgtm
/approve

Issue #1570, opened 2026-04-09
IndexStaticFaced::ConfigCheck dispatch silently broken since #888 — staticConfigCheckMap is never populated


foxspy:
## Description

`IndexStaticFaced::ConfigCheck` (introduced in #888) is supposed to dispatch to per-index `StaticConfigCheck` implementations to validate `(indexType, config)` combinations before building. The dispatch has been silently dead since the feature merged: `staticConfigCheckMap` is never populated for any registered index, so the per-index checks never run.

Root cause is a type mismatch in the SFINAE probe that gates registration:

- `IndexStaticFaced::InternalConfigCheck` is declared with `(const BaseConfig&, const IndexVersion&, std::string&)` (`include/knowhere/index/index_static.h:134`).
- All real `StaticConfigCheck` implementations take `(const Config&, PARAM_TYPE, std::string&)` (e.g. `src/index/ivf/ivf.cc:138`, `src/index/hnsw/hnsw.h:94`).

`PARAM_TYPE` is an unscoped enum, and `int32_t` (`IndexVersion`) does not implicitly convert to it, so `has_static_StaticConfigCheck<...>::value` evaluates to `false` for every index. The `if constexpr` block in `RegisterStaticFunc` (`index_static.h:114-117`) is therefore dead code, `staticConfigCheckMap` stays empty, and `ConfigCheck` always falls through to `return Status::success` at `index_static.cc:63`.

## Expected Behavior

`IndexStaticFaced<fp32>::ConfigCheck("IVF_FLAT", version, {"metric_type":"HAMMING","nlist":128}, msg)` should return `Status::invalid_metric_type` (since IVF float indexes only accept L2/IP/COSINE).

## Actual Behavior

Returns `Status::success`. The metric/data-type validation defined in `IvfIndexNode::StaticConfigCheck` and `HnswIndexNode::StaticConfigCheck` never executes. Build-time pre-check is a no-op for any caller.

## Impact

- All callers (e.g. Milvus pre-build validation) get false-positive success on invalid config combinations. The non-fatal layers downstream still catch most cases at build time, but later and with worse error messages.
- The `tests/ut/test_config.cc::checkBuildConfig` helper only asserts `Status::success`, so the bug never surfaced in CI.
- Bug has been silent since #888 (2024-10-15).

PR #1571, opened 2026-04-09
fix: wire up IndexStaticFaced::ConfigCheck dispatch

foxspy:
issue: #1570

## Summary

- Align `InternalConfigCheck` signature to take `PARAM_TYPE` so the SFINAE probe in `RegisterStaticFunc` actually matches every index's real `StaticConfigCheck` impl. Previously the probe used `IndexVersion` (`int32_t`), which does not implicitly convert to the unscoped `PARAM_TYPE` enum, so `staticConfigCheckMap` was permanently empty and `IndexStaticFaced::ConfigCheck` always fell through to `Status::success`.
- Pass `PARAM_TYPE::TRAIN` at the dispatch site (`index_static.cc:60`).
- Companion fix in `IvfIndexNode::StaticConfigCheck` and `HnswIndexNode::StaticConfigCheck`: classify the float vs binary metric branch with `!std::is_same_v<DataType, knowhere::bin1>` instead of `KnowhereFloatTypeCheck<DataType>`. The latter excludes `int8/fp16/bf16`, but those are mocked to `fp32` at runtime via `MockData<>`, so they should accept float metrics — without this companion change, fixing the dispatch alone would regress `IndexStaticFaced<int8>::ConfigCheck("IVF_FLAT", L2)`, which `tests/ut/test_config.cc::checkBuildConfig` already exercises.
- Add negative + positive cases in `tests/ut/test_config.cc` that pin the dispatch path: `fp32+IVF_FLAT+HAMMING` → `invalid_metric_type`, `bin1+BIN_IVF_FLAT+HAMMING` → `success`, `int8+IVF_FLAT+L2` → `success`, `int8+IVF_FLAT+JACCARD` → `invalid_metric_type`.

## Test Plan

- [x] `pre-commit run` clean on all 5 changed files
- [x] Reproduced bug locally before the fix: `Test config json parse / static ConfigCheck rejects metric/data-type mismatch` aborts at the negative HAMMING CHECK with `Check failed: ConfigCheck(IVF_FLAT, v, {HAMMING}) == invalid_metric_type`
- [x] After the fix, the same section passes; full `Test config json parse` test case stays green (no regression on the existing int8 + L2 path)

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -139,23 +139,29 @@ class IvfIndexNode : public IndexNode {
         auto ivf_cfg = static_cast<const IvfConfig&>(cfg);
 
         if (paramType == PARAM_TYPE::TRAIN) {
+            // Decompose compound metrics (MAX_SIM_*, DTW_*) used by emb-list IVF variants down
+            // to their underlying base metric so the float/binary classification still works.
+            // For plain L2/IP/COSINE/HAMMING/JACCARD this returns nullopt and we fall back to
+            // the raw value.
+            const auto& raw_metric = ivf_cfg.metric_type.value();
+            const auto& base_metric = get_sub_metric_type(raw_metric).value_or(raw_metric);
             // Note: int8/fp16/bf16 are mocked to fp32 at runtime via MockData<>, so they only
             // accept float metrics. The only data type that should take the binary branch is bin1.
             if constexpr (!std::is_same_v<DataType, knowhere::bin1>) {
-                if (IsMetricType(ivf_cfg.metric_type.value(), metric::L2) ||
-                    IsMetricType(ivf_cfg.metric_type.value(), metric::IP) ||
-                    IsMetricType(ivf_cfg.metric_type.value(), metric::COSINE)) {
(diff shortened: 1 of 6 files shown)


src/index/ivf/ivf.cc
```diff
@@ -139,7 +139,9 @@ class IvfIndexNode : public IndexNode {
         auto ivf_cfg = static_cast<const IvfConfig&>(cfg);
 
         if (paramType == PARAM_TYPE::TRAIN) {
-            if constexpr (KnowhereFloatTypeCheck<DataType>::value) {
+            // Note: int8/fp16/bf16 are mocked to fp32 at runtime via MockData<>, so they only
+            // accept float metrics. The only data type that should take the binary branch is bin1.
+            if constexpr (!std::is_same_v<DataType, knowhere::bin1>) {
```
SpadeA-Tang, line 150: This method is not reachable before. Now it will reject MAX_SIM* metrics and we should fix it.

alexanderguzhva: @foxspy any comments?

foxspy: fixed

SpadeA-Tang (commented):
/lgtm

marcelo-cjl (approved):

PR #1573, opened 2026-04-09
fix: [2.6] wire up IndexStaticFaced::ConfigCheck dispatch

foxspy:
issue: #1570

Cherry-pick of #1571 to 2.6.

## Summary

- Align `InternalConfigCheck` signature to take `PARAM_TYPE` so the SFINAE probe in `RegisterStaticFunc` actually matches every index's real `StaticConfigCheck` impl. Previously the probe used `IndexVersion` (`int32_t`), which does not implicitly convert to the unscoped `PARAM_TYPE` enum, so `staticConfigCheckMap` was permanently empty and `IndexStaticFaced::ConfigCheck` always fell through to `Status::success`.
- Pass `PARAM_TYPE::TRAIN` at the dispatch site (`index_static.cc:60`).
- Companion fix in `IvfIndexNode::StaticConfigCheck` and `HnswIndexNode::StaticConfigCheck`: classify the float vs binary metric branch with `!std::is_same_v<DataType, knowhere::bin1>` instead of `KnowhereFloatTypeCheck<DataType>`. The latter excludes `int8/fp16/bf16`, but those are mocked to `fp32` at runtime via `MockData<>`, so they should accept float metrics — without this companion change, fixing the dispatch alone would regress `IndexStaticFaced<int8>::ConfigCheck("IVF_FLAT", L2)`, which `tests/ut/test_config.cc::checkBuildConfig` already exercises.
- Add negative + positive cases in `tests/ut/test_config.cc` that pin the dispatch path: `fp32+IVF_FLAT+HAMMING` → `invalid_metric_type`, `bin1+BIN_IVF_FLAT+HAMMING` → `success`, `int8+IVF_FLAT+L2` → `success`, `int8+IVF_FLAT+JACCARD` → `invalid_metric_type`.

## Test Plan

- [x] `pre-commit run` clean on all 5 changed files
- [x] The corresponding main-branch PR (#1571) was verified end-to-end locally: buggy reproduction abort at the negative HAMMING CHECK, fixed state passes; this 2.6 cherry-pick applies the same patch on a byte-identical base.

### DIFF
--- a/include/knowhere/index/index_static.h
+++ b/include/knowhere/index/index_static.h
@@ -131,7 +131,7 @@ class IndexStaticFaced {
     InternalStaticHasRawData(const knowhere::BaseConfig& config, const IndexVersion& version);
 
     static knowhere::Status
-    InternalConfigCheck(const knowhere::BaseConfig& config, const IndexVersion& version, std::string& msg);
+    InternalConfigCheck(const knowhere::BaseConfig& config, PARAM_TYPE paramType, std::string& msg);
 
     static std::unique_ptr<BaseConfig>
     InternalStaticCreateConfig();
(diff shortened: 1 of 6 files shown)


alexanderguzhva:
wait for #1571

marcelo-cjl (approved):

PR #1577, opened 2026-04-10
Add index-type prometheus latency metrics

sparknack:
issue: #1579

- add Prometheus latency metrics labeled by index_type for Knowhere
- keep the metric names as build_latency, load_latency, search_latency, and range_search_latency
- preserve Cardinals existing metric compatibility and make the metric cache concurrency-safe

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -573,7 +573,7 @@ class Config {
     }
 };
 
-#define KNOHWERE_DECLARE_CONFIG(CONFIG) CONFIG()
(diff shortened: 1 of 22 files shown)


alexanderguzhva:
@sparknack I'd like to confirm that objects are not leaked. It seems that new objects are introduced on per-index-type basis, though

sparknack:
> @sparknack I'd like to confirm that objects are not leaked. It seems that new objects are introduced on per-index-type basis, though

it's not a leak, `index_type` is bounded by the enum/string set used by callers.

alexanderguzhva:
/kind improvement
issue: #1579

PR #1421, opened 2026-01-07
Add fallback and uts for sq4u

CLiqing:
issue: https://github.com/zilliztech/knowhere/issues/1428
/kind improvement

Bugfix and add uts

PR #1586, opened 2026-04-16
fix: add missing simdwidth to RVV Similarity classes for SQ4U fallback

CLiqing:
issue: https://github.com/zilliztech/knowhere/issues/1587
/kind bug

- `SimilarityL2_rvv<0>` and `SimilarityIP_rvv<0>` are missing `static constexpr int simdwidth`, causing compilation failure on riscv64 when the `QT_4bit_uniform` case in `select_distance_computer_rvv` falls back to base `select_distance_computer` (which requires `Sim::simdwidth`)
- Added `simdwidth = 1` to both classes — value is 1 (scalar) because the fallback `DistanceComputerSQ4UByte` uses scalar quantization internally
- Issue introduced in 55c18bb ("Add fallback and uts for sq4u #1421")

## Compilation error (before fix)

```
ScalarQuantizerCodec.h:744:36: error: 'simdwidth' is not a member of 'faiss::SimilarityL2_rvv<0>'
  744 |     constexpr int SIMDWIDTH = Sim::simdwidth;
```


foxspy (approved):
/lgtm

Issue #1587, opened 2026-04-16
fix: RVV Similarity classes missing simdwidth causes compilation failure on riscv64


CLiqing:
## Description

Compiling knowhere on riscv64 (SG2044 server) fails with:

```
ScalarQuantizerCodec.h:744:36: error: 'simdwidth' is not a member of 'faiss::SimilarityL2_rvv<0>'
  744 |     constexpr int SIMDWIDTH = Sim::simdwidth;
```

## Root Cause

Commit 55c18bb ("Add fallback and uts for sq4u #1421") added a `QT_4bit_uniform` case in `select_distance_computer_rvv` that falls back to the base `select_distance_computer`. The base function requires `Sim::simdwidth`, but `SimilarityL2_rvv<0>` and `SimilarityIP_rvv<0>` only define `metric_type`, missing the `simdwidth` member.

## Steps to Reproduce

Build knowhere on riscv64, or cross-compile with `riscv64-linux-gnu-g++`:

```bash
riscv64-linux-gnu-g++-12 -std=c++17 -c ScalarQuantizerDC_rvv.cpp
```

## Expected Behavior

Compilation succeeds on riscv64.

maxiaohong2333:
On version 2.6.11, I added fix in https://github.com/zilliztech/knowhere/pull/1586.
But I still got compile failure :
I work on  SG2044 + gcc_14.2.0 + "-std=c++17 -march=rv64gcv_zvfhmin".
It won't happen if I use :
        case ScalarQuantizer::QT_4bit_uniform:
            return new DCTemplate_rvv<
                    QuantizerTemplate_rvv<
                            Codec4bit_rvv,
                            QuantizerTemplateScaling::UNIFORM,
                            0>,
                    Similarity,
                    0>(dim, trained);

The build error log:
[build.txt](https://github.com/user-attachments/files/26780070/build.txt)

maxiaohong2333:
The build had passed after PR1586, thank you for your help.
https://github.com/zilliztech/knowhere/pull/1586l

alexanderguzhva:
@maxiaohong2333 would you please also check https://github.com/zilliztech/knowhere/pull/1594, if you have time, of course ? I'm planning to donate the RVV SQ code to vanilla faiss and rely on its Scalar Quantization facilities

PR #1588, opened 2026-04-16
fix: use base scalar Similarity types in RVV SQ4U fallback path

CLiqing:
Follow-up fix for #1586. The previous fix added `simdwidth` to the RVV Similarity tag types, but the `QT_4bit_uniform` fallback path in `select_distance_computer_rvv` still passed `SimilarityL2_rvv<0>` / `SimilarityIP_rvv<0>` directly to the base `select_distance_computer`, causing compilation failure on riscv64.

**Fix**:
1. Map RVV tag types to their base scalar equivalents before calling `select_distance_computer`:
   - `SimilarityL2_rvv<0>` (metric_type == METRIC_L2) → `SimilarityL2<1>`
   - `SimilarityIP_rvv<0>` (metric_type == METRIC_INNER_PRODUCT) → `SimilarityIP<1>`
2. Revert the incomplete fix from #1586 (remove `simdwidth` from RVV tag types), since the RVV code path never reads `Similarity::simdwidth` — keeping tag types minimal and consistent with their pure-tag design.

issue: https://github.com/zilliztech/knowhere/issues/1587

## Compilation errors before this fix (on riscv64 with v2.6.11 + #1586)

```
ScalarQuantizerCodec.h:659:20: error: no matching function for call to 'SimilarityIP_rvv<0>::SimilarityIP_rvv(std::nullptr_t)'
ScalarQuantizerCodec.h:660:13: error: 'struct SimilarityIP_rvv<0>' has no member named 'begin'
ScalarQuantizerCodec.h:664:17: error: 'struct SimilarityIP_rvv<0>' has no member named 'add_component_2'
ScalarQuantizerCodec.h:666:20: error: 'struct SimilarityIP_rvv<0>' has no member named 'result'
ScalarQuantizerCodec.h:648:20: error: no matching function for call to 'SimilarityL2_rvv<0>::SimilarityL2_rvv(const float*&)'
ScalarQuantizerCodec.h:649:13: error: 'struct SimilarityL2_rvv<0>' has no member named 'begin'
...
```

### DIFF
--- a/thirdparty/faiss/faiss/cppcontrib/knowhere/impl/ScalarQuantizerCodec_rvv.h
+++ b/thirdparty/faiss/faiss/cppcontrib/knowhere/impl/ScalarQuantizerCodec_rvv.h
@@ -403,7 +403,6 @@ template <int SIMDWIDTH>
 struct SimilarityL2_rvv {};
 template <>
 struct SimilarityL2_rvv<0> {
-    static constexpr int simdwidth = 1;
     static constexpr MetricType metric_type = METRIC_L2;
 };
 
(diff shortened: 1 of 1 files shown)


foxspy (approved):
/lgtm

Issue #1580, opened 2026-04-13
use conan 2.x to manage dependencies


yhmo:
Requirement comes from here: issue: milvus-io/milvus#47425

Conan 1.x has been marked as deprecated in 2024.

PR #1592, opened 2026-04-17
Add MacOS CI pipeline

yhmo:

### DIFF
--- a/.github/workflows/ut.yaml
+++ b/.github/workflows/ut.yaml
@@ -55,6 +55,40 @@ jobs:
         with:
           os: ubuntu-22.04
 
+  ut-macos:
+    name: ut on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 180
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-14, macos-15]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: 'none'
+      - name: Install Dependency
+        run: |
+          bash scripts/install_deps.sh
+      - name: Restore Cache
+        uses: ./.github/actions/cache-restore
+        with:
+          os: ${{ matrix.os }}
+      - name: Build
+        run: |
+          # CMake 4.x (Homebrew) removed compat with cmake_minimum_required < 3.5;
+          # some Conan recipes (e.g. bzip2/1.0.8) still use older values.
+          export CMAKE_POLICY_VERSION_MINIMUM=3.5
+          make WITH_UT=True
+      - name: Test
+        run: |
+          make test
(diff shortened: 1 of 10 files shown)


scripts/install_deps.sh
```diff
@@ -110,15 +110,25 @@ else
...
-pip3 install bfloat16   # wheel: bfloat16 dtype support for PyKnowhere
-pip3 install auditwheel  # wheel: manylinux wheel repair
+if [[ "${OS}" == "Mac" ]]; then
+    # macOS Homebrew Python (PEP 668) blocks system-wide pip install;
+    # use pipx for conan. Wheel-building packages (setuptools, wheel,
+    # bfloat16, auditwheel) are Linux-only for the swig-build job.
+    brew install pipx
+    pipx ensurepath
+    pipx install conan==${CONAN_VERSION}
+    export PATH="$HOME/.local/bin:$PATH"
```
liliu-z, line 120: conan is installed via pipx but PATH is only exported in the current shell. In GitHub Actions each step spawns a new shell, so subsequent steps (Build, Test) won't find conan. Write `~/.local/bin` to `$GITHUB_PATH` to persist the path across steps.

.github/workflows/ut.yaml
```diff
@@ -55,6 +55,40 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: 'none'
+      - name: Install Dependency
+        run: |
+          bash scripts/install_deps.sh
+      - name: Restore Cache
+        uses: ./.github/actions/cache-restore
```
liliu-z, line 75: Pre-existing issue: ccache cache path mismatch between CMakeLists.txt (`~/.ccache`), the cache action (`~/.cache/ccache`), and macOS default (`~/Library/Caches/ccache`) means restored cache is never read on macOS runners. Not a blocker for this PR but worth a follow-up fix.

thirdparty/faiss/faiss/impl/mapped_io.cpp
```diff
@@ -27,7 +27,7 @@
 
 namespace faiss {
 
-#if defined(__linux__) || defined(__FreeBSD__)
+#if defined(__linux__) || defined(__FreeBSD__) || defined(__APPLE__)
```
liliu-z, line 30: mmap() returns MAP_FAILED ((void*)-1) on failure, not nullptr. This check will miss failures, leading to madvise()/munmap() on an invalid address. Must compare against MAP_FAILED.

liliu-z, line 30: Same mmap() failure check bug here — comparing against nullptr instead of MAP_FAILED.

zhengbuqian (approved):

PR #1594, opened 2026-04-20
Remove fork SQ in favor of vanilla FAISS version

alexanderguzhva:
this should be supported by baseline PR https://github.com/facebookresearch/faiss/pull/5128, which introduces RVV SQ in the vanilla faiss

Removes fork `ScalarQuantizer` facilities and uses the ones from the baseline


alexanderguzhva:
issue: #1501 
/kind improvement

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/sq-avx2-fastpath.cpp
```diff
@@ -0,0 +1,328 @@
...
+        const float final_scale = vdiff / 15.0f;
+        final_scale_sq = final_scale * final_scale;
+    }
+
+    void set_query(const float* x) final {
+        this->q = x;
+        const float inv_scale = (vdiff == 0.0f) ? 0.0f : 15.0f / vdiff;
+        for (size_t i = 0; i < d; i++) {
+            float val = (x[i] - vmin) * inv_scale;
+            int code = static_cast<int>(std::floor(val + 0.5f));
```
foxspy: encode the query with static_cast<int>floor(val + 0.5f) (round-half-up), but the database encoder at thirdparty/faiss/faiss/impl/scalar_quantizer/codecs.h:52 uses (int)(x * 15.0) (truncation toward zero). The two roundings disagree on every input whose fractional part after scaling lands in (0.5, 1.0) — roughly 50% of values for random inputs. please check this

foxspy (approved):
/lgtm

PR #1606, opened 2026-04-28
upgrade faiss 1.14.1+ baseline f9f116d30 -> 9d567497e (22 upstream commits)

alexanderguzhva:
/kind improvement
issue: #1501

### DIFF
--- a/thirdparty/faiss/faiss/IndexIVF.cpp
+++ b/thirdparty/faiss/faiss/IndexIVF.cpp
@@ -8,9 +8,9 @@
 #include <faiss/IndexIVF.h>
 
 #include <omp.h>
+#include <atomic>
 #include <cstdint>
 #include <memory>
-#include <mutex>
 
 #include <algorithm>
 #include <cinttypes>
@@ -58,6 +58,7 @@ void Level1Quantizer::train_q1(
         const float* x,
         bool verbose,
         MetricType metric_type) {
+    FAISS_THROW_IF_NOT_MSG(quantizer, "IVF quantizer must not be null");
     size_t d = quantizer->d;
     if (quantizer->is_trained &&
         (static_cast<size_t>(quantizer->ntotal) == nlist)) {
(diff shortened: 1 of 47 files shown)


thirdparty/faiss/faiss/impl/svs_io.cpp
```diff
@@ -77,9 +78,17 @@ std::streamsize ReaderStreambuf::xsgetn(char* s, std::streamsize n) {
...
         return 0;
     }
+    size_t nbytes = n;
 
-    size_t got = (*r)(s, 1, n);
-    return static_cast<std::streamsize>(got);
+    // Per-read byte limit: reject individual reads that exceed the limit,
+    // matching READVECTOR semantics where each vector allocation is
+    // independently checked against deserialization_vector_byte_limit.
+    if (per_read_byte_limit > 0 && nbytes >= per_read_byte_limit) {
```
foxspy, line 86: nbytes > per_read_byte_limit ？

thirdparty/faiss/faiss/IndexIVF.cpp
```diff
@@ -610,39 +601,39 @@ void IndexIVF::search_preassigned(
...
 
-                    if (InterruptCallback::is_interrupted()) {
-                        interrupt = true;
+                        InterruptCallback::check();
+                    } catch (...) {
+                        omp_capture_exception(ex, [&] { interrupt = true; });
                     }
-
                 } // parallel for
             } else if (pmode == 1) {
```
foxspy, line 638: This exception handling could be deadlock in pmode == 1.

`scanner->set_query()` and `init_result()` run before the `#pragma omp for` / barrier sequence. If either throws in one worker, the outer catch captures the exception and that worker skips the later OpenMP worksharing/barrier constructs. Other workers may still enter the `omp for` and then wait at the implicit/explicit barriers, but the failed worker will never arrive.

`omp_capture_exception()` itself is not the deadlock source; the issue is that catching at this outer level changes one thread's control flow so it bypasses collective OpenMP constructs. Please restructure this path so all threads still reach the same `omp for`/barrier constructs after an exception is captured, or move exception-prone setup into a region that cannot make a worker skip those constructs.

foxspy (approved):
/lgtm

PR #1610, opened 2026-04-30
Fix Conan cache restore isolation

jamesgao-jpg:
## What

This PR makes the GitHub Actions Conan cache restore safer on self-hosted runners.

It was prompted by PR https://github.com/zilliztech/knowhere/pull/1609, whose Unit Test run restored a stale `main` Conan cache via the broad `restore-keys` prefix and then failed with a Conan package directory collision.

Changes:
- remove Conan `restore-keys` so a changed `conanfile.py` no longer falls back to an older Conan cache
- remove `~/.conan2/p` before restoring the Conan cache so persistent runner leftovers cannot collide with restored Conan metadata
- keep ccache restore behavior unchanged

## Verification

Ran locally:
- `git diff --check`
- structural check confirming the reset step exists and the Conan cache block has no `restore-keys`
- `python3` YAML parse of `.github/actions/cache-restore/action.yaml`

Note: full validation requires a GitHub Actions run on the self-hosted runner.


foxspy (approved):
/lgtm

alexanderguzhva:
this is part of #1608 process
issue: #1608 
/kind improvement

PR #1605, opened 2026-04-28
Mostly get rid of FAISS fork in favor of vanilla FAISS version

alexanderguzhva:
putting here as a backup copy, just in case if my hard drive burns away tomorrow :)

includes #1606 

---

This is a broad FAISS fork-pruning pass for Knowhere. It removes large duplicated fork implementations and shifts IVF, FastScan, RaBitQ, Flat, DirectMap, and inverted-list plumbing toward baseline FAISS, while keeping Knowhere-specific behavior as focused adapters: cosine norm handling, CC segment-aware search, iterator hooks, ScaNN integration, and legacy serialization compatibility.

Faiss 1.14.1+ commit 46def46902231ed57552ddbcfd751e17720c6d32 plus merged https://github.com/facebookresearch/faiss/pull/5156 plus merged https://github.com/facebookresearch/faiss/pull/5172

Main pieces completed:

* Replaced fork IndexIVFFastScan with a baseline alias and deleted the fork .cpp; ordinary IndexIVFPQFastScan now aliases baseline FAISS, with only a small Knowhere cosine adapter retained.
* Preserved IwPf FastScan serialization while allowing baseline FastScan objects and HasInverseL2Norms-based cosine round-trips.
* Moved non-FastScan RaBitQ onto baseline faiss::IndexIVFRaBitQ, removed the fork RaBitQuantizer, added a thin iterator-hook shim, and exposed rbq_bits for multi-bit RaBitQ.
* Kept RaBitQ wire compatibility: legacy 1-bit indexes still write IwrQ, while baseline Iwrq and multi-bit Iwrr are readable/writable.
* Collapsed fork IVF/DirectMap/InvertedLists surfaces toward baseline types, adding InvertedListScannerStats and making IVF stats count post-filter computed distances.
* Isolated CC/qianya behavior into ConcurrentDirectMap and shared cc_search helpers.
* Slimmed IndexFlat to a baseline subclass that only preserves Knowhere’s Jaccard/Tanimoto behavior; removed IndexFlatElkan.
* Added/extended golden IVF serialization tests, RaBitQ config/search coverage, FastScan early-termination tests, and C/Python bindings for new IVF search parameters.

The commit is intentionally compatibility-heavy: the runtime implementation is pruned toward baseline FAISS, but existing Knowhere file formats, ScaNN recall-sensitive behavior, iterator paths, and cosine/refine semantics are kept as guardrails for this alpha version.

PR #1613, opened 2026-05-04
Optimize fp32 RVV distance kernels with e32m8 accumulator

ihb2032:
## What

This PR optimizes non-batch fp32 RVV distance kernels by replacing the current `e32m2 x 4` accumulator implementation with a simpler `e32m8 x 1` strip-mining implementation.

Updated kernels:

- `fvec_inner_product_rvv`
- `fvec_L2sqr_rvv`
- `fvec_norm_L2sqr_rvv`

The new implementation uses one `vfloat32m8_t` accumulator and handles the whole loop through `vsetvl_e32m8`, removing the manual 4-way unrolled main loop and separate tail loop.

## Why

The previous implementation used four `e32m2` accumulators to process `4 * VLMAX(e32m2)` elements per main-loop iteration. Since this is equivalent to `VLMAX(e32m8)`, the same number of elements can be processed with a single `e32m8` vector operation.

On Muse Pi Pro, the `e32m8 x 1` version is consistently faster than the current `e32m2 x 4` implementation. The improvement is especially stable for medium and large dimensions.

## Benchmark

Tested on Muse Pi Pro.

### `fvec_inner_product_rvv`

`speedup = old_m2x4_ns / new_m8x1_ns`

| dim | speedup |
|---:|---:|
| 16 | 1.38x |
| 32 | 1.85x |
| 64 | 2.39x |
| 128 | 2.52x |
| 256 | 2.62x |
| 512 | 2.68x |
| 1024 | 2.72x |
| 2048 | 2.74x |

For `d >= 256`, the speedup is stable around `2.62x ~ 2.74x`.

The maximum observed absolute difference from the long-double reference was around `2e-6`.

### `fvec_L2sqr_rvv`

`speedup = old_m2x4_ns / new_m8x1_ns`

| dim | speedup |
|---:|---:|
| 16 | 1.33x |
| 32 | 1.85x |
| 64 | 2.43x |
| 128 | 2.57x |
| 256 | 2.66x |
| 512 | 2.71x |
| 1024 | 2.74x |
| 2048 | 2.77x |

For `d >= 256`, the speedup is stable around `2.65x ~ 2.77x`.

The maximum observed absolute difference from the long-double reference was around `7.1e-5`.

## Notes

The new implementation may produce slightly different floating-point results from the previous RVV implementation because the accumulation/reduction order is different. This is expected for vectorized floating-point reduction and the observed numerical differences are small.
issue: #1614

### DIFF
--- a/src/simd/distances_rvv.cc
+++ b/src/simd/distances_rvv.cc
@@ -54,106 +54,49 @@ bf16_float_rvv(vfloat32m1_t f, size_t vl) {
 // =================== float distances ===================
 float
 fvec_inner_product_rvv(const float* x, const float* y, size_t d) {
-    size_t vlmax = __riscv_vsetvlmax_e32m2();  // Use m2 to support 4-way parallelism
-
-    // 4 accumulators
-    vfloat32m2_t vacc0 = __riscv_vfmv_v_f_f32m2(0.0f, vlmax);
-    vfloat32m2_t vacc1 = __riscv_vfmv_v_f_f32m2(0.0f, vlmax);
-    vfloat32m2_t vacc2 = __riscv_vfmv_v_f_f32m2(0.0f, vlmax);
-    vfloat32m2_t vacc3 = __riscv_vfmv_v_f_f32m2(0.0f, vlmax);
+    const size_t vlmax = __riscv_vsetvlmax_e32m8();
+    vfloat32m8_t acc = __riscv_vfmv_v_f_f32m8(0.0f, vlmax);
 
     size_t offset = 0;
+    while (offset < d) {
+        const size_t vl = __riscv_vsetvl_e32m8(d - offset);
 
-    // 4-way unrolled loop
-    while (d >= 4 * vlmax) {
-        size_t vl = vlmax;
+        vfloat32m8_t vx = __riscv_vle32_v_f32m8(x + offset, vl);
+        vfloat32m8_t vy = __riscv_vle32_v_f32m8(y + offset, vl);
 
-        vfloat32m2_t vx0 = __riscv_vle32_v_f32m2(x + offset, vl);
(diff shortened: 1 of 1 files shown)


alexanderguzhva:
@ihb2032 Thanks for your donation, . A question: is this `e32m8` supported only on high-end chips? Basically, I wonder whether the situation with `e32m2` vs `e32m8` is somewhat similar to AVX2 vs AVX512 for x86.

ihb2032:
Thanks for the question.

According to the RISC-V Vector spec, LMUL is the vector register group multiplier. When LMUL is greater than 1, it represents the number of vector registers combined to form one vector register group. The spec also says implementations must support integer LMUL values 1, 2, 4, and 8.

So `e32m8` means SEW=32 with LMUL=8: one vector operand uses a group of 8 vector registers, and its VLMAX is `8 * VLEN / 32`.

RVV also uses a vector-length-agnostic / strip-mining model. The program provides AVL, and `vsetvl`/`vsetvli` sets `vl` to the number of elements the hardware will process in that iteration, based on the implementation and the current `vtype`. Therefore the same loop is not tied to a fixed SIMD width.

alexanderguzhva:
understood.

/hold
wait for #1605

alexanderguzhva:
@ihb2032  please rebase to master. Thanks.

alexanderguzhva:
@ihb2032 
please rebase to master again. Sorry about that, there was a needed change to enable proper compilation for ubuntu 22.04, 24.04 and 26.04. It is over, I think.

ihb2032:
Hi @alexanderguzhva , could you please take another look at this PR when you have a chance? It has been rebased as requested, but `e2e`, `e2e(arm)`, and `Jenkins` are still failing with build errors. Please let me know if anything else is needed from my side. Thanks!

PR #1618, opened 2026-05-06
Upgrade faiss to 1.14.1+, commit 967eda65dec2515

alexanderguzhva:
As far as I understand, this finalizes `Faiss` introduction of dynamic dispatch

/kind improvement
issue: #1501


thirdparty/faiss/faiss/utils/simd_impl/hamming_impl.h
```diff
@@ -0,0 +1,481 @@
+/*
```
liliu-z, line 1: This file was added by this PR but has zero include references anywhere in the codebase. The actual implementation used by `hamming.cpp` and the per-ISA TUs is `utils/hamming_distance/hamming_impl.h`. This dead duplicate should be deleted to avoid the two implementations drifting apart over time.

liliu-z:
**thirdparty/faiss/faiss/utils/pq_code_distance.h:0** This refactor replaced the old `pq_code_distance_single`/`pq_code_distance_four` symbols with `pq_code_distance_8bit_single`/`pq_code_distance_8bit_four`. The header was removed from the `FAISS_HEADERS` install list and its only test consumer migrated to the new `impl/pq_code_distance/` includes, but the old file still exists on disk. It could mislead future maintainers. Consider deleting it or converting it to a compatibility forwarding header.

thirdparty/faiss/faiss/utils/simd_impl/hamming_impl.h
alexanderguzhva: this is the baseline faiss code, so this file should stay as is

alexanderguzhva:
> **thirdparty/faiss/faiss/utils/pq_code_distance.h:0** This refactor replaced the old `pq_code_distance_single`/`pq_code_distance_four` symbols with `pq_code_distance_8bit_single`/`pq_code_distance_8bit_four`. The header was removed from the `FAISS_HEADERS` install list and its only test consumer migrated to the new `impl/pq_code_distance/` includes, but the old file still exists on disk. It could mislead future maintainers. Consider deleting it or converting it to a compatibility forwarding header.

same story. This is the baseline faiss, should it should not be touched

foxspy (approved):
/lgtm

Issue #1615, opened 2026-05-05
Cannot build knowhere in Ubuntu 26.04


FantasqueX:
When executing `make` in Ubuntu 26.04, an error occurs.
```
-------- Installing package double-conversion/3.3.0 (6 of 42) --------
double-conversion/3.3.0: Building from source
double-conversion/3.3.0: Package double-conversion/3.3.0:813417aa23475817fe97360a04a1d7d9d77b11c5
double-conversion/3.3.0: settings: os=Linux arch=x86_64 compiler=gcc compiler.cppstd=20 compiler.libcxx=libstdc++11 compiler.version=15 build_type=Release
double-conversion/3.3.0: options: fPIC=True shared=False
double-conversion/3.3.0: Copying sources to build folder
double-conversion/3.3.0: Building your package in /home/fanta/.conan2/p/b/doubla4ad9373e4c2e/b
double-conversion/3.3.0: Calling generate()
double-conversion/3.3.0: Generators folder: /home/fanta/.conan2/p/b/doubla4ad9373e4c2e/b/build/Release/generators
double-conversion/3.3.0: CMakeToolchain generated: conan_toolchain.cmake
double-conversion/3.3.0: CMakeToolchain generated: /home/fanta/.conan2/p/b/doubla4ad9373e4c2e/b/build/Release/generators/CMakePresets.json
double-conversion/3.3.0: CMakeToolchain generated: /home/fanta/.conan2/p/b/doubla4ad9373e4c2e/b/src/CMakeUserPresets.json
double-conversion/3.3.0: Generating aggregated env files
double-conversion/3.3.0: Generated aggregated env files: ['conanbuild.sh', 'conanrun.sh']
double-conversion/3.3.0: Calling build()
double-conversion/3.3.0: Running CMake.configure()
double-conversion/3.3.0: RUN: cmake -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/home/fanta/.conan2/p/b/doubla4ad9373e4c2e/p" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/home/fanta/.conan2/p/b/doubla4ad9373e4c2e/b/src"
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

[...]

PR #1619, opened 2026-05-06
Update build scripts

alexanderguzhva:
This change allows building on Ubuntu 22.04, Ubuntu 24.04 and Ubuntu 26.04 or x86 and ARM, for `gcc-12`/`gcc-13`/`gcc-15` and `clang-22`.

Additional changes:
* Upgrades `conan` to `2.28.1` in order to support `clang-22`
* Introduces `uv`
* replaces `bfloat16` with `ml-dtypes`
* replaces `numpy<2` with `numpy` 2+
* fixes a problem with `SWIG` on Ubuntu 26.04
* dumps Python 3.8 and 3.9, but adds 3.12, 3.13 and 3.14

This is the first change of the following updates.

/kind improvement
issue: #1615

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,10 @@ CARDINAL_VERSION_FORCE_CHECKOUT ?=
 WITH_DEBUG ?=
 CONAN_PROFILE ?=
 
+# CMake 4.x removed compatibility with cmake_minimum_required versions below
+# 3.5. Export this so Conan-built third-party packages inherit it too.
+export CMAKE_POLICY_VERSION_MINIMUM ?= 3.5
+
 # Prevent build-flag variables from leaking into the environment of child
 # processes (conan / cmake).  Without this, GNU Make exports command-line
 # variables such as WITH_ASAN to every sub-process, which causes the custom
@@ -104,7 +108,7 @@ ifneq ($(CARDINAL_VERSION_FORCE_CHECKOUT),)
 endif
 
 ifdef CONAN_PROFILE
-    CONAN_SETTINGS += -pr $(CONAN_PROFILE)
+    CONAN_SETTINGS += -pr:h $(CONAN_PROFILE) -pr:b $(CONAN_PROFILE)
 endif
 
 .PHONY: build test \
@@ -173,7 +177,7 @@ help: ## Show available targets
 	@echo "  WITH_CARDINAL=True Enable Cardinal build"
 	@echo "  CARDINAL_VERSION_FORCE_CHECKOUT=True Force Cardinal checkout to configured version"
 	@echo "  WITH_DEBUG=True    Debug build (default: Release)"
-	@echo "  CONAN_PROFILE=<p>  Use a custom Conan profile (e.g. clang, gcc-15)"
+	@echo "  CONAN_PROFILE=<p>  Use one custom Conan profile for host and build"
 	@echo "  LIBCXX=<lib>       Override compiler.libcxx (auto-detected from OS)"
 	@echo ""
 	@echo "Examples:"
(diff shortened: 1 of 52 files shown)


scripts/conan_profiles/clang_22_arm
```diff
@@ -0,0 +1,31 @@
+[settings]
+arch=armv8
+build_type=Release
+compiler=clang
+compiler.cppstd=20
+compiler.libcxx=libstdc++11
+compiler.version=22
```
liliu-z, line 12: The profile sets `compiler.version=22`, but Conan 2.25.1's `settings.yml` only defines clang versions up to 21. Running `conan profile show` with this profile produces `Invalid setting '22'`, making the profile completely unusable. Either wait for a Conan release that includes clang 22 in its settings, or override `settings.yml` to add version 22.

scripts/install_deps.sh
```diff
@@ -39,65 +43,272 @@ esac
...
         )
-        ${SUDO} apt-get install -y "${packages[@]}"
+        apt_get install -y "${packages[@]}"
+        install_recent_cmake_apt
 
     elif command -v yum >/dev/null 2>&1; then
-        ${SUDO} yum makecache || true
+        yum_cmd makecache || true
 
         packages=(
```
liliu-z, line 283: The `clang_22_arm` profile references `clang-22` and `clang++-22` (line 17 of the profile), but `install_deps.sh` only installs `clang-tidy-22` and `clang-tools-extra` — the actual compiler binaries are missing. This affects both the `apt` (Ubuntu) and `yum` (RHEL) code paths. There is no reproducible "install deps → use clang-22 profile" path.

Makefile
```diff
@@ -93,7 +97,7 @@ ifdef WITH_CARDINAL
 endif
 
 ifdef CONAN_PROFILE
-    CONAN_SETTINGS += -pr $(CONAN_PROFILE)
+    CONAN_SETTINGS += -pr:h $(CONAN_PROFILE) -pr:b $(CONAN_PROFILE)
```
liliu-z, line 111: `-pr:b $(CONAN_PROFILE)` passes the cross-compilation profile (e.g. `clang_22_arm` with `arch=armv8` and `--target=aarch64-linux-gnu`) as the *build* profile. This means host-tool dependencies like `grpc_cpp_plugin` and `protoc` (referenced in Makefile L61-64 comments) will be compiled for ARM and cannot execute on the x86 build machine. This is a guaranteed failure for any cross-compilation profile, not just `clang_22_arm`. The build profile should describe the *build* machine (typically the default profile or an explicit x86 profile).

scripts/install_deps.sh
```diff
@@ -39,65 +43,272 @@ esac
...
+
 if [[ "${OS}" == "Linux" ]]; then
     # Use sudo for package manager commands when not running as root.
     if [ "$(id -u)" -ne 0 ]; then
-        SUDO="sudo"
+        if ! command -v sudo >/dev/null 2>&1; then
+            echo "[install_deps] Need root privileges or sudo to install packages."
+            exit 1
+        fi
+        SUDO=(sudo -n)
```
liliu-z: The script now uses `sudo -n` which will silently fail on developer machines that require a password for sudo. Since `install_deps.sh` is documented as the developer-facing install entry point (README L34), the previous interactive behavior was more appropriate for local use. CI environments (running as root or with passwordless sudo) are unaffected.

yhmo:
@alexanderguzhva 
I just called a review bot of @liliu-z to review this pr, the comments might be useful.

scripts/conan_profiles/clang_22_arm
alexanderguzhva: I will upgrade conan to 2.28.1

scripts/install_deps.sh
alexanderguzhva: there's no need to install clang-22, this is by design

Makefile
alexanderguzhva: @liliu-z please confirm that we're doing the cross-compilation

alexanderguzhva:
@yhmo updated. Please also re-read the description for this commit.

Makefile
liliu-z: No, ignore this

liliu-z (commented):
/lgtm
/approve

foxspy (approved):
/lgtm

PR #1620, opened 2026-05-07
ci: disable SSE CI

jamesgao-jpg:
## Summary
- replace the SSE Jenkins pipeline with a no-op stage
- use `agent none` so the disabled job does not allocate the SSE Kubernetes pod

## Testing
- `if rg -n "ci/pod/e2e-sse.yaml|build_portable_wheel|pytest -v -m 'L0'" ci/E2E2-SSE.groovy; then echo "SSE CI is still active"; exit 1; fi; echo "SSE CI no-op check passed"`
- `git diff --check`

### DIFF
--- a/ci/E2E2-SSE.groovy
+++ b/ci/E2E2-SSE.groovy
@@ -1,92 +1,10 @@
-int total_timeout_minutes = 480
-def knowhere_wheel=''
 pipeline {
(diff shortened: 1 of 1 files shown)


foxspy (approved):
/lgtm

PR #1621, opened 2026-05-07
Include milvus-common via conan

yhmo:


alexanderguzhva:
@yhmo please rebase to master branch, bcz we disabled e2e for sse and increased the timeout for ARM. Thanks

cmake/libs/libfaiss.cmake
```diff
@@ -331,13 +335,27 @@ if(__PPC64)
   add_library(knowhere_utils STATIC ${UTILS_SRC})
   target_link_libraries(knowhere_utils PUBLIC glog::glog)
   target_link_libraries(knowhere_utils PUBLIC xxHash::xxhash)
+  target_link_libraries(knowhere_utils PUBLIC milvus-common::milvus-common)
 endif()
 
 
 if(APPLE)
```
liliu-z, line 342: The new code uses `find_library(... openblas REQUIRED)` on macOS (lines 342-358), but `conanfile.py` only declares `openblas/0.3.30` as a Conan dependency on Linux/Android (line 137-140). The old code used the system Apple Accelerate framework, which required no extra dependencies. This means macOS environments that previously built successfully will now fail at CMake configure time because openblas is not found.

Suggested fixes (pick one):
1. Preserve Apple Accelerate on macOS (minimal change)
2. Add macOS openblas to Conan management or `system_requirements()`
3. Document `brew install openblas` as a macOS build prerequisite

alexanderguzhva:
excellent
/lgtm
/approve

PR #1624, opened 2026-05-09
Fix build errors when WITH_CARDINAL=true

yhmo:

We are currently using the recipe branch of the Cardinal repository because the pull request has not yet been merged into the Cardinal repository: https://github.com/zilliztech/cardinal/pull/834. The Cardinal repository fetches Knowhere to execute the CI workflow, creating a coupling between the two repositories. To decouple them:
1. Use the Cardinal recipe branch in Knowhere to pass the CI check.
2. In Cardinal's PR [#834](https://github.com/zilliztech/cardinal/pull/834), use the Knowhere branch to run CI.
3. Merge the Knowhere PR.
4. In Cardinal's PR [#834](https://github.com/zilliztech/cardinal/pull/834), switch to using the Knowhere master branch to run CI, and merge the Cardinal PR once CI has passed.

Expose a variable CARDINAL_VERSION_FORCE_CHECKOUT, so local/CI builds can force thirdparty/cardinalv2 to the configured branch/tag.
After this fix, both "make" command and "make WITH_CARDINAL=true" can succeed.

### DIFF
--- a/cmake/libs/cardinal/v1/CMakeLists.txt
+++ b/cmake/libs/cardinal/v1/CMakeLists.txt
@@ -3,7 +3,7 @@ project(knowhere CXX C)
 
 # Use short SHA1 as version
 # v2.5 tag is used for cardinal v1
-set(CARDINAL_VERSION v2.5.105)
+set(CARDINAL_VERSION v1_blas)
 set(CARDINAL_REPO_URL "https://github.com/zilliztech/cardinal.git")
 
 set(CARDINAL_ROOT  "${KNOWHERE_THRID_ROOT}/cardinalv1")
--- a/cmake/libs/cardinal/v2/CMakeLists.txt
+++ b/cmake/libs/cardinal/v2/CMakeLists.txt
@@ -3,7 +3,7 @@ project(knowhere CXX C)
 
 # Use short SHA1 as version
 # v2.6 tag is used for cardinal v2
-set(CARDINAL_VERSION v3.0.1)
+set(CARDINAL_VERSION recipe)
 set(CARDINAL_REPO_URL "https://github.com/zilliztech/cardinal.git")
 
 set(CARDINAL_ROOT  "${KNOWHERE_THRID_ROOT}/cardinalv2")
--- a/conanfile.py
+++ b/conanfile.py
@@ -28,6 +28,7 @@ class KnowhereConan(ConanFile):
         "with_diskann": [True, False],
         "with_svs": [True, False],
         "with_cardinal": [True, False],
+        "cardinal_version_force_checkout": [True, False],
         "with_profiler": [True, False],
         "with_ut": [True, False],
         "with_benchmark": [True, False],
@@ -44,6 +45,7 @@ class KnowhereConan(ConanFile):
         "with_diskann": False,
         "with_svs": False,
         "with_cardinal": False,
+        "cardinal_version_force_checkout": False,
         "with_profiler": False,
         "with_ut": False,
         "glog/*:shared": True,
@@ -59,6 +61,7 @@ class KnowhereConan(ConanFile):
         "openssl/*:no_apps": True,
         "gflags/*:shared": True,
         "fmt/*:header_only": False,
+        "openblas/*:dynamic_arch": True,
         "with_faiss_tests": False,
         "opentelemetry-cpp/*:with_stl": True,
         "libcurl/*:with_ssl": False,
@@ -107,7 +110,7 @@ class KnowhereConan(ConanFile):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        if self.settings.os in ["Linux", "Android"] and str(self.settings.arch) in self._openblas_dynamic_arches:
+        if self.settings.os == "Linux" and str(self.settings.arch) in self._openblas_dynamic_arches:
             self.options["openblas"].dynamic_arch = True
 
     def requirements(self):
@@ -136,11 +139,9 @@ class KnowhereConan(ConanFile):
         self.requires("xxhash/0.8.3#caa6d0af1b951c247922e38fbcebdbe6")
         if self.settings.os == "Linux":
             self.requires("openblas/0.3.30")
-        if self.settings.os == "Android":
-            self.requires("openblas/0.3.30")
         if not self.options.with_light:
             self.requires("opentelemetry-cpp/1.23.0@milvus/dev#11bc565ec6e82910ae8f7471da756720")
-        if self.settings.os not in ["Macos", "Android"]:
+        if self.settings.os != "Macos":
             self.requires("libunwind/1.8.1#748a981ace010b80163a08867b732e71")
         if self.options.with_ut:
             self.requires("catch2/3.7.1")
@@ -210,6 +211,7 @@ class KnowhereConan(ConanFile):
         tc.variables["WITH_DISKANN"] = self.options.with_diskann
         tc.variables["WITH_SVS"] = self.options.with_svs
         tc.variables["WITH_CARDINAL"] = self.options.with_cardinal
+        tc.variables["CARDINAL_VERSION_FORCE_CHECKOUT"] = self.options.cardinal_version_force_checkout
         tc.variables["WITH_CUVS"] = self.options.with_cuvs
         tc.variables["WITH_PROFILER"] = self.options.with_profiler
         tc.variables["WITH_UT"] = self.options.with_ut
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ WITH_BENCHMARK ?=
 WITH_ASAN ?=
 WITH_SVS ?=
 WITH_CARDINAL ?=
+CARDINAL_VERSION_FORCE_CHECKOUT ?=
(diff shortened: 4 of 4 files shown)


cmake/libs/cardinal/v2/CMakeLists.txt
```diff
@@ -3,7 +3,13 @@ project(knowhere CXX C)
...
 # Use short SHA1 as version
 # v2.6 tag is used for cardinal v2
-set(CARDINAL_VERSION v3.0.1)
+# now we use recipe branch of cardinal repo since the pr is not merged into cardinal repo: https://github.com/zilliztech/cardinal/pull/834
+# the cardinal repo fetches knowhere to run ci workflow so the two repos are coupling. to decouple them:
+# 1. use cardinal recipe branch in knowhere to pass the ci check
+# 2. in the cardianl pr#834, use the knowhere branch to run ci
+# 3. merge the knowhere pr
+# 4. in the cardinal pr#834, use the knowhere master branch to run ci, and merge the cardinal pr after ci passed
+set(CARDINAL_VERSION recipe)
```
liliu-z, line 6: Changing `CARDINAL_VERSION` from tag `v3.0.1` to branch `recipe` introduces three related problems:

1. **Build non-reproducibility**: Fresh clones of the same Knowhere revision will produce different results as `recipe` moves.
2. **Cached workspace mismatch**: Existing workspaces with `thirdparty/cardinalv2` checked out at `v3.0.1` will not automatically switch to `recipe` (default `CARDINAL_VERSION_FORCE_CHECKOUT=OFF`), causing build failures since the new Knowhere code depends on API changes from `recipe`.
3. **force-checkout doesn't fast-forward branches**: The `git checkout recipe` at lines 57-75 won't advance an already-existing local `recipe` branch to `origin/recipe`, so even enabling force-checkout may not produce the correct result.

This is understood to be a temporary coordination measure. Suggestions:
- Ensure Cardinal PR #834 has tracking to switch back to a stable tag promptly.
- Change L71 to `git -C ${CARDINAL_ROOT} checkout -B ${CARDINAL_VERSION} origin/${CARDINAL_VERSION}` so force-checkout is branch-safe.
- Note in the PR description that developers with existing `thirdparty/cardinalv2` must delete it or set `CARDINAL_VERSION_FORCE_CHECKOUT=True`.

alexanderguzhva:
/hold
@yhmo why is this change needed here at all? Shouldn't the change be introduced in cardinal v1 instead?

cmake/libs/cardinal/v1/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(knowhere CXX C)
 
 # Use short SHA1 as version
 # v2.5 tag is used for cardinal v1
-set(CARDINAL_VERSION v2.5.105)
+set(CARDINAL_VERSION 12ad9e4687962624d8af7721b58367d11cf3fddb)
```
liliu-z: The pinned SHA `12ad9e4...` is a dangling commit not reachable from any branch, tag, or PR ref in the Cardinal repo. After `git clone`, `git checkout <sha>` fails with `fatal: reference is not a tree` because the object is not fetched. The `CARDINAL_VERSION_FORCE_CHECKOUT` path (line 47-75) also fails since plain `git fetch` only retrieves advertised refs. This means `make WITH_CARDINAL=true` is broken on any clean workspace.

Fix: push these commits to named branches in the Cardinal repo (e.g. `recipe/v1`) so they become fetchable via normal clone/fetch.

cmake/libs/cardinal/v2/CMakeLists.txt
```diff
@@ -3,7 +3,13 @@ project(knowhere CXX C)
...
 # Use short SHA1 as version
 # v2.6 tag is used for cardinal v2
-set(CARDINAL_VERSION v3.0.1)
+# now we use recipe branch of cardinal repo since the pr is not merged into cardinal repo: https://github.com/zilliztech/cardinal/pull/834
+# the cardinal repo fetches knowhere to run ci workflow so the two repos are coupling. to decouple them:
+# 1. use cardinal recipe branch in knowhere to pass the ci check
+# 2. in the cardianl pr#834, use the knowhere branch to run ci
+# 3. merge the knowhere pr
+# 4. in the cardinal pr#834, use the knowhere master branch to run ci, and merge the cardinal pr after ci passed
+set(CARDINAL_VERSION f0654cb40be0db57e73770707f44a1980b328423)
```
liliu-z: The pinned SHA `f0654cb...` is a dangling commit not reachable from any branch, tag, or PR ref. Same root cause as the v1 issue: `git clone` + `git checkout <sha>` fails, and `git fetch` in the force-checkout path (line 53) won't retrieve unreferenced objects. Fresh builds with `WITH_CARDINAL=true` will fail.

Fix: push this commit to a named branch in the Cardinal repo (e.g. `recipe/v2`).

cmake/libs/cardinal/v1/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(knowhere CXX C)
 
 # Use short SHA1 as version
 # v2.5 tag is used for cardinal v1
```
liliu-z, line 5: Line 5 says `# v2.5 tag is used for cardinal v1` but the version is now a commit SHA, not a tag. The v2 file has a detailed explanatory comment block (lines 6-11) describing the temporary state; v1 has no such context. Update the comment to match the current pinning strategy.

CMakeLists.txt
```diff
@@ -180,6 +180,7 @@ set(CARDINAL_LIBS "")
 if(WITH_CARDINAL)
   add_definitions(-DKNOWHERE_WITH_CARDINAL)
   add_subdirectory(cmake/libs/cardinal/v1)
+  target_link_libraries(cardinalv1 PRIVATE milvus-common::milvus-common)
```
liliu-z: The added `target_link_libraries(cardinalv1 PRIVATE milvus-common::milvus-common)` correctly fixes the target name change from PR #1621. However, the pinned Cardinal v1 commit's own `libcardinal.cmake` already links `milvus-common::milvus-common`, making this potentially redundant. CMake deduplicates so it's harmless, but should be reviewed when the Cardinal pin is updated to a proper tag. Also note `cardinalv2` (line 185-186) does NOT have this line added — verify the asymmetry is intentional.

cmake/libs/cardinal/v1/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(knowhere CXX C)
...
-set(CARDINAL_VERSION v2.5.105)
+set(CARDINAL_VERSION v1_blas)
```
liliu-z, line 6: `git fetch` + `git checkout ${CARDINAL_VERSION}` leaves a stale local branch in place when `CARDINAL_VERSION` is a branch name. With tag-based versions, `git checkout v3.0.1` entered detached HEAD (deterministic). With mutable branch names, the checkout is a no-op if the local branch already exists, so CI builds silently use wrong code.

This directly undermines the `CARDINAL_VERSION_FORCE_CHECKOUT` feature this PR introduces.

Fix: replace `git checkout ${CARDINAL_VERSION}` with `git checkout -B ${CARDINAL_VERSION} origin/${CARDINAL_VERSION}`, or use detached HEAD via `git fetch origin ${CARDINAL_VERSION}` + `git checkout FETCH_HEAD` if both tags and branches must be supported.

conanfile.py
```diff
@@ -136,11 +139,9 @@ def requirements(self):
         self.requires("xxhash/0.8.3#caa6d0af1b951c247922e38fbcebdbe6")
         if self.settings.os == "Linux":
```
liliu-z, line 140: The PR removes Android-specific OpenBLAS wiring and re-adds `libunwind` on Android (previously excluded). However, `cmake/utils/platform_check.cmake:18` still has explicit Android/aarch64 handling, and `cmake/libs/libfaiss.cmake:349` unconditionally does `find_package(OpenBLAS CONFIG REQUIRED)` on non-Apple platforms. If Android is still a target, this is a concrete compatibility regression. If Android is being dropped, the CMake-level Android paths should be cleaned up in the same PR.

conanfile.py
```diff
@@ -59,6 +61,7 @@ class KnowhereConan(ConanFile):
         "openssl/*:no_apps": True,
         "gflags/*:shared": True,
         "fmt/*:header_only": False,
+        "openblas/*:dynamic_arch": True,
```
liliu-z, line 64: Adding `dynamic_arch=True` to `default_options` (L64) makes it unconditionally True for all platforms. The `configure()` conditional at L113-114 only ever sets it to True on top of True — it's fully redundant and never sets False. This also widens scope beyond `_openblas_dynamic_arches` (e.g., RISC-V now gets `dynamic_arch=True` when it previously got the default False). Either remove the `default_options` entry (if arch-gating is intentional) or remove the `configure()` conditional (if it should always be True).

foxspy (approved):
/lgtm 
Going to merge first, then revert the nits later.

foxspy (approved):
/lgtm 
Going to merge first, then revert the nits later.

PR #1630, opened 2026-05-16
sparse: warm-start pruning threshold for BM25 DAAT_MAXSCORE

lyang24:
Bootstrap a safe non-zero pruning threshold before topk_ fills, so update_non_essential_lists() can demote terms early and current_score + lookup_bound pruning kicks in from the first candidate.

Mechanism: scan one seed term's posting list, compute single-term BM25 contribution per matching doc, take the k-th best as warm_threshold_. Since per-term BM25 contributions are non-negative, the k-th best single-term score is a safe lower bound on the k-th best full-query score (at least k docs are guaranteed to have full_score >= S_k). Exact-recall preserving.

Seed = query term with the largest per-dim max_score (high IDF, short posting list). Activated only for BM25, query.size() >= 2, k >= 2, and seed must hit >= k docs (otherwise S_m is not a valid bound on the k-th score). Falls back to current behavior on guard misses.

No change to ResultHeap semantics; warm threshold is searcher-local.

### DIFF
--- a/src/index/sparse/searcher/daat_maxscore.h
+++ b/src/index/sparse/searcher/daat_maxscore.h
@@ -61,6 +61,9 @@ class DaatMaxScoreSearcher : public RankedSearcher {
           max_vec_id_(max_vec_id),
           row_sums_(index.get_row_sums()),
           scorer_type_(search_scorer->config().scorer_type) {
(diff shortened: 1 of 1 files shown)


lyang24:
bench marked no gain closing

PR #1633, opened 2026-05-19
fix: allow CAGRA adapt-for-cpu load without GPU

### LINKED ISSUE #1631: GPU_CAGRA adapt_for_cpu=true still requires GPU during load/create on CPU-only nodes
See also: https://github.com/milvus-io/milvus/issues/49836

## Background

For `GPU_CAGRA` / `GPU_CUVS_CAGRA`, `adapt_for_cpu=true` is expected to allow a GPU-built CAGRA index to be loaded and searched on CPU-only QueryNodes.

This is needed for mixed Milvus deployments where the IndexNode/DataNode side has GPU resources, but some QueryNodes do not.

## Problem

The current Knowhere load/create path can still require CUDA before the CPU-compatible CAGRA deserialize path is reached:

- `IndexFactory::Create()` checks GPU availability for `GPU_CAGRA` / `GPU_CUVS_CAGRA`, but this function does not receive index config and cannot know whether `adapt_for_cpu=true`.
- `GpuCuvsIndexNode` eagerly owns a `cuvs_knowhere_index` value member, so constructing the node can initialize CUDA-backed state before `Deserialize()` has a chance to choose the CPU-compatible path.
- CAGRA thread-pool registration currently calls `RAFT_CUDA_TRY(cudaGetDeviceCount(...))`, which can fail during registration on CPU-only hosts.

As a result, Milvus can fail to load a CAGRA index on CPU-only QueryNodes even when the index was built with `adapt_for_cpu=true`.

## Expected Behavior

When loading `GPU_CAGRA` / `GPU_CUVS_CAGRA` with `adapt_for_cpu=true`:

- create/load should not require GPU availability before deserialization
- CPU-compatible CAGRA deserialize/search should be reachable
- normal GPU build/search behavior should remain unchanged
- other GPU index types should keep the existing GPU availability behavior

## Fix Direction

A minimal Knowhere-side fix is:

1. Remove `GPU_CAGRA` / `GPU_CUVS_CAGRA` from the generic create-time GPU availability check, because `Create()` cannot inspect `adapt_for_cpu`.
2. Lazily create the underlying `cuvs_knowhere_index` only in GPU paths such as train/search/regular GPU deserialize.
3. Make CAGRA thread-pool registration tolerate no-GPU environments by falling back to a small non-zero thread pool size.

This keeps the special behavior scoped to CAGRA, where `adapt_for_cpu` is already a supported load/search mode.

marcelo-cjl:
## Summary

- Allow `GPU_CAGRA` / `GPU_CUVS_CAGRA` factory creation to proceed without a create-time GPU availability check.
- Lazily instantiate the underlying cuVS CAGRA index so `adapt_for_cpu=true` deserialize can enter the CPU-compatible path on CPU-only nodes.
- Make CAGRA thread-pool registration tolerate no-GPU hosts by falling back to a non-zero thread pool size.

Fixes #1631
Related Milvus issue: milvus-io/milvus#49836

## Test plan

- [x] `git diff --check -- src/index/gpu_cuvs/gpu_cuvs.h src/index/gpu_cuvs/gpu_cuvs_cagra.cc src/index/index_factory.cc`
- [x] Validated through Milvus mixed GPU/CPU deployment with GPU-built `GPU_CAGRA` and CPU-only QueryNode load/search using `adapt_for_cpu=true`.
- [ ] Knowhere standalone build/UT not run in this workspace.

### DIFF
--- a/src/index/gpu_cuvs/gpu_cuvs.h
+++ b/src/index/gpu_cuvs/gpu_cuvs.h
@@ -21,6 +21,7 @@
 #include <exception>
 #include <fstream>
 #include <istream>
+#include <memory>
 #include <numeric>
 #include <thread>
 #include <tuple>
@@ -73,7 +74,7 @@ struct GpuCuvsIndexNode : public IndexNode {
     using knowhere_config_type = typename KnowhereConfigType<index_kind>::Type;
     using data_type = typename cuvs_knowhere::cuvs_data_type_mapper<DataType>::data_type;
 
-    GpuCuvsIndexNode(int32_t, const Object& object) : index_{} {
+    GpuCuvsIndexNode(int32_t, const Object& object) {
     }
 
     Status
@@ -86,7 +87,8 @@ struct GpuCuvsIndexNode : public IndexNode {
             LOG_KNOWHERE_ERROR_ << e.what();
             result = Status::invalid_args;
         }
-        if (index_.is_trained()) {
+        auto& cuvs_index = GetOrCreateIndex();
+        if (cuvs_index.is_trained()) {
             result = Status::index_already_trained;
         }
         if (result == Status::success) {
@@ -94,8 +96,8 @@ struct GpuCuvsIndexNode : public IndexNode {
             auto dim = dataset->GetDim();
             auto const* data = reinterpret_cast<const data_type*>(dataset->GetTensor());
             try {
-                index_.train(cuvs_cfg, data, rows, dim);
-                index_.synchronize(true);
+                cuvs_index.train(cuvs_cfg, data, rows, dim);
+                cuvs_index.synchronize(true);
             } catch (const std::exception& e) {
                 LOG_KNOWHERE_ERROR_ << e.what();
                 result = Status::cuvs_inner_error;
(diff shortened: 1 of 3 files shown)


src/index/gpu_cuvs/gpu_cuvs.h
```diff
@@ -250,14 +253,22 @@ struct GpuCuvsIndexNode : public IndexNode {
     using cuvs_knowhere_index_type = typename cuvs_knowhere::cuvs_knowhere_index<K, DataType>;
 
  protected:
-    cuvs_knowhere_index_type index_;
+    mutable std::unique_ptr<cuvs_knowhere_index_type> index_;
```
alexanderguzhva: can this `mutable` be avoided?

src/index/gpu_cuvs/gpu_cuvs.h
```diff
@@ -250,14 +253,22 @@ struct GpuCuvsIndexNode : public IndexNode {
...
+
+    cuvs_knowhere_index_type&
+    GetOrCreateIndex() const {
+        if (!index_) {
+            index_ = std::make_unique<cuvs_knowhere_index_type>();
```
alexanderguzhva, line 264: basically, this can cause problems in case of concurrent calls of this `GetOrCreateIndex()` function, especially given that it is marked as `const`

marcelo-cjl: Updated, thanks. I removed `mutable` and made `GetOrCreateIndex()` non-const. `Search()` no longer creates the cuVS index; it only uses an existing trained/deserialized index and returns `empty_index` if none exists.

src/index/gpu_cuvs/gpu_cuvs.h
marcelo-cjl: Yes, removed it. `index_` is now a regular `std::unique_ptr`; only non-const paths create it.

alexanderguzhva:
@marcelo-cjl this is rebased on top of master, correct? #1619 has upgraded numpy

marcelo-cjl:
@alexanderguzhva previously it was not rebased. It has now been rebased onto the latest `main`, so it includes #1619.

Issue #1612, opened 2026-05-04
Add KMEANS support


gland1:
Knowhere has defined an interface called cluster_node which has interfaces for kmeans training, assignment, and centroid fetching. 
Such interface is required by Milvus clustering compaction (by vector field), however knowhere has no implementation for this interface. We intend to submit a PR which adds a cluster_node implementation which is using faiss clustering.
The implementation we provide has the following features:
1)	Implement all cluster_node APIs 
2)	Enables repeated kmeans – run kmeans for several times and select the most balanced result.
3)	Mini batch kmeans – enables consecutive calls to Train, each call starts with previous computed centroids and calculates the next centroids by a weighted average.

PR #1634, opened 2026-05-19
Add KMEANS support

liorf95:
Related to #1612


alexanderguzhva:
@liorf95 fyi https://github.com/facebookresearch/faiss/commit/b7618fadc7fdf9b677277fbe96864c1902a6d8dc

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+        return expected<DataSetPtr>::Err(Status::invalid_param_in_json, "rain called again with different params");
+    }
+    std::vector<size_t> total_points_per_cluster = points_per_cluster;
+    size_t best_score = (size_t)-1, current_score = 0;
+    std::vector<size_t> best_point_per_cluster;
+    expected<DataSetPtr> best_res;
+    std::shared_ptr<faiss::Clustering> best_clustering;
+    // for first train batch try several kmeans to select the one which brings max balance
+    int num_iter = previous_clustering ? 1 : kmeans_conf.num_iter.value();
+    for (int i = 0; i < num_iter; i++) {
```
alexanderguzhva, line 122: how this is related to https://github.com/facebookresearch/faiss/blob/cb69d7c5665241d1e84f476105ffb6fd6c7c192c/faiss/Clustering.h#L22-L27 , because it seems that Faiss already has this feature?

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+                index->search(1, &queries[current * dataset.GetDim()], probes, dis.data() + current * probes,
+                              idx.data() + current * probes);
+            }));
+        }
+        WaitAllSuccess(futs);
+    } catch (const std::exception& e) {
+        LOG_KNOWHERE_WARNING_ << "faiss inner error: " << e.what();
+        return expected<DataSetPtr>::Err(Status::faiss_inner_error, e.what());
+    }
+    if (balance) {
```
alexanderguzhva, line 230: It would be nice to have some comments about the exact way this get balanced.

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+                        }
+                    }
+                    if (found) {
+                        break;
+                    }
+                }
+            }
+            p_id.get()[selected_point] = i;
+            points_per_cluster[i]++;
+            points_per_cluster[selected_point]--;
```
alexanderguzhva: this line seems to be wrong, because `selected_point` is a row index, while `points_per_cluster` is a vector of a size of cluster indices

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+    std::vector<DataType> dis(num_rows * probes);
+    std::vector<faiss::idx_t> idx(num_rows * probes);
+    size_t minimum_row = 3;
+    size_t avg_points_per_cluster = (num_rows / clustering->k);
+    size_t max_points_per_cluster = avg_points_per_cluster * 2;
+    size_t min_points_per_cluster = avg_points_per_cluster / 100 + 1;
+    size_t min_points_for_transfer = min_points_per_cluster * 10;
+    size_t num_points_assigned_not_optimally = 0;
+    std::vector<size_t> location_diff(num_rows);
+    memset(points_per_cluster.data(), 0, clustering->k * sizeof(size_t));
```
alexanderguzhva: I'd use `std::fill()` here

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+    memset(location_diff.data(), 0, num_rows * sizeof(size_t));
```
alexanderguzhva: remove this one `memset(location_diff.data(), 0, num_rows * sizeof(size_t));` and go with `std::vector<size_t> location_diff(num_rows, 0);` instead

src/cluster/kmeans/faiss_kmeans.cc
```diff
@@ -0,0 +1,331 @@
...
+            }
+        }
+        std::sort(location_diff.begin(), location_diff.end(), std::greater<size_t>());
+    } else {
+        for (int i = 0; i < num_rows; i++) {
+            p_id.get()[i] = idx[i];
+            points_per_cluster[idx[i]]++;
+        }
+    }
+    size_t min_points = num_rows, max_points = 0;
```
alexanderguzhva: please, 1 declaration per row
```C++
size_t min_points = num_rows;
size_t max_points = 0;
```

alexanderguzhva:
/kind improvement
issue: #1612

src/cluster/kmeans/faiss_kmeans.cc
gland1: > how this is related to https://github.com/facebookresearch/faiss/blob/cb69d7c5665241d1e84f476105ffb6fd6c7c192c/faiss/Clustering.h#L22-L27 , because it seems that Faiss already has this feature?

We've found that faiss objective with redo does not neccessairly reproduce the most  balanced result in terms of diff between max size cluster and min

src/cluster/kmeans/faiss_kmeans.cc
gland1: actually the balancing code is something we've used in the past but it is not currently active.we thought to leave it and maybe someone else will find use for it.we now have a much better algorithm which takes care of final segment balancing and it happens in milvus level

src/cluster/kmeans/faiss_kmeans.cc
gland1: The balancing code looks for cluster which didnt have enough points and tries to move points from other clusters which have more than enough points. but it moves a point from that cluster only if it is close enough to the cluster we want to move it to. points_per_cluster[x] counts the number of vectors assigned to each cluster. when we decide to move a point due to above, we decrease the number of points from the cluster it moved from and increase the number of points it moved to.

src/cluster/kmeans/faiss_kmeans.cc
gland1: ok

src/cluster/kmeans/faiss_kmeans.cc
gland1: ok

src/cluster/kmeans/faiss_kmeans.cc
gland1: ok

src/cluster/kmeans/faiss_kmeans.cc
alexanderguzhva: @gland1 Say, we have 1M points and 1024 clusters.

* line 200 states `memset(points_per_cluster.data(), 0, clustering->k * sizeof(size_t));`, so the size is `clustering->k` (1024)
* line 260 states `for (int j = 0; j < num_rows; j++) {`, so `j` can be 999999
* line 290 states `selected_point = j;`, so `selected_point` can become 999999
* line 284 states `points_per_cluster[selected_point]--;`, which may lead to out-of-range error

At least, it is not evident to me, why `selected_point` ends up being less than `clustering->k`

gland1: right...guess we didnt notice this because the code was not really in use

alexanderguzhva:
@liorf95 FYI, macos compilation error
```
2026-05-20T16:00:24.9752690Z [ 88%] Building CXX object tests/ut/CMakeFiles/knowhere_tests.dir/test_config.cc.o
2026-05-20T16:00:26.3252130Z In file included from /Users/runner/work/knowhere/knowhere/tests/ut/test_cluster.cc:16:
2026-05-20T16:00:26.3353730Z /Users/runner/work/knowhere/knowhere/thirdparty/hnswlib/../DiskANN/include/diskann/partition_and_pq.h:13:10: fatal error: 'tsl/robin_set.h' file not found
2026-05-20T16:00:26.3455200Z    13 | #include "tsl/robin_set.h"
2026-05-20T16:00:26.3498950Z       |          ^~~~~~~~~~~~~~~~~
2026-05-20T16:00:26.4707680Z 1 error generated.
2026-05-20T16:00:26.4809820Z make[3]: *** [tests/ut/CMakeFiles/knowhere_tests.dir/test_cluster.cc.o] Error 1
2026-05-20T16:00:26.4912040Z make[3]: *** Waiting for unfinished jobs....
```

alexanderguzhva:
@liorf95 this is related to some github error, I'll restart manually

alexanderguzhva:
@liorf95 looks good overall, I'll merge it after the UT passes

PR #1645, opened 2026-05-26
fix: sparse SINDI build validation

sparknack:
issue: #1555

- Reject unsupported explicit sparse inverted_index_algo values during build
- Fix empty SINDI index POSTING_LISTS section size accounting

### DIFF
--- a/src/index/sparse/sparse_index_node.cc
+++ b/src/index/sparse/sparse_index_node.cc
@@ -11,7 +11,9 @@
 
 #include <sys/mman.h>
 
+#include <algorithm>
 #include <boost/intrusive/pack_options.hpp>
+#include <cctype>
 #include <exception>
 #include <optional>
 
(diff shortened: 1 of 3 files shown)


src/index/sparse/sparse_index_node.cc
```diff
@@ -420,6 +420,24 @@ class SparseInvertedIndexNode : public IndexNode {
  protected:
     std::unique_ptr<sparse::inverted::InvertedIndex<value_type>> index_;
 
+    Status
+    ValidateInvertedIndexAlgo(const SparseInvertedIndexConfig& cfg) const {
+        const auto algo = cfg.inverted_index_algo.value_or("");
```
alexanderguzhva: @sparknack  should `algo` be converted to upper-case before the check?

alexanderguzhva:
@sparknack looks good overall, waiting for the reply on comments

src/index/sparse/sparse_index_node.cc
sparknack: Done. The explicit inverted_index_algo is normalized to upper-case before validation, and CreateIndexImpl now uses the same normalized value when selecting/setting the build algorithm. I also changed the regression test to build a non-SINDI index with lower-case daat_maxscore so we cover both validation and actual index creation.

alexanderguzhva:
/kind improvement
issue: #1555

PR #1637, opened 2026-05-20
Aisaq improvments

liorf95:
Related to #1611

PR #1646, opened 2026-05-27
fix: Configure DiskANN before Cardinal

jamesgao-jpg:
## Context

Cardinal's Knowhere Unit Test builds Cardinal through Knowhere: the workflow checks out Knowhere, places Cardinal under `thirdparty/cardinalv1` and `thirdparty/cardinalv2`, then builds Knowhere with both `WITH_CARDINAL=True` and `WITH_DISKANN=True`.

Cardinal PR https://github.com/zilliztech/cardinal/pull/848 exposed this in CI, but the source-level trigger is in Knowhere. Knowhere PR https://github.com/zilliztech/knowhere/pull/1637 added `#include "diskann/defaults.h"` to `src/index/diskann/diskann_config.h` so it could use DiskANN default constants. Cardinal already includes Knowhere's `index/diskann/diskann_config.h` from its `know/cardinal_config.h`, so Cardinal targets now also need the DiskANN include directory when compiling `know/cardinal.cc`.

The failing Cardinal run is https://github.com/zilliztech/cardinal/actions/runs/26273280319/job/77885729100. It failed with:

```text
fatal error: diskann/defaults.h: No such file or directory
```

Related issue: https://github.com/zilliztech/knowhere/issues/1647

## Fix

Move the `WITH_DISKANN` setup before the `WITH_CARDINAL` setup in the top-level CMake file. This keeps the temporary fix in one place: when Cardinal is added, DiskANN's compile definitions and include directories have already been configured.

This is intentionally a block reorder only. It avoids adding duplicate DiskANN include paths to both Cardinal v1 and Cardinal v2 CMake files.

## Longer-term fix

The cleaner long-term fix is to avoid exposing DiskANN's private third-party header from Knowhere's config header. For example, `diskann_config.h` could keep Knowhere-owned default constants or use an explicit CMake interface dependency for DiskANN headers. That would make the header dependency explicit instead of relying on global include-directory order.

Tracked by https://github.com/zilliztech/knowhere/issues/1647.

## Verification

- Ran a red/green structural check locally:
  - before the change: failed because `WITH_DISKANN` was configured after `WITH_CARDINAL`
  - after the change: passed because `WITH_DISKANN` is configured before `WITH_CARDINAL`
- Ran `git diff --check`
- Attempted `make BUILD_DIR=/tmp/knowhere-cardinal-diskann-order-build WITH_CARDINAL=True` with a temporary Conan 2 venv and isolated `CONAN_HOME`; stopped it because the empty local Conan cache started rebuilding many third-party packages from source before reaching Knowhere CMake.


chasingegg (approved):
/lgtm

PR #1652, opened 2026-05-27
Fix SVS recall variability

mihaic:
Use SVS nightly release with improves filtered search, fixing the root cause of recall variability.

Update fmt to match SVS.

Fix Conan SVS flag.

issue: #1654

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ ifdef WITH_ASAN
 endif
 
 ifdef WITH_SVS
-    CONAN_FLAGS += -o \&:with_svs=True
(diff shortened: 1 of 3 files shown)


mihaic:
@alexanderguzhva, I see unit test failures unrelated to the changes; the failures are caused by using alpha 1.2 with metric IP. Alpha should be unspecified in most cases so the appropriate default is used.

foxspy:
issue: #1654

foxspy (approved):
/lgtm

foxspy (approved):
/lgtm

PR #1664, opened 2026-06-03
enhance: sparse: add skip_window_calc pruning for SINDI search

sparknack:
issue: #1555

Add early termination optimization that skips remaining terms in a window when the current max score plus remaining max contributions cannot exceed the top-k threshold. Includes per-dimension max score computation, query term sorting by max contribution, and suffix sum based pruning.

### DIFF
--- a/src/index/sparse/sindi_inverted_index.h
+++ b/src/index/sparse/sindi_inverted_index.h
@@ -198,12 +198,14 @@ class SindiInvertedIndex : public InvertedIndex<DataType> {
 
     Status
     add(const SparseRow<DataType>* data, size_t rows, int64_t dim) override {
+        const size_t old_nr_rows = this->nr_rows_;
         this->max_dim_ = std::max(this->max_dim_, static_cast<uint32_t>(dim));
         // update dim_map_
         for (size_t i = 0; i < rows; ++i) {
             for (size_t j = 0; j < data[i].size(); ++j) {
-                auto [dim, val] = data[i][j];
-                if (std::abs(val) < std::numeric_limits<DataType>::epsilon()) {
+                const auto [dim, val] = data[i][j];
+                const auto abs_val = std::abs(val);
+                if (abs_val < std::numeric_limits<DataType>::epsilon()) {
                     continue;
                 }
                 if (this->dim_map_.find(dim) == this->dim_map_.end()) {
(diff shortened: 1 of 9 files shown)


zhengbuqian (approved):

src/index/sparse/sindi_inverted_index.h
```diff
@@ -310,7 +345,7 @@ class SindiInvertedIndex : public InvertedIndex<DataType> {
         // Determine if we need to write row sums (BM25 support)
         const bool has_row_sums = !row_sums_span_.empty();
 
-        uint32_t nr_sections = 2;  // base sections: inverted index and dim map
+        uint32_t nr_sections = 3;  // base sections: inverted index, dim map and max scores per dim
```
liliu-z, line 347: The on-disk layout now writes 3/4 sections (previously 2/3) but `current_index_file_format_version_` is left unchanged, so the old and new formats share the same version tag. Since the loader keys off that version to decide how many sections to read, the stream is misparsed across versions — a v3.0.x SINDI index loaded by this build fails outright for IP and silently returns empty results for BM25, with no error surfaced to the caller. Bump `current_index_file_format_version_` and branch the section count on the version during load.

src/index/sparse/sindi_inverted_index.h
```diff
@@ -663,6 +714,36 @@ class SindiInvertedIndex : public InvertedIndex<DataType> {
             return;
         }
 
+        // Compute max contributions for each query term: qval * max_dim_val
+        // and sort query terms by max contributions descending for early termination
+        std::vector<float> max_contributions(q_vec.size());
+        for (size_t i = 0; i < q_vec.size(); ++i) {
+            auto& [qid, qval] = q_vec[i];
+            float dim_max = (qid < max_scores_per_dim_span_.size()) ? max_scores_per_dim_span_[qid] : 0.0f;
```
liliu-z, line 721: The suffix upper bound is computed as `qval * dim_max`, but for signed (negative) query values this product clamps to 0 instead of the true maximum contribution, so the accumulated suffix bound under-counts the achievable score and is no longer a valid upper bound. As a result the early-termination check prunes documents that could still rank into the top-k, returning incomplete results for any query containing negative IP values. Use a sign-correct extremum so the suffix bound never under-counts negative contributions.

src/index/sparse/sindi_inverted_index.h
```diff
@@ -258,6 +236,60 @@ class SindiInvertedIndex : public InvertedIndex<DataType> {
             row_sums_span_ = boost::span<const float>(row_sums_.data(), row_sums_.size());
         }
 
+        // Compute max score per dimension for early termination optimization
+        // For IP: max_score = max(val) across all postings
+        // For BM25: max_score = max((k1+1)*tf / (tf + k1*(1-b+b*dl/avgdl))) for each posting
+        max_scores_per_dim_.resize(this->dim_map_.size());
```
liliu-z, line 245: Each `add()` on the CC incremental path recomputes the per-dimension max scores by rescanning all previously inserted vectors (lines 242–289) rather than updating incrementally, so building an n-vector index through repeated Add costs O(n^2). On the concurrent/incremental Add path this makes growth quadratic and large builds degrade severely. Maintain the running max incrementally as rows are added instead of full-rescanning on every call.

src/index/sparse/sindi_simd_sve.cc
```diff
@@ -0,0 +1,281 @@
...
+            v_max = svmax_f32_m(pg32_odd, v_max, vsum_odd);
+        }
+    }
+    return svmaxv_f32(svptrue_b32(), v_max);
+}
+
+namespace {
+
+svfloat32_t
+sv_fast_recip_f32(svbool_t pg, svfloat32_t x) {
```
liliu-z: The SVE BM25 path uses an approximate reciprocal estimate instead of exact division, so per-element scores on ARM differ slightly from the x86 path. This is usually benign for ranking but yields non-bitwise-identical scores across architectures, which can break cross-platform score comparisons and exact-match tests. Use exact division or add a Newton-Raphson refinement step if score parity with x86 is required.

src/index/sparse/sindi_inverted_index.h
sparknack: we have made sure the qval is postive.

src/index/sparse/sindi_inverted_index.h
sparknack: We can still introduce breaking changes prior to the release of Milvus 3.0.

src/index/sparse/sindi_simd_sve.cc
sparknack: done.

src/index/sparse/sindi_inverted_index.h
sparknack: done.

chasingegg (approved):
/lgtm

PR #1666, opened 2026-06-04
Add SVS leanvec ood support

ethanglaser:
Adds knowhere wrapper over SVS LeanVec OOD, which is present in FAISS


ethanglaser:
On hold for benchmarking

PR #1672, opened 2026-06-10
feat: support nullable external id mapping

marcelo-cjl:
## Summary
- Add ExternalIdMap to centralize nullable and relayout id mapping.
- Wire mapped filtering and one-pass result id mapping across Knowhere index nodes, brute force, iterators, and vector retrieval paths.
- Add nullable external id map unit coverage.

## Test plan
- [x] `git diff --check HEAD~1..HEAD`
- [ ] Full compile/tests not rerun in this PR creation step.

### DIFF
--- a/include/knowhere/bitsetview.h
+++ b/include/knowhere/bitsetview.h
@@ -13,12 +13,23 @@
 #define BITSET_H
 
 #include <cassert>
+#include <cstddef>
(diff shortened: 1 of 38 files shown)


marcelo-cjl:
Closed in favor of fork-based PR #1673.

PR #1673, opened 2026-06-10
enhance: support nullable external id mapping

marcelo-cjl:
issue: #1687

related: zilliztech/cardinal#859
related: zilliztech/cardinal#860
related: milvus-io/milvus#50524

## Summary
- Move nullable id mapping into Knowhere's index layer. Public row/list ids remain the API, bitset, selected-id, and result boundary; backend storage can use compact internal ids.
- Add `DataSet::IdMapData` so `Build` and `Add` consume nullable mapping input together with the vector payload.
- Split id-map storage by lifecycle: sealed maps publish complete contiguous arrays with optional mmap, while growing maps append ranges and preserve existing backend-id labels.
- Use adaptive public-to-internal lookup so sparse public-id domains do not force dense heap storage unless mmap or growing storage requires it.
- Project public bitsets into backend/local id domains and map backend results back to public ids across search, range search, iterators, retrieval, data-view refinement, HNSW multi-index, DiskANN/AiSAQ, SVS, sparse, cuVS, and EmbList paths.
- Let brute-force paths use `BitsetView::out_ids` so compact raw buffers filter on public ids and return public ids without building a Knowhere `IdMap`.
- Add a design doc and nullable id-map coverage for build/add/reload/mmap/search/range/iterator/retrieve/brute-force/EmbList cases.

## Verification
- [x] `git diff --check upstream/main..HEAD`
- [x] `cmake --build build/Release --target knowhere_tests -j 8` after rebasing onto `upstream/main` `3b78cba8`
- [x] Full local CPU UT coverage was triaged for this final diff shape. The only non-passing signature was the known OpenBLAS HNSW_PRQ/EmbList allocator issue; with `OPENBLAS_NUM_THREADS=16`, the focused nullable HNSW_PRQ matrix and remaining non-redundant cases passed.
- [x] Milvus dependency-overlay validation against the nullable id-map branch: full `make test-cpp jobs="$(nproc)"` passed; `all_tests` 8212 passed / 10 skipped, cwrapper/json-stats 79 passed.

### DIFF
--- a/include/knowhere/array_store.h
+++ b/include/knowhere/array_store.h
@@ -1,0 +1,565 @@
+// Copyright (C) 2019-2026 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#ifndef ARRAY_STORE_H
+#define ARRAY_STORE_H
+
+#include <algorithm>
+#include <array>
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "knowhere/mmap.h"
+
+namespace knowhere {
+
+// Contiguous sealed array allocation, heap-backed or mmap-backed.
+template <typename T>
+struct ArrayData {
+    explicit ArrayData(size_t size) : size(size), data(std::make_unique<T[]>(size)) {
+    }
+
+    ArrayData(const T* view_data, size_t size) : size(size), view_data(view_data) {
+    }
+
+    ArrayData(size_t size, const std::string& filepath)
+        : size(size), mmap_region(MmapRegion::Create(filepath, size * sizeof(T))) {
+    }
+
+    T*
+    mutable_data() {
+        if (view_data != nullptr) {
+            throw std::runtime_error("array data view is read only");
+        }
+        return mmap_region == nullptr ? data.get() : static_cast<T*>(mmap_region->data());
+    }
+
+    const T*
+    data_ptr() const {
+        if (view_data != nullptr) {
+            return view_data;
+        }
+        return mmap_region == nullptr ? data.get() : static_cast<const T*>(mmap_region->data());
+    }
+
+    size_t size;
+    const T* view_data = nullptr;
+    std::unique_ptr<T[]> data;
+    std::shared_ptr<MmapRegion> mmap_region;
+};
+
+template <typename T>
+class AppendArrayData {
+ public:
+    AppendArrayData() {
+        for (auto& chunk : chunk_ptrs_) {
+            chunk.store(nullptr, std::memory_order_relaxed);
+        }
+    }
+
+    AppendArrayData(const AppendArrayData&) = delete;
+    AppendArrayData&
+    operator=(const AppendArrayData&) = delete;
+
+    void
+    Append(const T* data, size_t count) {
+        const auto begin = committed_count_.load(std::memory_order_acquire);
+        if (count > std::numeric_limits<size_t>::max() - begin) {
+            throw std::runtime_error("append array size overflows");
+        }
+        const auto end = begin + count;
+        EnsureCapacity(end);
+
+        auto remaining = count;
+        auto offset = begin;
+        auto* source = data;
+        while (remaining != 0) {
+            const auto chunk_id = ChunkIndex(offset);
+            const auto chunk_offset = offset - ChunkBegin(chunk_id);
+            const auto write_count = std::min(remaining, ChunkSize(chunk_id) - chunk_offset);
+            auto* chunk = chunk_ptrs_[chunk_id].load(std::memory_order_acquire);
+            std::copy(source, source + write_count, chunk + chunk_offset);
+            source += write_count;
+            offset += write_count;
+            remaining -= write_count;
+        }
+
+        committed_count_.store(end, std::memory_order_release);
+    }
+
+    size_t
+    size() const {
+        return committed_count_.load(std::memory_order_acquire);
+    }
+
+    T
+    Get(size_t offset) const {
+        const auto chunk_id = ChunkIndex(offset);
+        const auto* chunk = chunk_ptrs_[chunk_id].load(std::memory_order_acquire);
+        return chunk[offset - ChunkBegin(chunk_id)];
+    }
+
+ private:
+    static constexpr size_t kFirstChunkBits = 10;
+    static constexpr size_t kFirstChunkSize = size_t{1} << kFirstChunkBits;
+    static constexpr size_t kMaxChunks = std::numeric_limits<size_t>::digits - kFirstChunkBits;
+
+    static size_t
+    ChunkIndex(size_t offset) {
+        const auto block = (offset >> kFirstChunkBits) + 1;
+        const auto chunk_id = static_cast<size_t>(std::numeric_limits<unsigned long long>::digits - 1 -
+                                                  __builtin_clzll(static_cast<unsigned long long>(block)));
+        if (chunk_id >= kMaxChunks) {
+            throw std::runtime_error("append array capacity is exhausted");
+        }
+        return chunk_id;
+    }
+
+    static size_t
+    ChunkBegin(size_t chunk_id) {
+        return (kFirstChunkSize << chunk_id) - kFirstChunkSize;
+    }
+
+    static size_t
+    ChunkSize(size_t chunk_id) {
+        return kFirstChunkSize << chunk_id;
+    }
+
+    void
+    EnsureCapacity(size_t count) {
+        if (count == 0) {
+            return;
+        }
+        const auto last_chunk_id = ChunkIndex(count - 1);
+        for (size_t chunk_id = 0; chunk_id <= last_chunk_id; ++chunk_id) {
+            if (chunks_[chunk_id] != nullptr) {
+                continue;
+            }
+            chunks_[chunk_id] = std::make_unique<T[]>(ChunkSize(chunk_id));
+            chunk_ptrs_[chunk_id].store(chunks_[chunk_id].get(), std::memory_order_release);
+        }
+    }
+
+    std::array<std::unique_ptr<T[]>, kMaxChunks> chunks_;
+    std::array<std::atomic<T*>, kMaxChunks> chunk_ptrs_;
+    std::atomic<size_t> committed_count_{0};
+};
+
+template <typename T>
+class ArrayStore {
+ public:
+    enum class Type {
+        ARRAY,
+        APPEND_ARRAY,
+    };
+
+    // ARRAY stores a complete sealed array through Set(). APPEND_ARRAY stores
+    // growing data through append-only chunks.
+    ArrayStore() = default;
+    ArrayStore(const T* data, size_t count) {
+        SetView(data, count);
+    }
+    ArrayStore(const ArrayStore& other) {
+        CopyFrom(other);
+    }
+    ArrayStore(ArrayStore&&) noexcept = default;
+    ArrayStore&
+    operator=(const ArrayStore& other) {
+        if (this != &other) {
+            CopyFrom(other);
+        }
+        return *this;
+    }
+    ArrayStore&
+    operator=(ArrayStore&&) noexcept = default;
+
+    void
+    SetType(Type type) {
+        type_ = type;
+        if (type_ == Type::APPEND_ARRAY && append_array_ == nullptr) {
+            append_array_ = std::make_shared<AppendArrayData<T>>();
+        }
+        visible_count_ = kDynamicCount;
+    }
+
+    void
+    SetView(const T* data, size_t count) {
+        if (count != 0 && data == nullptr) {
+            throw std::runtime_error("array store view data is null");
+        }
+        Clear();
+        type_ = Type::ARRAY;
+        if (count == 0) {
+            return;
+        }
+        array_ = std::make_shared<ArrayData<T>>(data, count);
+    }
+
+    void
+    Set(const T* data, size_t count, const std::string& filepath = std::string{}) {
+        if (count != 0 && data == nullptr) {
+            throw std::runtime_error("array store data is null");
+        }
+        Clear();
+        type_ = Type::ARRAY;
+        if (count == 0) {
+            return;
+        }
+        array_ = NewArray(count, filepath);
+        for (size_t i = 0; i < count; ++i) {
+            array_->mutable_data()[i] = data[i];
+        }
+    }
+
+    void
+    Append(const T* data, size_t count) {
+        if (count != 0 && data == nullptr) {
+            throw std::runtime_error("array store data is null");
+        }
+        if (count == 0) {
+            return;
+        }
+        type_ = Type::APPEND_ARRAY;
+        if (append_array_ == nullptr) {
+            append_array_ = std::make_shared<AppendArrayData<T>>();
+        }
+        visible_count_ = kDynamicCount;
+        append_array_->Append(data, count);
+    }
+
+    bool
+    empty() const {
+        return size() == 0;
+    }
+
+    size_t
+    size() const {
+        const auto storage_count = StorageSize();
+        return visible_count_ == kDynamicCount ? storage_count : std::min(visible_count_, storage_count);
+    }
+
+    bool
+    is_array() const {
+        return type_ == Type::ARRAY;
+    }
+
+    bool
+    is_append_array() const {
+        return type_ == Type::APPEND_ARRAY;
+    }
+
+    const T*
+    data() const {
+        return type_ == Type::ARRAY && array_ != nullptr ? array_->data_ptr() : nullptr;
+    }
+
+    T
+    operator[](size_t offset) const {
+        if (type_ == Type::ARRAY) {
+            return array_->data_ptr()[offset];
+        }
+        return append_array_->Get(offset);
+    }
+
+    T&
+    operator[](size_t offset) {
+        if (type_ == Type::ARRAY) {
+            return array_->mutable_data()[offset];
+        }
+        throw std::runtime_error("append array storage is read only");
+    }
+
+    void
+    Clear() {
+        array_.reset();
+        append_array_.reset();
+        visible_count_ = kDynamicCount;
+    }
+
+    ArrayStore
+    Prefix(size_t count) const {
+        ArrayStore copy(*this);
+        copy.visible_count_ = std::min(count, copy.size());
+        return copy;
+    }
+
+ private:
+    static constexpr size_t kDynamicCount = std::numeric_limits<size_t>::max();
+
+    void
+    CopyFrom(const ArrayStore& other) {
+        type_ = other.type_;
+        array_ = other.array_;
+        append_array_ = other.append_array_;
+        visible_count_ = other.size();
+    }
+
+    static std::shared_ptr<ArrayData<T>>
+    NewArray(size_t capacity, const std::string& filepath) {
+        if (filepath.empty()) {
+            return std::make_shared<ArrayData<T>>(capacity);
+        }
+        return std::make_shared<ArrayData<T>>(capacity, filepath);
+    }
+
+    size_t
+    StorageSize() const {
+        if (type_ == Type::ARRAY) {
+            return array_ == nullptr ? 0 : array_->size;
+        }
+        return append_array_ == nullptr ? 0 : append_array_->size();
+    }
+
+    Type type_ = Type::ARRAY;
+    std::shared_ptr<ArrayData<T>> array_;
+    std::shared_ptr<AppendArrayData<T>> append_array_;
+    size_t visible_count_ = kDynamicCount;
+};
+
+using IdArray = ArrayStore<int32_t>;
+
+struct BitmapRecord {
+    size_t bit_begin = 0;
+    size_t bit_count = 0;
+    std::shared_ptr<const std::vector<uint8_t>> bytes;
+
+    bool
+    Contains(size_t bit) const {
+        return bit >= bit_begin && bit < bit_begin + bit_count;
+    }
+
+    bool
+    Test(size_t bit) const {
+        const auto local_bit = bit - bit_begin;
+        return ((*bytes)[local_bit >> 3] & (1U << (local_bit & 7))) != 0;
+    }
+};
+
+class AppendBitmapData {
+ public:
+    AppendBitmapData() {
+        records_.SetType(ArrayStore<BitmapRecord>::Type::APPEND_ARRAY);
+    }
+
+    void
+    Append(const uint8_t* data, size_t bit_count) {
+        const auto bit_begin = bit_count_.load(std::memory_order_acquire);
+        const auto bytes = std::make_shared<const std::vector<uint8_t>>(data, data + ByteSize(bit_count));
+        const BitmapRecord record{bit_begin, bit_count, bytes};
+        records_.Append(&record, 1);
+        bit_count_.store(bit_begin + bit_count, std::memory_order_release);
+    }
+
+    size_t
+    size() const {
+        return bit_count_.load(std::memory_order_acquire);
+    }
+
+    uint8_t
+    GetByte(size_t byte_offset) const {
+        const auto bit_begin = byte_offset << 3;
+        const auto bit_end = size();
+        uint8_t value = 0;
+        for (size_t bit = 0; bit < 8; ++bit) {
+            const auto absolute_bit = bit_begin + bit;
+            if (absolute_bit >= bit_end) {
+                break;
+            }
+            if (Test(absolute_bit)) {
+                value |= static_cast<uint8_t>(1U << bit);
+            }
+        }
+        return value;
+    }
+
+ private:
+    static size_t
+    ByteSize(size_t bit_count) {
+        return (bit_count + 7) / 8;
+    }
+
+    bool
+    Test(size_t bit) const {
+        auto left = static_cast<size_t>(0);
+        auto right = records_.size();
+        while (left < right) {
+            const auto middle = left + (right - left) / 2;
+            if (records_[middle].bit_begin <= bit) {
+                left = middle + 1;
+            } else {
+                right = middle;
+            }
+        }
+        if (left == 0) {
+            return false;
+        }
+
+        const auto record = records_[left - 1];
+        return record.Contains(bit) && record.Test(bit);
+    }
+
+    ArrayStore<BitmapRecord> records_;
+    std::atomic<size_t> bit_count_{0};
+};
+
+// Packed public-id validity bitmap. size() returns the logical bit count.
+class BitmapArray {
+ public:
+    using Type = ArrayStore<uint8_t>::Type;
+
+    BitmapArray() {
+        bytes_.SetType(Type::ARRAY);
+    }
+
+    BitmapArray(const BitmapArray& other) {
+        CopyFrom(other);
+    }
+
+    BitmapArray(BitmapArray&&) noexcept = default;
+
+    BitmapArray&
+    operator=(const BitmapArray& other) {
+        if (this != &other) {
+            CopyFrom(other);
+        }
+        return *this;
+    }
+
+    BitmapArray&
+    operator=(BitmapArray&&) noexcept = default;
+
+    void
+    SetType(Type type) {
+        type_ = type;
+        if (type_ == Type::ARRAY) {
+            bytes_.SetType(Type::ARRAY);
+            bit_count_ = 0;
+            return;
+        }
+        if (append_data_ == nullptr) {
+            append_data_ = std::make_shared<AppendBitmapData>();
+        }
+        bit_count_ = kDynamicBitCount;
+    }
+
+    const uint8_t*
+    data() const {
+        return type_ == Type::ARRAY && bit_count_ != 0 ? bytes_.data() : nullptr;
+    }
+
+    size_t
+    size() const {
+        if (type_ == Type::ARRAY) {
+            return bit_count_;
+        }
+        if (append_data_ == nullptr) {
+            return 0;
+        }
+        return bit_count_ == kDynamicBitCount ? append_data_->size() : bit_count_;
+    }
+
+    bool
+    empty() const {
+        return size() == 0;
+    }
+
+    uint8_t
+    operator[](size_t offset) const {
+        return type_ == Type::ARRAY ? bytes_[offset] : append_data_->GetByte(offset);
+    }
+
+    void
+    Set(const uint8_t* data, size_t bit_count) {
+        Clear();
+        if (bit_count == 0) {
+            return;
+        }
+        if (data == nullptr) {
+            throw std::runtime_error("bitmap array data is null");
+        }
+        type_ = Type::ARRAY;
+        bytes_.Set(data, ByteSize(bit_count));
+        bit_count_ = bit_count;
+        MaskTail();
+    }
+
+    void
+    Append(const uint8_t* data, size_t bit_count) {
+        // Append is bit-oriented; batches may end in the middle of a byte.
+        if (bit_count == 0) {
+            return;
+        }
+        if (data == nullptr) {
+            throw std::runtime_error("bitmap array data is null");
+        }
+        type_ = Type::APPEND_ARRAY;
+        if (append_data_ == nullptr) {
+            append_data_ = std::make_shared<AppendBitmapData>();
+        }
+        bit_count_ = kDynamicBitCount;
+        append_data_->Append(data, bit_count);
+    }
+
+    void
+    Clear() {
+        bytes_.Clear();
+        bit_count_ = 0;
+        append_data_.reset();
+    }
+
+    BitmapArray
+    Prefix(size_t bit_count) const {
+        BitmapArray copy(*this);
+        copy.bit_count_ = std::min(bit_count, copy.size());
+        return copy;
+    }
+
+ private:
+    static constexpr size_t kDynamicBitCount = std::numeric_limits<size_t>::max();
+
+    static size_t
+    ByteSize(size_t bit_count) {
+        return (bit_count + 7) / 8;
+    }
+
+    void
+    CopyFrom(const BitmapArray& other) {
+        bytes_ = other.bytes_;
+        type_ = other.type_;
+        bit_count_ = type_ == Type::APPEND_ARRAY ? other.size() : other.bit_count_;
+        append_data_ = other.append_data_;
+    }
+
+    void
+    MaskTail() {
+        const auto used_bits = bit_count_ & 7U;
+        if (used_bits == 0 || bit_count_ == 0) {
+            return;
+        }
+        const auto byte = ByteSize(bit_count_) - 1;
+        bytes_[byte] = static_cast<uint8_t>(bytes_[byte] & static_cast<uint8_t>((1U << used_bits) - 1U));
+    }
+
+    ArrayStore<uint8_t> bytes_;
+    Type type_ = Type::ARRAY;
+    size_t bit_count_ = 0;
+    std::shared_ptr<AppendBitmapData> append_data_;
+};
+
+}  // namespace knowhere
+
+#endif /* ARRAY_STORE_H */
--- a/include/knowhere/array_store.h
+++ b/include/knowhere/array_store.h
@@ -233,11 +233,9 @@ class ArrayStore {
         if (count == 0) {
             return;
         }
-        type_ = Type::APPEND_ARRAY;
-        if (append_array_ == nullptr) {
-            append_array_ = std::make_shared<AppendArrayData<T>>();
+        if (type_ != Type::APPEND_ARRAY || append_array_ == nullptr || visible_count_ != kDynamicCount) {
+            throw std::runtime_error("array store append requires append storage");
         }
-        visible_count_ = kDynamicCount;
         append_array_->Append(data, count);
     }
 
@@ -506,11 +504,9 @@ class BitmapArray {
         if (data == nullptr) {
             throw std::runtime_error("bitmap array data is null");
         }
-        type_ = Type::APPEND_ARRAY;
-        if (append_data_ == nullptr) {
-            append_data_ = std::make_shared<AppendBitmapData>();
+        if (type_ != Type::APPEND_ARRAY || append_data_ == nullptr || bit_count_ != kDynamicBitCount) {
+            throw std::runtime_error("bitmap array append requires append storage");
         }
-        bit_count_ = kDynamicBitCount;
         append_data_->Append(data, bit_count);
     }
 
--- a/include/knowhere/bitsetview.h
+++ b/include/knowhere/bitsetview.h
@@ -15,19 +15,33 @@
 #include <algorithm>
 #include <bit>
 #include <cassert>
+#include <cstddef>
 #include <cstdint>
+#include <cstring>
+#include <limits>
 #include <optional>
 #include <sstream>
+#include <stdexcept>
 #include <string>
+#include <utility>
+
+#include "knowhere/array_store.h"
 
 namespace knowhere {
+
+// Non-owning filter view.
+//
+// bits_ is always addressed by public ids. Backend selectors pass dense
+// internal/local ids to test(); id_offset_ handles contiguous windows and
+// out_ids_ handles compaction or backend relayout. Count fields are expressed
+// in the backend vector domain used by the selector.
 class BitsetView {
  public:
     BitsetView() = default;
     ~BitsetView() = default;
 
-    BitsetView(const uint8_t* data, size_t num_bits, size_t num_filtered_out_bits = 0, size_t id_offset = 0)
-        : bits_(data), num_bits_(num_bits), num_filtered_out_bits_(num_filtered_out_bits), id_offset_(id_offset) {
+    BitsetView(const uint8_t* data, size_t num_bits, std::optional<size_t> filtered_count = std::nullopt)
+        : bits_(data), num_bits_(num_bits), vector_count_(num_bits), filtered_count_(filtered_count) {
     }
 
     BitsetView(const std::nullptr_t) : BitsetView() {
@@ -35,25 +49,34 @@ class BitsetView {
 
     bool
     empty() const {
-        return num_bits_ == 0;
+        if (num_bits_ == 0) {
+            return true;
+        }
+        if (!filtered_count_.has_value() || filtered_count_.value() != 0) {
+            return false;
+        }
+        return !has_id_boundary_filter_();
     }
 
-    // return the number of the bits. if with id mapping, return the number of the internal ids.
     size_t
     size() const {
-        if (out_ids_ != nullptr) {
-            return num_internal_ids_;
-        }
-        return num_bits_;
+        return vector_count_;
     }
 
-    // return the number of filtered out bits. if with id mapping, return the number of filtered out ids.
+    bool
+    has_known_count() const {
+        return num_bits_ == 0 || filtered_count_.has_value();
+    }
+
     size_t
     count() const {
-        if (out_ids_ != nullptr) {
-            return num_filtered_out_ids_;
+        if (num_bits_ == 0) {
+            return 0;
         }
-        return num_filtered_out_bits_;
+        if (!filtered_count_.has_value()) {
+            throw std::logic_error("BitsetView filtered count is unknown");
+        }
+        return filtered_count_.value();
     }
 
     size_t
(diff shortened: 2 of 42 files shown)


foxspy:
I think the ID ownership boundary here needs to be made explicit.

`IndexNodeWithDataViewRefiner` is the public `IndexNode` boundary, so public APIs should consume/return external labels. But `DataViewIndexFlat` is an internal refiner and its subset-distance path indexes directly into the data view, so `SearchWithIds()` / `RangeSearchWithIds()` / `ComputeDistanceSubset()` should operate on internal row offsets only.

Right now the refiner path mixes the two semantics: the ids are used directly by `ComputeDistanceSubset()` as internal offsets, while the cosine norm path later calls `external_id_map_.ToInternalIds()` as if the same ids were external labels. For nullable or non-identity mappings this can use the wrong norm id, wrong data-view row, or return labels with inconsistent mapping semantics.

Can we keep the refiner contract internal-only and move all external/internal conversion to `IndexNodeWithDataViewRefiner`? In other words:
- map external filters before calling the base index;
- keep base/refiner candidate ids as internal offsets;
- map final result labels to external ids exactly once before returning from the public `IndexNode` API.

foxspy:
The `ExternalIdMap` wire format looks incomplete for the runtime state it now represents.

`Serialize()` currently persists only `external_count_` and `internal_to_external_ids_`, but the map can also carry `external_id_offset_` and `internal_id_to_emb_list_id_`. After save/load, those fields are reset by `Clear()` and never restored, so indexes relying on identity + non-zero offset or emb-list/DVR mapping can silently return different labels after reload.

Can we persist all required source-of-truth fields and rebuild only derived state during `Deserialize()`? For example:
- persist `external_id_offset_`;
- persist `internal_id_to_emb_list_id_` when present;
- rebuild `external_to_internal_ids_` and `valid_bitmap_` from the persisted data;
- add save/load regression tests for non-zero offset, nullable mapping, same-size non-identity mapping, and emb-list mapping.

foxspy:
`GetExternalIdMap(true)` looks expensive to call on the BruteForce search hot path.

`DataSet::GetExternalIdMap()` constructs a fresh `ExternalIdMap` every time. When an explicit internal-to-external mapping is present, `SetInternalToExternalIds()` rebuilds the derived state, including the inverse `unordered_map` and the valid bitmap. The BruteForce paths call this per request (`SearchWithBuf`, range/search variants, and iterator setup), so the request now pays an extra O(nb) map/bitmap construction cost before the actual brute-force computation starts.

This may be acceptable for very small datasets, but for large `base_dataset` inputs it is a material latency and allocation regression, especially because it is independent of `nq`/`topk` and repeated for every call.

Can we avoid rebuilding this per search? A few options:
- cache the constructed `ExternalIdMap` inside `DataSet` and invalidate it only when the source ids/bitmap/offset changes;
- pass a stable `ExternalIdMap` snapshot/reference with the dataset;
- or provide benchmark data showing the per-request rebuild cost is negligible for the intended BruteForce workloads.

foxspy:
`EnsureBitsetCount()` adds another full scan on filtered search paths.

For every non-empty bitset, `Index::Search()` / `RangeSearch()` / `AnnIterator()` call `EnsureBitsetCount()` before dispatching to the index node. In the normal path this calls `bitset.count_filtered_bits(valid_bitmap)`, which scans the full bitset byte range and optionally ANDs with the valid bitmap. In the emb-list path, `CountEmbListBitset()` loops over every internal doc id and calls `bitset.test()`.

So filtered search now has an extra O(bitset_size) or O(num_emb_lists) pass per request before the actual index search. That could be significant for large collections and high-QPS filtered workloads, even when `nq`/`topk` are small.

Can we either avoid this full count on the hot path or quantify the overhead? Possible approaches:
- cache the mapped filtered count together with the mapped bitset;
- compute the count lazily only for index implementations that actually need it;
- preserve an "unknown count" state instead of forcing an exact count everywhere;
- add benchmark coverage for large bitsets / nullable maps / emb-list filters to show the overhead is acceptable.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -1296,35 +1345,18 @@ class BaseFaissRegularIndexHNSWNode : public BaseFaissRegularIndexNode {
...
-            // calculate more accurate filter statistics for the single mv-index.
-            size_t num_mv_ids = labels[index_id].get()->size();
-            size_t num_mv_filtered_out_ids = num_mv_ids - (bitset.size() - bitset.count());
-            if (!bitset.has_out_ids()) {
-                bitset.set_out_ids(labels[index_id].get()->data(), num_mv_ids, num_mv_filtered_out_ids);
-            } else {
-                bitset.set_out_ids(internal_offset_to_most_external_id.data(), num_mv_ids, num_mv_filtered_out_ids);
+            const auto num_mv_ids = labels[index_id].get()->size();
+            if (bitset.has_out_ids()) {
+                bitset.set_out_ids(bitset.out_ids_data(), num_mv_ids);
```
liliu-z: In the multi-index (MV / emb-list partitioned) branch, set_out_ids(bitset.out_ids_data(), num_mv_ids) re-invokes set_out_ids, which unconditionally sets num_filtered_out_ids_ = num_filtered_out_bits_ (bitsetview.h:154). For emb-list/relayout multi-index HNSW — an exact-count index — PrepareBitset has already computed the exact filtered count into num_filtered_out_ids_, while num_filtered_out_bits_ still holds the caller's pre-mapping value (0 when the input bitset's count was never pre-populated, which is valid for an exact-count index since PrepareBitset would otherwise compute it). The re-set_out_ids therefore discards the prepared count: count() becomes 0 and need_filter() disables filtering, so emb-list-partitioned multi-index search returns unfiltered results. The identical reset recurs at faiss_hnsw.cc:1691 and 2037.

include/knowhere/external_id_map.h
```diff
@@ -0,0 +1,660 @@
...
+            if (emb_list_count_value == 0) {
+                return;
+            }
+            if (emb_list_offsets == nullptr) {
+                throw std::runtime_error("invalid emb list id map data");
+            }
+            if (emb_list_offsets[0] != 0) {
+                throw std::runtime_error("invalid emb list id map data");
+            }
+            if (internal_to_external_ids_.size() < emb_list_count_value) {
```
liliu-z: Base FinalizeExternalIdMap calls BuildEmbListIds, which throws here when internal_to_external_ids_.size() < emb_list_count. On a non-nullable build nothing seeds internal_to_external_ids_: BuildIdsFromValidBitmap is a no-op because valid_bitmap_ is empty, so the vector stays size 0 and any index using the base path with a TokenANN strategy (NeedsBaseIndexIDMap()==true) throws. DiskANN (never seeds the map; uses the base FinalizeExternalIdMap) and single-index HNSW (HasRelayoutIdMap()==(indexes.size()>1)==false, which routes to the base) both hit this on ordinary supported emb-list configs; IVF is exempt because its emb-list add seeds the map via AddEmbListIdsAndBitmap. The throw is uncaught in Index::Build so the build fails. Note test_emb_list.cc does exercise the failing path: with MV_ONLYs={false,true} the mv_only_enable=false case builds single-index HNSW with the default (empty -> TokenANN) strategy and would itself throw. Seed an identity map on the non-nullable path, or make BuildEmbListIds a no-op when unmapped.

include/knowhere/index/index_node.h
```diff
@@ -232,6 +241,107 @@ class IndexNode : public Object {
...
+
+    virtual BitsetIdArrayView
+    GetInternalToExternalIdsForBitset() const {
+        auto ids = external_id_map_.GetInternalToExternalIds(false);
+        return {ids.data, ids.size};
+    }
+
+    virtual BitsetIdArrayView
+    GetInternalToEmbListIdsForBitset() const {
+        auto ids = external_id_map_.GetInternalToEmbListIds(false);
```
liliu-z: use_lock_ is never enabled on any ExternalIdMap (no caller constructs ExternalIdMap(true)), so the internal shared_mutex is inert and accessors hand out zero-copy pointers. GetInternalToEmbListIdsForBitset/GetInternalToExternalIdsForBitset force copy_data=false, and PrepareBitset — which runs in the Index::Search wrapper (index.cc:148) before IvfIndexNode::Search acquires base_index_lock_ — installs those raw pointers as the bitset's out_ids; AnnIterator takes no lock at all and its iterators hold the map by reference. A concurrent AddEmbListIdsAndBitmap (under the write lock) resize()s internal_id_to_emb_list_id_/internal_to_external_ids_, reallocating that storage and leaving the captured out_ids and iterator views dangling — a use-after-free on the concurrent emb-list add+search path. base_index_lock_ does not cover the pointer capture, and enabling use_lock_ alone is insufficient because the *ForBitset getters force zero-copy; the prepared data must be copied or the lock held across capture-and-use.

include/knowhere/external_id_map.h
```diff
@@ -0,0 +1,660 @@
...
+                throw std::runtime_error("external id map changed while appending ids");
+            }
+            external_count_ += external_count_value;
+            valid_bitmap_.resize((external_count_ + 7) / 8, 0);
+
+            internal_to_external_ids_.resize(internal_id_begin + count_value);
+            for (size_t i = 0; i < count_value; ++i) {
+                const auto external_id = new_internal_to_external_ids[i];
+                internal_to_external_ids_[internal_id_begin + i] = external_id;
+                external_to_internal_ids_.Add(external_id, static_cast<int32_t>(internal_id_begin + i));
```
liliu-z: AddIdsAndBitmap's commit block updates external_count_ (line 539), valid_bitmap_ (540), and internal_to_external_ids_ (542), and then in the loop calls external_to_internal_ids_.Add, which throws 'external id map contains duplicate id' when the input ids contain a repeat. Because that duplicate check runs only after the structure has been partially mutated, a malformed batch leaves the map half-written (inflated external_count_, partially set bitmap and id vectors). AddEmbListIdsAndBitmap (commit block around line 631) has the same ordering. The IVF/refiner callers catch the exception and keep the index object live, so a CC index can keep serving on a corrupted map. Validate for duplicates before mutating, or roll back on throw.

src/index/hnsw/faiss_hnsw.cc
marcelo-cjl: Current head no longer has this failure mode. `BitsetView::set_out_ids()` only installs the mapped id view now; it does not reset the filtered count. The multi-index path recalculates the local sub-index count in `PrepareBitsetForSubIndex()` after applying the partition-local out-id map. Nullable HNSW multi-index search/range/iterator and emb-list multi-index cases are covered.

include/knowhere/external_id_map.h
marcelo-cjl: Outdated after the `IdMap` refactor. `FinalizeIdMap()` builds ids from the valid bitmap only when needed, and `BuildInToOutEblIds()` treats an empty `in_to_out_ids` as identity, so non-nullable TokenANN does not throw. The TokenANN HNSW/DiskANN build path and GetEmbListByIds cases are covered.

include/knowhere/index/index_node.h
marcelo-cjl: Fixed by snapshot ownership, not by holding the index lock through search. `PrepareBitset()` captures an `IdMapSnapshot` backed by shared buffers, and iterator paths keep the `PreparedBitset` alive while the iterator uses the mapped bitset. Concurrent growing add/search is handled by enabling `IdMap::SetUseLock(true)` on the Milvus side; immutable indexes do not need the lock.

include/knowhere/external_id_map.h
marcelo-cjl: This is invalid input, not a supported recovery path. `AddIdsAndBitmap()` assumes the batch out ids are unique; Milvus builds them from valid rows/bitmap, which guarantees uniqueness. I do not want to make this internal map update transactional for malformed input and add another duplicate pre-scan. The existing duplicate check is diagnostic.

marcelo-cjl:
@foxspy Re DataView refiner id boundary: current head keeps `DataViewIndexFlat` internal-only. `SearchWithIds()` / `RangeSearchWithIds()` / `CalcDistByIDs()` consume internal row offsets; external-to-internal conversion is only done in `IndexNodeWithDataViewRefiner`, and final results are mapped to out ids once before returning. Tests cover selected-id APIs and cosine norm ids.

marcelo-cjl:
@foxspy Re IdMap wire format: current design changed here. `IdMap` is runtime-only and is not serialized into the index BinarySet. The caller persists valid-row state externally and sets the runtime map before build/load. Tests assert `EXTERNAL_ID_MAP` is not written and reload uses the runtime bitmap/map.

marcelo-cjl:
@foxspy Re BruteForce hot path cost: outdated after the refactor. `GetExternalIdMap()` is gone, and BF does not construct an id map per request. It only applies the existing `BitsetView` offset/count for the searched range.

marcelo-cjl:
@foxspy Re bitset count scan: this is not an extra scan before another scan. Exact-count indexes already need `bitset.count()` / `filter_ratio()` later (HNSW BF fallback/kAlpha, IVF max_codes, etc.). `PrepareBitset()` computes that once up front; downstream reads the stored count. Indexes that do not need exact count skip it.

alexanderguzhva:
@marcelo-cjl is this PR ready to be reviewed?

include/knowhere/bitsetview.h
```diff
@@ -211,19 +271,26 @@ class BitsetView {
...
+        uint64_t value = 0;
+        std::memcpy(&value, data, sizeof(value));
+        return value;
+    }
+
     const uint8_t* bits_ = nullptr;
     size_t num_bits_ = 0;
-    size_t num_filtered_out_bits_ = 0;
+    size_t vector_count_ = 0;
+    int64_t filtered_count_ = -1;
```
alexanderguzhva: please make `size_t filtered_count_ = 0;` instead. I see no reason of it being `int64_t`

include/knowhere/bitsetview.h
```diff
@@ -211,19 +271,26 @@ class BitsetView {
...
 
     // optional. many indexes will share one bitset, requiring offset to distinguish between them.
     //  like multi-chunk brute-force in /src/common/comp/brute_force.cc, or mv-only in /src/index/hnsw/faiss_hnsw.cc
-    size_t id_offset_ = 0;  // offset of the internal ids
+    int64_t id_offset_ = 0;  // offset of the internal ids
```
alexanderguzhva: please make `size_t id_offset_ = 0;` instead. I see no reason of it being `int64_t`

include/knowhere/bitsetview.h
```diff
@@ -12,186 +12,246 @@
...
+        if (bits_ == nullptr || num_bits_ == 0 || bit_count == 0 || bit_offset >= num_bits_) {
+            set_vector_count(0);
+            set_filter_count(0);
+            return;
+        }
+
+        const auto count_bits = std::min(bit_count, num_bits_ - bit_offset);
+        const auto end_bit = bit_offset + count_bits;
+        size_t bit_pos = bit_offset;
+        size_t count = 0;
```
alexanderguzhva: please be consistent with names. If we're counting the future `filtered_count_` value, then this variable should be named as `filtered_count`.

include/knowhere/bitsetview.h
```diff
@@ -12,186 +12,246 @@
...
+        return num_bits_;
+    }
+
     const uint8_t*
     data() const {
         return bits_;
     }
 
+    void
+    count_filtered_bits(size_t bit_offset, size_t bit_count, const uint8_t* valid_bitmap = nullptr) {
```
alexanderguzhva, line 99: please add a comment on what this function does.

include/knowhere/bitsetview.h
```diff
@@ -12,186 +12,246 @@
...
     out_ids_data() const {
-        if (out_ids_ == nullptr) {
-            return nullptr;
-        }
         return out_ids_;
     }
 
     void
-    set_id_offset(size_t id_offset) {
+    set_id_offset(int64_t id_offset) {
```
alexanderguzhva: same comment about `int64_t`

include/knowhere/bitsetview.h
```diff
@@ -211,19 +271,26 @@ class BitsetView {
...
+    load_u64_unaligned_(const uint8_t* data) {
+        uint64_t value = 0;
+        std::memcpy(&value, data, sizeof(value));
+        return value;
+    }
+
     const uint8_t* bits_ = nullptr;
     size_t num_bits_ = 0;
-    size_t num_filtered_out_bits_ = 0;
+    size_t vector_count_ = 0;
```
alexanderguzhva, line 472: it would be nice to have comments on what `vector_count_` and `filtered_count_` are.

include/knowhere/id_map.h
```diff
@@ -0,0 +1,972 @@
...
+            }
+            return static_cast<int64_t>(out_to_in_ids_.Get(out_id));
+        });
+    }
+
+    const int64_t*
+    MapOutToIn(const int64_t* ids, size_t count, std::vector<int64_t>& in_ids) const {
+        return Read([&]() -> const int64_t* {
+            if (out_count_ == 0) {
+                return ids;
```
alexanderguzhva: I don't understand lines 686-688

include/knowhere/id_map.h
```diff
@@ -0,0 +1,972 @@
...
+            if (out_count < 0 || out_count > std::numeric_limits<int32_t>::max()) {
+                throw std::runtime_error("invalid external id map size");
+            }
+            out_count_ = static_cast<size_t>(out_count);
+            valid_bitmap_.Assign(valid_bitmap, out_count_);
+        });
+    }
+
+    void
+    SetValidBitmap(const bool* valid_data, int64_t out_count) {
```
alexanderguzhva: why `int64_t` out_count?

include/knowhere/id_map.h
```diff
@@ -0,0 +1,972 @@
...
+                out_to_in_ids_.AssignVector(out_count_, kInvalidId);
+            }
+            for (size_t in_id = 0; in_id < in_count; ++in_id) {
+                out_to_in_ids_.Set(in_to_out_ids_[in_id], static_cast<int32_t>(in_id));
+            }
+        });
+    }
+
+    void
+    BuildInToOutEblIds(const size_t* ebl_offsets, int64_t ebl_count) {
```
alexanderguzhva: same: why `int64_t ebl_count`?

include/knowhere/id_map.h
```diff
@@ -0,0 +1,972 @@
...
+        });
+    }
+
+    void
+    ClearEblIds() {
+        Write([&] { in_to_out_ebl_ids_.Clear(); });
+    }
+
+    void
+    AddIdsAndBitmap(const int32_t* out_ids, int64_t in_count, int64_t out_count) {
```
alexanderguzhva: same

include/knowhere/id_map.h
```diff
@@ -0,0 +1,972 @@
...
+    IdArraySnapshot in_to_out_ids_;
+    IdArraySnapshot in_to_out_ebl_ids_;
+    BitmapSnapshot valid_bitmap_;
+    size_t out_count_ = 0;
+};
+
+// Runtime-only logical-id mapping. Knowhere never serializes this structure into
+// index BinarySet payloads; callers persist nullable/valid-row state externally,
+// then configure the map exposed by Index/IndexNode before build or load.
+struct IdMap {
```
alexanderguzhva: Why is the locking mechanism built inside this structure?
Basically, why
```C++
struct IdMap {
    ...
    mutable std::shared_mutex mutex_;
};
```
instead of an externally managed access, such as
```
C++
struct IdMap {...};

IdMap id_map;
std::shared_mutex mutex;
```

One ore question for the whole structure: why does it rely on signed indices instead of purely unsigned ones everywhere? I understand that sometimes it is needed to map `size_t index` to `int64_t`, so I have an impression that certain `int64_t` need to be replaced with `size_t`-s

include/knowhere/comp/brute_force.h
```diff
@@ -53,7 +53,7 @@ class BruteForce {
 
     static Status
     SearchSparseWithBuf(const DataSetPtr base_dataset, const DataSetPtr query_dataset, sparse::label_t* ids, float* dis,
-                        const Json& config, const BitsetView& bitset, milvus::OpContext* op_context = nullptr) noexcept;
+                        const Json& config, BitsetView bitset, milvus::OpContext* op_context = nullptr) noexcept;
```
alexanderguzhva: at least, please use `const BitsetView bitset`

include/knowhere/index/index_node.h
```diff
@@ -318,6 +371,327 @@ class IndexNode : public Object {
...
+        return GetIdMap().MapOutToIn(out_id);
+    }
+
+    virtual const int64_t*
+    MapOutToIn(const int64_t* out_ids, size_t count, std::vector<int64_t>& in_ids) const {
+        return GetIdMap().MapOutToIn(out_ids, count, in_ids);
+    }
+
+    virtual Status
+    FinalizeIdMap() {
```
alexanderguzhva, line 413: please add a comment for this function about what it does and how to use it properly

src/index/data_view_dense_index/data_view_dense_index.h
```diff
@@ -348,13 +349,14 @@ DataViewIndexFlat::Search(const idx_t n, const void* __restrict x, const idx_t k
...
     const auto& search_pool = ThreadPool::GetGlobalSearchThreadPool();
+    const auto ntotal = Count();
     std::vector<folly::Future<folly::Unit>> futs;
     futs.reserve(n);
     if (k < faiss::distance_compute_min_k_reservoir) {
         if (metric_type_ == metric::L2) {
             faiss::cppcontrib::knowhere::HeapBlockResultHandler<CMAX> res(n, distances, labels, k);
             for (auto i = 0; i < n; i++) {
-                futs.emplace_back(search_pool->push([&] {
+                futs.emplace_back(search_pool->push([&, i = i] {
```
alexanderguzhva, line 359: so, you're saying that this `DataViewIndexFlat::Search()` code has been working incorrectly, because of the missing `i=i` clause in the `futs.emplace_back(search_pool->push([&, i = i] {` line.

This is a definite bug that this PR fixes. Same for the similar lines.

src/index/emb_list/emb_list_strategy_lemur.cc
```diff
@@ -420,11 +426,11 @@ class LemurEmbListStrategy : public EmbListStrategy {
             size_t nq = q_vec_end - q_vec_start;
             total_query_vecs += nq;
 
-            // Collect candidate doc IDs (no dedup needed — ANN returns doc-level IDs)
+            // Convert stage1 list/doc out ids to list/doc in ids for rerank.
             candidate_docs.clear();
             for (int32_t i = 0; i < ann_k; ++i) {
-                int64_t doc_id = ann_ids[q * ann_k + i];
-                if (doc_id >= 0 && doc_id < num_docs_) {
+                int64_t doc_id = to_in_list_id(ann_ids[q * ann_k + i]);
```
alexanderguzhva: `doc_id < num_docs_` condition is gone

src/index/emb_list/emb_list_strategy_muvera.cc
```diff
@@ -262,11 +268,11 @@ class MuveraEmbListStrategy : public EmbListStrategy {
             size_t nq = q_vec_end - q_vec_start;
             total_query_vecs += nq;
 
-            // Collect candidate doc IDs (no dedup needed — ANN returns doc-level IDs)
+            // Convert stage1 list/doc out ids to list/doc in ids for rerank.
             candidate_docs.clear();
             for (int32_t i = 0; i < ann_k; ++i) {
-                int64_t doc_id = ann_ids[q * ann_k + i];
-                if (doc_id >= 0 && doc_id < num_docs_) {
+                int64_t doc_id = to_in_list_id(ann_ids[q * ann_k + i]);
```
alexanderguzhva: same

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -345,36 +350,68 @@ class BaseFaissRegularIndexNode : public BaseFaissIndexNode {
...
-        return internal_offset_to_label;
     }
 
-    Status
-    SetInternalIdToMostExternalIdMap(std::vector<uint32_t>&& map) override {
-        internal_offset_to_most_external_id = std::move(map);
-        return Status::success;
+    void
+    PrepareBitsetForSubIndex(BitsetView& bitset, size_t index_id) const {
+        const auto& out_ids = bitset_out_ids_[index_id];
```
alexanderguzhva, line 414: does this code preserve mv-related logic?

alexanderguzhva:
@marcelo-cjl is it good to re-review the PR?

include/knowhere/bitsetview.h
marcelo-cjl: I kept -1 as the "not calculated yet" sentinel. 0 already means the count is known and no participating vector is filtered, so using size_t{0} as the default would change BitsetView::empty() semantics.

include/knowhere/bitsetview.h
marcelo-cjl: Done.

include/knowhere/bitsetview.h
marcelo-cjl: Done.

include/knowhere/bitsetview.h
marcelo-cjl: Done.

include/knowhere/bitsetview.h
marcelo-cjl: Done.

include/knowhere/bitsetview.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: I kept the lock inside IdMap because the map is not a single array. Snapshot/search reads and build/load/append updates need a consistent view of in_to_out_ids_, in_to_out_ebl_ids_, out_to_in_ids_, valid_bitmap_, and out_count_. Keeping the optional lock with those fields makes that consistency boundary explicit.

I also kept signed ids only where the API needs the -1 invalid-id sentinel. Pure counts and offsets are changed to size_t where applicable.

include/knowhere/comp/brute_force.h
marcelo-cjl: Done.

include/knowhere/index/index_node.h
marcelo-cjl: Done.

src/index/data_view_dense_index/data_view_dense_index.h
marcelo-cjl: Done.

src/index/emb_list/emb_list_strategy_lemur.cc
marcelo-cjl: Done.

src/index/emb_list/emb_list_strategy_muvera.cc
marcelo-cjl: Done.

src/index/hnsw/faiss_hnsw.cc
marcelo-cjl: The MV path is still covered. The old helper exposed a global internal-id map, but the multi-index selector needs a per-sub-index local-id map. BuildBitsetOutIds() builds exactly that:

local FAISS id -> labels[index_id][local_id] -> public bitset id

For nullable/EmbList cases, the last hop uses the finalized in_to_out_ids or in_to_out_ebl_ids. Search results still map local ids back through labels, then through the normal result-id map.

include/knowhere/id_map.h
```diff
@@ -0,0 +1,984 @@
...
+            id_map_ = {};
+            use_map_ = false;
+        }
+
+        std::vector<int32_t> ids_;
+        std::unordered_map<int32_t, int32_t> id_map_;
+        bool use_map_ = false;
+    };
+
+    struct IdArrayStorage {
```
alexanderguzhva: I have a strange question.
Why is it needed to have `IdStorageArray` and `IdArrayBuffer` instead of plain `std::vector<std::vector<>>` and `std::vector<>` ?

include/knowhere/id_map.h
```diff
@@ -0,0 +1,984 @@
...
+                std::copy(ids_->data.get(), ids_->data.get() + size_, ids->data.get());
+            }
+            ids_ = std::move(ids);
+        }
+
+        std::shared_ptr<IdArrayBuffer> ids_;
+        size_t size_ = 0;
+    };
+
+    struct BitmapStorage {
```
alexanderguzhva: same question
can it be replaced with something standard and simpler?

include/knowhere/id_map.h
marcelo-cjl: The array snapshots expose raw pointers to prepared bitsets, result mapping, and iterators after the IdMap read lock is released. A vector owned directly by IdMap can reallocate on append/rebuild and leave those pointers dangling. The buffer object gives each snapshot shared ownership of its backing storage generation, while IdMap can move to a new generation on growth. I added comments to make that contract explicit.

include/knowhere/id_map.h
marcelo-cjl: Same reason here: BitmapSnapshot exposes raw bytes to BitsetView after the IdMap read lock is released, so a plain vector owned directly by IdMap would not preserve old snapshot pointers across resize/rebuild. BitmapStorage also needs the partial-tail-byte copy rule when a shared snapshot ends mid-byte; otherwise extending the bitmap can mutate bytes visible through the old snapshot. I added comments for this.

tests/ut/test_nullable_id_map.cc
```diff
@@ -0,0 +1,4266 @@
...
+        {"GPU_CAGRA", knowhere::IndexEnum::INDEX_GPU_CAGRA, DataKind::GpuFp32},
+        {"GPU_FAISS_FLAT", knowhere::IndexEnum::INDEX_FAISS_GPU_IDMAP, DataKind::GpuFp32},
+        {"GPU_FAISS_IVF_FLAT", knowhere::IndexEnum::INDEX_FAISS_GPU_IVFFLAT, DataKind::GpuFp32},
+        {"GPU_FAISS_IVF_PQ", knowhere::IndexEnum::INDEX_FAISS_GPU_IVFPQ, DataKind::GpuFp32},
+        {"GPU_FAISS_IVF_SQ8", knowhere::IndexEnum::INDEX_FAISS_GPU_IVFSQ8, DataKind::GpuFp32},
+        {"SVS_FLAT", knowhere::IndexEnum::INDEX_SVS_FLAT, DataKind::DenseFp32},
+        {"SVS_VAMANA", knowhere::IndexEnum::INDEX_SVS_VAMANA, DataKind::DenseFp32},
+        {"SVS_VAMANA_LVQ", knowhere::IndexEnum::INDEX_SVS_VAMANA_LVQ, DataKind::DenseFp32},
+        {"SVS_VAMANA_LEANVEC", knowhere::IndexEnum::INDEX_SVS_VAMANA_LEANVEC, DataKind::DenseFp32},
+        {"HNSW (Cardinal v2)", knowhere::IndexEnum::INDEX_HNSW, DataKind::DenseFp32},
```
foxspy: remove this

foxspy:
@marcelo-cjl We need to add documentation explaining the current internal/external ID model and how bitsets and ID mappings are handled when combined with embedding lists.

foxspy:
Regarding `BitsetView` (`include/knowhere/bitsetview.h:35-58`): could we model the tri-state as `std::optional<size_t>` instead of changing the public count type to `int64_t` and using `-1` as a sentinel?

We do need to distinguish **unknown** from **known zero**, so that `empty()` cannot incorrectly skip a real filter. However, the current `count()` only guards the sentinel with a debug `assert`; in an `NDEBUG` build, `-1` can escape as a valid return value. An optional keeps count non-negative, makes the unknown state explicit to callers, and avoids a signed sentinel in the public API. A `has_known_count()` or another explicit checked accessor would also make the lazy-count contract clearer.

tests/ut/test_nullable_id_map.cc
marcelo-cjl: Done.

marcelo-cjl:
@foxspy Re documentation: Done. Added `docs/design_docs/20260731-nullable-id-mapping.md` and tightened the source comments around the id-map, bitset, and embedding-list boundaries.

Re `BitsetView` tri-state count: Done. The count state is now `std::optional<size_t>`, `has_known_count()` exposes the lazy-count contract, and `count()` throws if the count is still unknown instead of returning a signed sentinel.

marcelo-cjl:
@foxspy @alexanderguzhva I've added design documents and important comments, and implemented the mmap feature for idmap to support tier and lake scenarios. Please take a look at them when you have time.

alexanderguzhva:
/lgtm
this is a big change. We will fix it along the way if something goes wrong

foxspy:
[Blocking] Please provide a safe publication boundary for the ID mapping in `src/index/index.cc:236`.

Under the current single-Add-writer contract, Search can still call `GetSnapshot()` concurrently with this Add. `IdMap` defaults to `use_lock=false`, so the writer mutates the member `shared_ptr`s, sizes, bitmap, and `out_count_` while a reader copies the same objects. The current copy-on-grow / `use_count()` logic only keeps an already-acquired snapshot alive; without synchronization, snapshot acquisition itself is a C++ data race and may observe a mixed generation.

Please either publish one immutable `IdMapGeneration` through an atomic pointer, or use the same short synchronization boundary for snapshot capture and writer COW/publication. Search can remain lock-free after it has acquired the snapshot.

Also, rolling back only the IdMap when the backend Add returns an error is not sufficient: HNSW/sparse Add may have already inserted a prefix of the batch. The Add contract must either guarantee all-or-nothing backend mutation, or mark the index/node unusable so the caller discards it and falls back/rebuilds; it must not continue using a potentially partial backend with a rolled-back mapping.

include/knowhere/id_map.h
marcelo-cjl: Done. The internal lock is gone. IdMap now has explicit sealed/growing storage: sealed maps publish complete arrays, and growing maps append ranges while preserving existing backend-id labels. Callers set the storage type before nullable data is consumed.

include/knowhere/id_map.h
marcelo-cjl: Done. The old IdStorageArray / IdArrayBuffer snapshot model has been replaced. Forward maps now use ArrayStore: sealed maps use complete contiguous arrays with optional mmap, while growing maps use appendable storage so existing ids remain stable. A plain std::vector only covers the heap sealed case and would push mmap/growing policy into callers.

include/knowhere/id_map.h
marcelo-cjl: Done. Bitmap storage was simplified with the same lifecycle boundary. It stores packed bytes plus the logical public-id bit count; sealed maps publish complete bitmap state, while growing maps append public-id ranges. It is not a separate snapshot abstraction now.

marcelo-cjl:
@foxspy Done. I removed the old snapshot/COW and rollback-only path. Growing id maps now use appendable storage for forward arrays, so existing backend ids stay stable while new public-id ranges are appended. `PrepareBitset()` captures the mapped-id count used by a request, and caller visibility is still bounded by the published public-id prefix. `Index::Add()` now propagates backend Add failures instead of rolling back only the id map, so a partial backend mutation is not treated as a recovered synchronized index.

foxspy:
[Blocking] Please use data-first / commit-last publication and avoid introducing oneTBB for this ID-map change.

Under the established single-Add-writer contract, no concurrent container is required here. Please use ordinary append-only chunk storage: write the bitmap, forward/reverse mappings, and count metadata into the uncommitted tail first, then publish one committed offset/record with a release store. Search should acquire-load that commit point once and only read the committed prefix. This provides the required happens-before relationship without `tbb::concurrent_vector` or another third-party dependency.

The writer must not modify memory that is already inside the committed prefix (including a packed bitmap tail byte), and the chunk metadata read by Search must be completely initialized before the commit is published.

If backend Add returns an error, propagate it directly and abort the current index operation. This is an unexpected failure path; please do not attempt a partial IdMap rollback or continue using the result of that Add.

Please validate the replacement with a deterministic Add/Search interleaving test and all existing x86-64/ARM64 build and package modes.

marcelo-cjl:
@foxspy Done. Removed oneTBB/TBB from the growing id-map path. Growing storage now uses append-only chunks with data-first publication: ids/bitmap bytes are written first, then the visible prefix is advanced, so readers only observe committed prefixes. `BitsetView` keeps a shared id-array view plus an explicit count without copying id contents. Added prefix-stability coverage for id arrays, bitmaps, and EmbList append paths.

alexanderguzhva:
waiting for @foxspy

include/knowhere/array_store.h
```diff
@@ -0,0 +1,565 @@
...
+
+    void
+    Append(const T* data, size_t count) {
+        if (count != 0 && data == nullptr) {
+            throw std::runtime_error("array store data is null");
+        }
+        if (count == 0) {
+            return;
+        }
+        type_ = Type::APPEND_ARRAY;
```
foxspy: [Blocking][F3] `ArrayStore::Append()` must not rewrite wrapper metadata while old-prefix readers may concurrently read it. Even assigning the same values to `type_` / `visible_count_`, or checking-and-initializing `append_array_`, races with `size()`, `operator[]`, `StorageSize()`, and snapshot/copy readers; the later `PublishCounts()` release cannot synchronize these pre-publication ordinary-field writes with readers that still observe the old top-level offset. Please initialize/freeze `type_`, `append_array_`, and dynamic visibility once in `SetType(APPEND_ARRAY)` before concurrent access. `Append()` should only validate the frozen state and call `append_array_->Append(data, count)`.

foxspy (commented):
F3 wrapper metadata data race：两处行级建议。append-only + PublishCounts 的逻辑 prefix 模型可以保留；这里只要求 growing wrapper 在并发前初始化并冻结。

include/knowhere/array_store.h
```diff
@@ -0,0 +1,565 @@
...
+    void
+    Append(const uint8_t* data, size_t bit_count) {
+        // Append is bit-oriented; batches may end in the middle of a byte.
+        if (bit_count == 0) {
+            return;
+        }
+        if (data == nullptr) {
+            throw std::runtime_error("bitmap array data is null");
+        }
+        type_ = Type::APPEND_ARRAY;
```
foxspy: [Blocking][F3] Same issue for `BitmapArray::Append()`: do not rewrite `type_` / `bit_count_` or lazily initialize `append_data_` during concurrent append. Old-prefix readers legitimately access these ordinary wrapper fields before the new `out_count_` commit marker is published, so `PublishCounts()` does not order those accesses. Initialize/freeze the append wrapper in `SetType(APPEND_ARRAY)`; then make `Append()` only validate the state and call `append_data_->Append(data, bit_count)`.

include/knowhere/array_store.h
marcelo-cjl: Done.

include/knowhere/array_store.h
marcelo-cjl: Done.

include/knowhere/id_map.h
```diff
@@ -0,0 +1,506 @@
...
+                }
+
+                const auto old_out_count = type_ == Type::GROWING ? OutCount() : 0;
+                const auto old_valid_count = type_ == Type::GROWING ? InCount() : 0;
+                const auto in_id_begin = type_ == Type::GROWING ? old_valid_count : 0;
+
+                std::vector<int32_t> ids(data.in_count);
+                std::vector<uint8_t> bitmap(BitmapByteSize(data.out_count), 0);
+                for (size_t i = 0; i < data.in_count; ++i) {
+                    const auto local_out_id = static_cast<size_t>(data.out_ids[i]);
```
foxspy, line 173: **Blocking (F6): validate the complete `IDS` batch before mutating any map/storage.** Here a negative public ID is cast to `size_t`, and an ID `>= data.out_count` also reaches `SetPackedBit()` without bounds checking, so malformed input can write outside `bitmap` before a controlled error is returned. Duplicate/range checking in `AdaptiveStore` is too late because the bitmap and forward map have already been appended. Please reject null input (when `in_count > 0`), negative/out-of-range/duplicate IDs, and addition/`int32_t` overflow before any append. The same validation boundary should reject mapped selected IDs equal to `-1` before calling Flat/IVF/DiskANN/AiSAQ/Sparse backends.

include/knowhere/id_map.h
```diff
@@ -0,0 +1,506 @@
...
+        const auto list_count = static_cast<size_t>(ebl_count);
+        if (list_count == 0) {
+            return;
+        }
+
+        const auto append_count = ebl_offsets[list_count];
+
+        std::vector<int32_t> ids(append_count, kInvalidId);
+        for (size_t ebl_id = 0; ebl_id < list_count; ++ebl_id) {
+            const auto out_id = static_cast<int32_t>(ebl_id_begin + static_cast<int64_t>(ebl_id));
```
foxspy: **Blocking (F8): this assigns the appended embedding-list's compact ordinal, not its public list ID.** For example, if the existing compact-to-public list map is `[1, 4]` and the new list's public ID is `7`, this code appends `2`; the full-build path correctly expands `in_to_out_ids_[ebl_id]`. The derived vector-level map is later consumed as public IDs for filtering/results, so append and rebuild disagree under non-identity mappings. Please read the public ID from the list-domain map at `ebl_id_begin + ebl_id` (with explicit length/domain validation), and use the ordinal only when no mapping exists and identity semantics are intentional.

src/index/index.cc
```diff
@@ -124,6 +202,11 @@ Index<T>::Add(const DataSetPtr dataset, const Json& json, bool use_knowhere_buil
         auto cfg = this->node->CreateConfig();
         std::string msg;
         RETURN_IF_ERROR(LoadConfig(cfg.get(), json, knowhere::TRAIN, "Add", &msg));
+        auto& id_map = this->node->GetIdMap();
+        AddDatasetIdMapData(dataset, id_map);
```
foxspy: **Blocking (F12): the map mode can silently diverge from backend cardinality across Add calls.** `AddDatasetIdMapData()` returns `false` for a plain dataset, but this result is ignored and backend Add still proceeds. After a mapped Add, a plain Add therefore extends backend storage without extending/resetting the ID map; new internal IDs map to `-1` and selected-ID lookup keeps the old public domain. The reverse plain-to-mapped transition is also misaligned because the first map starts from internal/public offset zero while backend rows already exist. Please freeze the index as plain or mapped after its first Build/Add and reject a mode change, or explicitly append/backfill identity entries using the current backend cardinality; do not silently mix modes.

include/knowhere/id_map.h
marcelo-cjl: Done.

include/knowhere/id_map.h
marcelo-cjl: Done.

src/index/index.cc
marcelo-cjl: Done.

src/common/comp/brute_force.cc
```diff
@@ -141,6 +141,101 @@ GetInverseVecNorms(const DataSetPtr& base) {
...
+    if (!bitset.has_out_ids()) {
+        return Status::success;
+    }
+    const auto total_out_ids = bitset.out_ids_count();
+    if (internal_id_offset > total_out_ids || num_internal_ids > total_out_ids - internal_id_offset) {
+        LOG_KNOWHERE_WARNING_ << "brute-force out id map is smaller than the internal id range";
+        return Status::invalid_args;
+    }
+    // Filtering and result mapping use the same local physical-id window.
+    bitset.set_id_offset(internal_id_offset);
```
liliu-z, line 170: `UseOutIdsWindow()` overwrites the BitsetView's id offset with the chunk-local base rather than composing it with whatever the caller already set. The flat path treats a caller-supplied offset as a real window into `out_ids` — `PrepareBitsetForInternalRange()` returns early when `has_out_ids()`, and `MapOneResultId()` indexes `out_ids[bitset.id_offset() + internal_id]`, which the new test at tests/ut/test_bruteforce.cc:388-408 asserts (out_ids `{0,1,5}` with `set_id_offset(1)` yields `{5, -1}`). Run the same bitset against a chunked base dataset and chunk 0 resets the offset to 0, so filtering and result mapping both read the wrong slice of `out_ids`. Add the chunk base to the existing offset at all three call sites (lines 792, 914, 1596), and extend the bounds check just above so it validates the composed offset against `total_out_ids` instead of only the chunk-local part.

src/index/minhash/minhash_index_node.cc
```diff
@@ -332,6 +332,7 @@ MinHashLSHNode<DataType>::Search(const DataSetPtr dataset, std::unique_ptr<Confi
         return expected<DataSetPtr>::Err(Status::internal_error, e.what());
     }
     auto res = GenResultDataSet(nq, topk, std::move(p_id), std::move(p_dist));
+    MapSearchResultIdsToOutIds(res);
```
liliu-z, line 335: This PR makes MinHash map backend result ids to public ids on the search path, but leaves the retrieval path in the backend domain: `GetVectorByIds` hands the caller's public ids straight to `minhash_lsh_->GetDataByIds()`. Every other node that gained result mapping also gained the reverse conversion for selected ids (flat.cc:245, faiss_hnsw.cc:1181, ivf.cc:1592, diskann.cc:1016, diskann_aisaq.cc:755, sparse_index_node.cc:992 all call `CompactOutToIn`), so a nullable MinHash index returns correct search labels but wrong vectors for the same ids. Route the ids through `CompactOutToIn(ids, rows, in_ids, Count())` before `GetDataByIds`.

Issue #1674, opened 2026-06-10
Facade APIs should not propagate C++ exceptions


foxspy:
## Description

Knowhere public facade APIs are designed to return `Status` or `expected<T>` values, but some call paths can still let C++ exceptions escape from lower layers. This makes error handling inconsistent for callers that expect error-code based results.

## Expected Behavior

Facade APIs should not propagate C++ exceptions to callers. Normal return values and existing error codes should pass through unchanged, and unexpected C++ exceptions should be converted into an internal error result.

## Actual Behavior

Some facade API paths do not consistently guard lower-layer calls, so a thrown C++ exception can escape instead of being reported through the existing error-code based API contract.

PR #1675, opened 2026-06-10
fix: guard facade APIs from exceptions

foxspy:

## Summary
- Add guarded facade calls so C++ exceptions are converted to error-code based results.
- Mark public facade APIs as `noexcept` and preserve existing returned status values.
- Add status-category helpers and unit coverage for error-code behavior.

### DIFF
--- a/src/index/index_static.cc
+++ b/src/index/index_static.cc
@@ -20,6 +20,13 @@ namespace knowhere {
 inline Status
 LoadStaticConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, const std::string& method,
                  std::string* const msg = nullptr) {
+    if (cfg == nullptr) {
+        if (msg != nullptr) {
+            *msg = "failed to create config";
+        }
+        LOG_KNOWHERE_ERROR_ << method << " failed to create config";
+        return Status::knowhere_inner_error;
+    }
     Json json_(json);
     auto res = Config::FormatAndCheck(*cfg, json_, msg);
     LOG_KNOWHERE_DEBUG_ << method << " config dump: " << json_.dump();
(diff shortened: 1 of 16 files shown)


src/index/index_static.cc
```diff
@@ -36,61 +36,69 @@ IndexStaticFaced<DataType>::Instance() {
...
-    auto cfg = IndexStaticFaced<DataType>::CreateConfig(indexType, version);
-
-    const Status status = LoadStaticConfig(cfg.get(), params, knowhere::PARAM_TYPE::TRAIN, "ConfigCheck", &msg);
-    if (status != Status::success) {
-        LOG_KNOWHERE_ERROR_ << "Load Config failed, msg = " << msg;
-        return status;
-    }
+                                        std::string& msg) noexcept {
+    return GuardedCall([&]() -> Status {
+        auto cfg = IndexStaticFaced<DataType>::CreateConfig(indexType, version);
```
liliu-z, line 61: CreateConfig now wraps its body in GuardedCall returning std::unique_ptr<BaseConfig>, and on a caught exception GuardedFailure takes the default-constructible branch (expected.h:318-319 `return Result{};`), yielding a null unique_ptr. All three callers then use that pointer without a null check: ConfigCheck (line 54/56), EstimateLoadResource (line 85/88) and HasRawData (line 125/127) pass cfg.get() to LoadStaticConfig, which dereferences it via `Config::FormatAndCheck(*cfg, json_, msg)` (line 24), so a swallowed exception becomes a null-pointer segfault that the callers' own GuardedCall cannot catch, defeating the noexcept guarantee. Note the only trigger is an exception during config construction (e.g. bad_alloc from make_unique, or a throwing StaticCreateConfig factory) — an unknown index type does NOT produce null because line 45 returns a valid make_unique<BaseConfig>(), and invalid params are reported by LoadStaticConfig after CreateConfig. Fix by null-checking cfg before use, or have CreateConfig surface a Status/expected instead of a bare pointer.

foxspy: Fixed in 9613b916. `LoadStaticConfig` now checks for a null config before dereferencing and returns `faiss_inner_error` with a message. Added an `[error_code]` test covering `ConfigCheck`, `EstimateLoadResource`, and `HasRawData` when config creation throws.

foxspy: Follow-up: corrected this in 5d2a2fbf. The facade-level GuardedCall fallback now returns `knowhere_inner_error` instead of `faiss_inner_error`, since the facade cannot assume the exception came from Faiss. Explicit status values returned by lower layers, such as `faiss_inner_error`, `invalid_args`, or `empty_index`, are still propagated unchanged.

liliu-z (commented):
/lgtm
/approve

Issue #1678, opened 2026-06-13
Upgrade vendored Faiss to e420e94


foxspy:
## Description

The vendored Faiss code should be updated to upstream commit e420e94 to pick up recent upstream fixes and SIMD implementation changes.

## Expected Behavior

Knowhere builds with the updated vendored Faiss snapshot and existing unit tests continue passing.

## Actual Behavior

Knowhere still vendors an older Faiss snapshot.

PR #1679, opened 2026-06-13
build: upgrade faiss to e420e94

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/1678

## Summary
- Upgrade vendored Faiss to upstream commit e420e94.
- Update the local Faiss CMake integration for new SIMD source files.
- Adapt the Knowhere HNSW wrapper to the updated Faiss visited-table API.

## Verification
- make BUILD_DIR=$(pwd)/build/faiss-upgrade WITH_UT=True
- source build/faiss-upgrade/Release/generators/conanrun.sh && build/faiss-upgrade/Release/tests/ut/knowhere_tests


thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -2572,9 +2630,10 @@ std::unique_ptr<Index> read_index_up(IOReader* f, int io_flags) {
         READ1(svs->construction_window_size);
         READ1(svs->max_candidate_pool_size);
         READ1(svs->prune_to);
-        READ1(svs->use_full_search_history);
+        READ1_BOOL(svs->use_full_search_history);
 
         svs->storage_kind = read_svs_storage_kind(f);
+        READ1_BOOL(svs->is_static);
```
liliu-z, line 2636: This PR adds a new `is_static` byte to the SVS Vamana serialization stream (written at index_write.cpp:1047, read at index_read.cpp:2636) without bumping the fourcc tags (ILVQ/ISVL/ISVD/ISV2). SVS Vamana has been serializable in knowhere since #1430, and the parent commit's reader wrote no Vamana is_static byte, so any index persisted by a prior WITH_SVS build lacks this byte; the new reader consumes the following `initialized` flag as is_static and misparses the rest of the stream. Because WITH_SVS is OFF by default the blast radius is small, but this silently changes the on-disk format and should be called out as an upgrade/migration concern (note that locally bumping the fourcc would diverge from upstream faiss and break the next sync). The reviewer's cited read location index_read.cpp:351 is wrong — that line is ITQTransform code; the actual read is at line 2636.

thirdparty/faiss/faiss/utils/simd_levels.cpp
```diff
@@ -155,16 +158,23 @@ SIMDLevel SIMDConfig::auto_detect_simd_level() {
...
+                bool has_avx512_fp16 = (cpuid7_edx & (1 << 23)) != 0;
                 unsigned int eax1, ebx1, ecx1, edx1;
                 eax1 = 7;
                 ecx1 = 1;
                 asm volatile("cpuid"
                              : "=a"(eax1), "=b"(ebx1), "=c"(ecx1), "=d"(edx1)
                              : "a"(eax1), "c"(ecx1));
                 bool has_avx512_bf16 = (eax1 & (1 << 5)) != 0;
-                if (has_avx512_bf16) {
+                if (has_avx512_bf16 && has_avx512_fp16) {
```
liliu-z, line 177: The SPR runtime gate dispatches to AVX512_SPR on `has_avx512_bf16 && has_avx512_fp16`, but the SPR kernels also issue VNNI (`_mm512_dpbusd_epi32`) and VPOPCNTDQ (`_mm512_popcnt_epi64`), whose CPUID bits are never verified. On real Intel silicon this is harmless because AVX512-FP16 (Sapphire Rapids and later) implies both VNNI and VPOPCNTDQ, which predate it — so the only way to hit it is a feature-masked VM that exposes FP16/BF16 while hiding VNNI or VPOPCNTDQ, which would take a SIGILL on the first such instruction. The fix is cheap: add AVX512-VNNI and AVX512-VPOPCNTDQ checks so the gate matches the instructions actually emitted, consistent with the FP16 check the code already added for the Zen4 case.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HNSW.cpp
```diff
@@ -742,7 +742,7 @@ int search_from_candidates(
             if (v1 < 0)
                 break;
 
-            prefetch_L2(vt.visited.data() + v1);
+            vt.prefetch(v1);
```
liliu-z, line 747: This PR makes faiss::VisitedTable an abstract base with virtual set/get/prefetch (VisitedTable.h), and the contrib HNSW search hot loops now call it through that abstract reference on every visited node — vt.prefetch(v1) at line 745 and vt.get/vt.set at 770-773 in search_from_candidates, plus 886 and 910-911 in search_from_candidate_unbounded. Upstream impl/HNSW.cpp avoids this by dynamic_cast-ing the table to its concrete type (VisitedTableVector/VisitedTableSet) once outside the loop and calling a templated non-virtual variant (impl/HNSW.cpp:569, 1056, 1483, 1578); the contrib copy was not given the same treatment, so the abstraction change regresses Knowhere's HNSW search path with a virtual dispatch and lost inlining per node visit. The reviewer's cited line 566 is wrong — that is the `VisitedTable& vt` parameter of the build-path add_links_starting_from, not a search hot loop.

thirdparty/faiss/faiss/utils/simd_levels.cpp
foxspy: This is an upstream Faiss runtime dispatch concern in the selected Faiss revision. The SPR gate currently relies on the real-hardware implication from AVX512-FP16/BF16 to VNNI/VPOPCNTDQ; a feature-masked VM could still expose a mismatch. I am leaving this unchanged in Knowhere to avoid local divergence. If we want stricter CPUID gating, it should be proposed upstream.

thirdparty/faiss/faiss/impl/index_read.cpp
foxspy: Keeping the serialized format aligned with upstream Faiss in this PR rather than bumping the fourcc locally. The SVS path has not been formally released in Knowhere, so the compatibility exposure is controlled; this should be called out as an upgrade/migration concern instead of introducing a local format divergence.

thirdparty/faiss/faiss/cppcontrib/knowhere/impl/HNSW.cpp
foxspy: Updated the Knowhere contrib HNSW path to mirror upstream Faiss: dispatch once to the concrete VisitedTableVector/VisitedTableSet type, then run templated helpers in the search loops. The hot-loop get/set/prefetch calls are now on the concrete type instead of virtual calls through VisitedTable.

liliu-z (commented):
/lgtm
/approve

PR #1683, opened 2026-06-17
fix: handle empty SINDI posting lists

sparknack:
issue: #1555

- Fix SINDI POSTING_LISTS section size for empty sparse data by counting metadata and dim offsets before returning.
- Run sparse unit tests against both current and maximum knowhere index versions.


foxspy (approved):
/lgtm

PR #1684, opened 2026-06-17
Use OpContext trace span for Cardinal tracing

jamesgao-jpg:
## Summary

- keep Cardinal v2 pinned to the existing released version `v3.0.2`; this PR does not advance Knowhere main to the tracing Cardinal branch
- prefer the active `milvus::OpContext` trace span in Knowhere search paths, while keeping the legacy `trace_id`/`span_id`/`trace_flags` config fields as a compatibility fallback for callers that do not yet seed `OpContext::trace_span`
- use the shared milvus-common `NestedSpanGuard` helper to temporarily expose the Knowhere search span to downstream Cardinal calls
- include shared trace helpers from `common/Tracer.h` and construct spans through the existing tracer utilities instead of going through the removed `StartScopedSpan` wrapper

## Dependency

PR https://github.com/zilliztech/cardinal/pull/865 introduces the Cardinal tracing implementation that a future Cardinal version bump can expose through Knowhere. This PR only prepares Knowhere to propagate the OpContext trace span and keeps the current Cardinal v2 version pin unchanged.

This branch pins the published helper package `milvus-common/1.0.0-ae43d5c@milvus/dev#fe24230e26e129f6bf303a8342ba8980`, produced from milvus-common commit `ae43d5cf11852185ad5d5b48808cb7b58cc4ccd4`.

## Verification

- Not run locally. This branch is intended to be validated through upstream CI and the downstream Milvus integration branch.

Follow-up issue: https://github.com/zilliztech/knowhere/issues/1693 tracks the future Cardinal v2 version bump after Cardinal tracing is released.


src/index/index.cc
```diff
@@ -160,38 +160,28 @@ Index<T>::Search(const DataSetPtr dataset, const Json& json, const BitsetView& b
...
-            auto ctx = tracer::TraceContext{.traceID = reinterpret_cast<const uint8_t*>(trace_id_str.c_str()),
-                                            .spanID = reinterpret_cast<const uint8_t*>(span_id_str.c_str()),
-                                            .traceFlags = static_cast<uint8_t>(b_cfg.trace_flags.value())};
-            span = tracer::StartSpan("knowhere search", &ctx);
-            span->SetAttribute(meta::METRIC_TYPE, b_cfg.metric_type.value());
-            span->SetAttribute(meta::TOPK, b_cfg.k.value());
-            span->SetAttribute(meta::ROWS, Count());
-            span->SetAttribute(meta::DIM, Dim());
-            span->SetAttribute(meta::NQ, dataset->GetRows());
+        auto span = milvus::tracer::ScopedSpan("knowhere search", milvus::OpContext::GetTraceSpan(op_context));
```
liliu-z: `index.cc:165` derives the search span from `op_context->GetTraceSpan()`, but no released Milvus populates `op_context->trace_span` — a grep for `trace_span =` across `internal/core/src` finds nothing — so `GetTraceSpan()` returns `nullptr` and every Knowhere search/brute-force span (here and at `brute_force.cc:512`) is silently dropped. The `trace_id`/`span_id`/`trace_flags` JSON fields still advertised in `config.h` are now dead: a caller that sets them gets a silent no-op instead of a trace. Built against anything but the unmerged integration branch all Knowhere tracing disappears with no error, so gate this on the upstream change that writes `trace_span`, or remove the dead config fields until it lands.

src/index/index.cc
```diff
@@ -160,38 +160,28 @@ Index<T>::Search(const DataSetPtr dataset, const Json& json, const BitsetView& b
...
+        const auto& span_ptr = span.Get();
+        if (span_ptr != nullptr) {
+            span_ptr->SetAttribute(meta::METRIC_TYPE, b_cfg.metric_type.value());
+            span_ptr->SetAttribute(meta::TOPK, b_cfg.k.value());
+            span_ptr->SetAttribute(meta::ROWS, Count());
+            span_ptr->SetAttribute(meta::DIM, Dim());
+            span_ptr->SetAttribute(meta::NQ, dataset->GetRows());
         }
+        auto op_context_trace_span_guard =
```
liliu-z, line 175: The NestedSpanGuard at index.cc:174-175 writes span_ptr into op_context->trace_span before fanning out to parallel workers and restores it on scope exit. It is safe today because Milvus builds a fresh stack-local OpContext per search (segment_c.cpp:399) and OpContext deletes its copy/move constructors, so no two concurrent top-level searches share one. But the guard mutating caller-owned state is a latent contract violation: a future caller that shares one OpContext across concurrent Index::Search calls would hit an unsynchronized shared_ptr write/read race. Document the single-owner-per-search requirement, or carry the span on a per-call local instead of mutating op_context.

conanfile.py
```diff
@@ -116,7 +116,7 @@ def configure(self):
     def requirements(self):
         self.requires("abseil/20250127.0#481edcc75deb0efb16500f511f0f0a1c")
         self.requires("boost/1.83.0#4e8a94ac1b88312af95eded83cd81ca8")
-        self.requires("milvus-common/1.0.0-ac18852@milvus/dev#d9aa7497425662eb17749b4871700200")
+        self.requires("milvus-common/1.0.0-45bff01@milvus/dev#97948822f1dacdeafed7e8d5251df493")
```
liliu-z: The Cardinal pin at cmake/libs/cardinal/v2/CMakeLists.txt:6 changes from the immutable tag v3.0.2 to a mutable branch (origin/codex/cardinal-master-config-carrier-tracing) that is git-checked-out at build time, so builds track a moving branch tip and are not reproducible. The milvus-common reference at conanfile.py:119 pins commit 45bff01, which is not an ancestor of milvus-common master (a side commit, 'move scoped trace helpers into Tracer'); its conan recipe revision hash does keep that package itself reproducible, but it likewise points at unmerged upstream work. Gate the merge on these upstream changes landing and repin Cardinal to an immutable tag/commit on mainline.

src/index/index.cc
```diff
@@ -160,38 +160,28 @@ Index<T>::Search(const DataSetPtr dataset, const Json& json, const BitsetView& b
         }
 
 #if defined(NOT_COMPILE_FOR_SWIG) && !defined(KNOWHERE_WITH_LIGHT)
-        const BaseConfig& b_cfg = static_cast<const BaseConfig&>(*cfg);
+        BaseConfig& b_cfg = static_cast<BaseConfig&>(*cfg);
```
liliu-z: b_cfg is needlessly non-const at index.cc:163 and :257 — it was const BaseConfig& before this PR and every remaining use is a read (e.g. b_cfg.metric_type.value() in the SetAttribute calls), and BaseConfig's optional fields expose a const .value() overload. Revert both casts to const BaseConfig&.

src/index/index.cc
jamesgao-jpg: Thanks, this was valid. Fixed in 199c60d3247b349b864f27e872712e1e2b115207: Knowhere now starts spans through `StartMilvusSpanFromOpContextOrConfig`, which prefers `OpContext::trace_span` but falls back to the legacy `trace_id`/`span_id`/`trace_flags` config fields when the OpContext span is absent. This keeps tracing working for Milvus versions that still pass trace context through JSON config, and applies to both `index.cc` search/range search and the brute-force span sites.

src/index/index.cc
jamesgao-jpg: Fixed in 199c60d3247b349b864f27e872712e1e2b115207. The two `b_cfg` casts in `index.cc` are now `const BaseConfig&` again; the remaining uses only read config fields for span attributes and fallback trace context.

src/index/index.cc
jamesgao-jpg: This is the same ownership concern raised and addressed in milvus-common PR https://github.com/zilliztech/milvus-common/pull/96#discussion_r3490266214.

We fixed it in milvus-common commit `75fd9691a82f89caab715907e3ff4d2c7b5e25c4` and replied here: https://github.com/zilliztech/milvus-common/pull/96#discussion_r3490431284. The shared `OpContext::trace_span` API now documents the contract directly: it is a single-threaded parent handoff slot, not a concurrent current-span stack, and code entering parallel work must copy the parent into per-worker state before nesting spans.

Knowhere follows that contract here: the guard publishes the current Knowhere search span at the search-wrapper level so downstream code can snapshot it as a parent. The shared API contract is intentionally documented in milvus-common, where the `OpContext` field and `NestedSpanGuard` live, instead of duplicating a second contract in every downstream caller.

conanfile.py
jamesgao-jpg: Agreed. This is a dependency-settlement issue rather than a Knowhere code issue.

This PR is intentionally staged on top of unmerged tracing dependencies: Cardinal PR https://github.com/zilliztech/cardinal/pull/865 and milvus-common PR https://github.com/zilliztech/milvus-common/pull/96. We should keep this Knowhere PR gated until those land. Before merging Knowhere, I will repin Cardinal from the mutable tracing branch to an immutable mainline tag/commit and update the milvus-common reference to the corresponding merged/published package revision.

PR #1685, opened 2026-06-18
fix: [2.6] guard facade APIs from exceptions

foxspy:

## Summary
- Add guarded facade calls so C++ exceptions are converted to error-code based results.
- Mark public facade APIs as `noexcept` and preserve existing returned status values.
- Add status-category helpers and unit coverage for error-code behavior.

### DIFF
--- a/include/knowhere/expected.h
+++ b/include/knowhere/expected.h
@@ -13,9 +13,21 @@
 #define EXPECTED_H
 
 #include <cassert>
+#include <functional>
 #include <iostream>
+#include <memory>
+#include <new>
 #include <optional>
+#include <stdexcept>
 #include <string>
+#include <type_traits>
+#include <utility>
+
+#if defined(SWIG)
+#define KNOWHERE_NODISCARD
+#else
+#define KNOWHERE_NODISCARD [[nodiscard]]
+#endif
 
 namespace knowhere {
 
@@ -52,7 +64,66 @@ enum class Status {
     brute_force_inner_error = 30,
     emb_list_inner_error = 31,
     aisaq_error = 32,
+    knowhere_inner_error = 33,
 };
+
+enum class StatusCategory {
+    success = 0,
+    input_error = 1,
+    inner_error = 2,
+};
+
+inline constexpr StatusCategory
+StatusCategoryOf(knowhere::Status status) {
+    switch (status) {
+        case knowhere::Status::success:
+            return StatusCategory::success;
+        case knowhere::Status::invalid_args:
+        case knowhere::Status::invalid_param_in_json:
+        case knowhere::Status::out_of_range_in_json:
+        case knowhere::Status::type_conflict_in_json:
+        case knowhere::Status::invalid_metric_type:
+        case knowhere::Status::empty_index:
+        case knowhere::Status::not_implemented:
+        case knowhere::Status::index_not_trained:
+        case knowhere::Status::index_already_trained:
+        case knowhere::Status::invalid_value_in_json:
+        case knowhere::Status::arithmetic_overflow:
+        case knowhere::Status::invalid_binary_set:
+        case knowhere::Status::invalid_instruction_set:
+        case knowhere::Status::invalid_index_error:
+        case knowhere::Status::invalid_cluster_error:
+        case knowhere::Status::invalid_serialized_index_type:
+            return StatusCategory::input_error;
+        case knowhere::Status::faiss_inner_error:
+        case knowhere::Status::hnsw_inner_error:
+        case knowhere::Status::malloc_error:
+        case knowhere::Status::diskann_inner_error:
+        case knowhere::Status::disk_file_error:
+        case knowhere::Status::cuvs_inner_error:
+        case knowhere::Status::cuda_runtime_error:
+        case knowhere::Status::cluster_inner_error:
+        case knowhere::Status::timeout:
+        case knowhere::Status::internal_error:
+        case knowhere::Status::sparse_inner_error:
+        case knowhere::Status::brute_force_inner_error:
+        case knowhere::Status::emb_list_inner_error:
+        case knowhere::Status::aisaq_error:
+        case knowhere::Status::knowhere_inner_error:
+        default:
+            return StatusCategory::inner_error;
+    }
+}
+
+inline constexpr bool
+IsInputError(knowhere::Status status) {
+    return StatusCategoryOf(status) == StatusCategory::input_error;
+}
+
+inline constexpr bool
+IsInnerError(knowhere::Status status) {
+    return StatusCategoryOf(status) == StatusCategory::inner_error;
+}
 
 inline std::string
 Status2String(knowhere::Status status) {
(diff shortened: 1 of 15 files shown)


include/knowhere/expected.h
```diff
@@ -202,6 +275,94 @@ class expected {
...
+    using Result = std::invoke_result_t<Func, Args...>;
+    try {
+        if constexpr (std::is_void_v<Result>) {
+            std::invoke(std::forward<Func>(func), std::forward<Args>(args)...);
+            return;
+        } else {
+            return std::invoke(std::forward<Func>(func), std::forward<Args>(args)...);
+        }
+    } catch (const std::bad_alloc& e) {
+        return detail::GuardedCallFailure<Result>(Status::malloc_error,
```
liliu-z, line 354: The failure-construction path here can itself allocate, so under memory pressure that allocation throws and the exception escapes the guard, reaching `terminate` — a hole in the exact no-throw guarantee this PR exists to provide. On OOM, a guarded call that should have returned a clean error status instead crashes the whole process. Make the failure path allocation-free / `noexcept` so it cannot throw.

src/index/index.cc
```diff
@@ -39,370 +39,396 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
+        std::string msg;
+        const Status load_status = LoadConfig(cfg.get(), json, knowhere::SEARCH, "Search", &msg);
+        if (load_status != Status::success) {
+            return expected<DataSetPtr>::Err(load_status, msg);
+        }
+        // when index is immutable, bitset size should always equal to data count in index
+        // when index is mutable, it could happen that data count larger than bitset size, see
+        // https://github.com/zilliztech/knowhere/issues/70
+        // so something must be wrong at caller side when passed bitset size larger than data count
+        if (bitset_.size() > (size_t)this->Count()) {
```
liliu-z, line 144: Wrapping `Count()` in the guard collapses an internal failure to `0`, which is then surfaced as `invalid_args`/`input_error` (same pattern in `AnnIterator` and `RangeSearch`). This inverts the error category — an inner/server-side failure is misreported as a user-input error, so downstream Milvus makes the wrong retry-vs-reject decision instead of seeing the real error. Propagate the underlying error and category rather than degrading the result to `0`.

src/index/index.cc
```diff
@@ -39,370 +39,396 @@ LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, c
...
 template <typename T>
 inline const std::shared_ptr<Interrupt>
-Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json, const std::chrono::seconds timeout) {
-    auto pool = ThreadPool::GetGlobalBuildThreadPool();
-    auto interrupt = std::make_shared<Interrupt>(timeout);
-    interrupt->Set(pool->push([this, dataset, json, interrupt]() {
-        auto cfg = this->node->CreateConfig();
-        RETURN_IF_ERROR(LoadConfig(cfg.get(), json, knowhere::TRAIN, "Build"));
+Index<T>::BuildAsync(const DataSetPtr dataset, const Json& json, const std::chrono::seconds timeout) noexcept {
+    return GuardedCall([&]() -> std::shared_ptr<Interrupt> {
```
liliu-z, line 43: BuildAsync now returns `GuardedCall([&]() -> std::shared_ptr<Interrupt> {...})` (Cardinal at index.cc:43, non-Cardinal at index.cc:70). Because the return type is a bare shared_ptr with no error channel, a throw from `make_shared<Interrupt>` (OOM) or `pool->push` (e.g. a shut-down pool) takes the default-constructible branch of GuardedFailure and yields a null pointer; pre-PR these propagated as exceptions. The usual caller pattern `BuildAsync(...)->Get()` then dereferences null and crashes, converting a catchable exception into a segfault. Return an error-carrying type, or confirm the Milvus caller null-checks the returned Interrupt and document that it can be null.

include/knowhere/expected.h
```diff
@@ -52,8 +64,67 @@ enum class Status {
...
+    switch (status) {
+        case knowhere::Status::success:
+            return StatusCategory::success;
+        case knowhere::Status::invalid_args:
+        case knowhere::Status::invalid_param_in_json:
+        case knowhere::Status::out_of_range_in_json:
+        case knowhere::Status::type_conflict_in_json:
+        case knowhere::Status::invalid_metric_type:
+        case knowhere::Status::empty_index:
+        case knowhere::Status::not_implemented:
```
liliu-z, line 87: `index_not_trained`/`index_already_trained` map to `input_error` (`:87-89`) and `timeout` maps to `inner_error` (`:106`), and that category directly drives the retry-vs-reject decision. These are state/deadline conditions rather than obvious user-input, so confirm the mapping matches intended Milvus semantics — otherwise a not-trained or timeout outcome could be retried or rejected incorrectly. This is a verify-before-merge sign-off, not a confirmed bug.

include/knowhere/expected.h
```diff
@@ -113,13 +184,15 @@ Status2String(knowhere::Status status) {
...
+        case knowhere::Status::knowhere_inner_error:
+            return "knowhere inner error";
         default:
             return "unexpected status";
     }
 }
 
 template <typename T>
-class expected {
+class KNOWHERE_NODISCARD expected {
```
liliu-z, line 195: Marking `expected<T>` `[[nodiscard]]` (via KNOWHERE_NODISCARD) is a real compile-surface change, and the in-repo test_diskann.cc fix shows discard sites exist. But knowhere's own build will not break: host C++ `-Werror` is commented out (compile_flags.cmake:19) so a discard is only a warning, and the GPU build suppresses host warnings via `-Xcompiler=-w` / `--diag-suppress` (compile_flags.cmake:31-33), so the cuVS `-Werror=all-warnings` path is unaffected. The only real risk is downstream Milvus if it compiles with `-Werror` — sweep it for discarded `expected<T>` returns before merge.

src/index/index_static.cc
```diff
@@ -36,53 +43,61 @@ IndexStaticFaced<DataType>::Instance() {
...
 template <typename DataType>
 std::unique_ptr<BaseConfig>
-IndexStaticFaced<DataType>::CreateConfig(const IndexType& indexType, const IndexVersion& version) {
-    if (Instance().staticCreateConfigMap.find(indexType) != Instance().staticCreateConfigMap.end()) {
-        return Instance().staticCreateConfigMap[indexType]();
-    }
-    LOG_KNOWHERE_WARNING_ << "unhandled create config for indexType: " << indexType;
-    return std::make_unique<BaseConfig>();
+IndexStaticFaced<DataType>::CreateConfig(const IndexType& indexType, const IndexVersion& version) noexcept {
+    return GuardedCall([&]() -> std::unique_ptr<BaseConfig> {
```
liliu-z, line 47: IndexStaticFaced::CreateConfig is now `noexcept` and wraps its body in GuardedCall, so on any exception it returns a null `unique_ptr<BaseConfig>` (the default-constructible branch). In-repo callers are safe — they route the result through LoadStaticConfig, which this PR teaches to null-check — but the public static facade now hands `nullptr` to external (Milvus) callers that previously received an exception; any `CreateConfig(...)->...` dereference becomes a null-deref crash. This is the same null-sentinel hazard flagged for BuildAsync; return an error-carrying type or guarantee/document a non-null default.

alexanderguzhva:
@foxspy same story as https://github.com/zilliztech/knowhere/pull/1699 : please remove `noexcept`

alexanderguzhva:
@foxspy waiting

PR #1686, opened 2026-06-19
kind/feat: GPU HNSW index with OCQ kernel and INT8 support

premal:
## Summary

Adds `GPU_HNSW` index type for GPU-accelerated HNSW search on INT8 vectors, enabling Milvus deployments with NVIDIA GPUs to serve HNSW queries entirely on-device.

- Implements Overflow Candidate Queue (OCQ) for correct HNSW graph traversal on GPU — without OCQ, beam-width search produces incorrect results at large ef values
- Stores INT8 vectors directly on GPU (4x VRAM reduction vs FP32)
- Eager GPU upload on Deserialize; eliminates per-search allocations and reduces sync overhead
- Warp-cooperative distance computation for layer0 search
- SMEM overflow clamped to avoid silent kernel failures at large ef + high-dim vectors
- Fixes correct `QT_8bit_direct_signed` decode bias on GPU upload
- Registers `StaticCreateConfig` and `StaticHasRawData` for full Knowhere lifecycle compatibility

## Test plan

- [ ] Multi-segment lifecycle (absorption) tests pass
- [ ] INT8 bias decode and smem overflow clamp unit tests pass
- [ ] 8/8 GPU HNSW integration tests pass
- [ ] Benchmarked at 2M × 1024-dim INT8 vectors: 8,500 vec/s, 99.8% recall


foxspy:
Thanks for the contribution and for the detailed work on this PR.

After internal discussion, we do not plan to accept PRs that add or maintain GPU index implementations directly in Knowhere in this form. Going forward, GPU-related capabilities should be integrated through upstream FAISS/cuVS whenever possible, rather than carrying a separate implementation in this repository.

If this work is generally useful, we would suggest proposing the corresponding changes to FAISS or cuVS upstream. Once the capability is available there, Knowhere can pick it up through the normal dependency upgrade path.

Given that direction, we will not move forward with this PR. Thanks again for the effort and for sharing the implementation.

PR #1688, opened 2026-06-23
test: run version matrix against max version only

jamesgao-jpg:
## What changed

- Update `GenTestVersionList()` and `GenTestEmbListVersionList()` to generate only `knowhere::Version::GetMaximumVersion()`.
- Remove the now-unused Catch2 range generator include.
- Bucket GitHub Actions ccache and Conan caches by detected CPU features, following the same pattern as Cardinal commit https://github.com/zilliztech/cardinal/commit/592d5838d4ce6301281c58e567b9074bd690d24f.

## Why

PR https://github.com/zilliztech/knowhere/pull/1683 expanded the general version helper from current-version-only to current plus maximum version. That doubled version-aware UT matrices, including expensive DiskANN/AiSAQ cases. The EmbList helper also still generated the current version, so this PR keeps both helpers focused on the max supported version instead of running version 8 and version 11 paths for these callers.

While validating this PR, the GitHub-hosted `python3 wheel` job also reproduced a separate cache portability issue: a Conan cache produced on one CPU-feature set can be restored on another GitHub-hosted runner and fail at runtime with `Illegal instruction`. The cache bucket avoids reusing CPU-specific Conan/ccache artifacts across incompatible runners.

## Verification

- `grep -n 'return GENERATE(as<int32_t>{}, knowhere::Version::GetMaximumVersion().VersionNumber());' tests/ut/utils.h`
  - Shows both `GenTestVersionList()` and `GenTestEmbListVersionList()` returning max version.
- `rg -n 'from_range|catch_generators_range|GetCurrentVersion\(\)\.VersionNumber\(\)' tests/ut/utils.h`
  - No matches.
- `clang-format --dry-run --Werror tests/ut/utils.h`
- `git diff --check`
- `python3 -c 'import yaml,sys; [yaml.safe_load(open(f)) for f in sys.argv[1:]]' .github/actions/cache-restore/action.yaml .github/actions/cache-save/action.yaml`

Attempted `make WITH_UT=True BUILD_DIR=/tmp/knowhere-max-version-helper-build`, but local verification stopped before compilation because this machine has Conan 1.64.1 while the repo requires Conan >= 2.0.

## Related issue

- https://github.com/zilliztech/knowhere/issues/1691 tracks the runner storage/AiSAQ timeout behavior observed while validating this PR.

### DIFF
--- a/.github/actions/cache-restore/action.yaml
+++ b/.github/actions/cache-restore/action.yaml
@@ -9,6 +9,40 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: Detect CPU Feature Bucket
+      id: cpu_feature
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        flags=""
+        if command -v lscpu >/dev/null 2>&1; then
+          flags="$(LC_ALL=C lscpu | awk -F: '/^(Flags|Features):/ {gsub(/^[[:space:]]+/, "", $2); print $2; exit}' | tr '[:upper:]' '[:lower:]')"
+        fi
+
+        if [ -z "${flags}" ] && command -v sysctl >/dev/null 2>&1; then
+          flags="$(sysctl -a machdep.cpu.features machdep.cpu.leaf7_features 2>/dev/null | tr '[:upper:]' '[:lower:]' || true)"
+        fi
+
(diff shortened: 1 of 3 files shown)


foxspy (approved):
/lgtm

foxspy (approved):
/lgtm

PR #1689, opened 2026-06-24
fix: dispatch SVE from baseline ARM hook

CLiqing:
## Summary
- keep the ARM SIMD hook compiled for baseline AArch64 instead of SVE
- expose SVE distance declarations without requiring `__ARM_FEATURE_SVE` at the call site
- use `KNOWHERE_USE_SVE` from CMake to wire runtime SVE dispatch only when SVE objects are built

issue: #1681

## Verification
- Compiled `src/simd/hook.cc` without SVE flags and verified it has no undefined `*_sve` implementation references.
- Compiled `src/simd/hook.cc` with `-DKNOWHERE_USE_SVE` but without SVE `-march`; it references SVE implementations while containing 0 SVE instructions.
- Compiled `src/simd/distances_sve.cc` with `-march=armv8-a+sve+bf16`; it defines the SVE symbols and contains SVE instructions.


src/simd/hook.cc
```diff
@@ -383,7 +383,7 @@ fvec_hook(std::string& simd_type) {
 
 #if defined(__aarch64__)
     if (supports_sve()) {
```
liliu-z: When the toolchain lacks SVE support, KNOWHERE_USE_SVE is left undefined (CMake's 'SVE not supported on this platform' branch at libfaiss.cmake:319-321), so the `if (supports_sve())` body at lines 386-427 compiles to nothing — yet `supports_sve()` is a pure runtime getauxval(AT_HWCAP) check that still returns true on SVE hardware, so execution enters the empty branch instead of the NEON `else`. The distance function pointers then stay at their `_ref` defaults and the binary silently runs scalar code instead of NEON. Guard the `if (supports_sve())` itself under `#if defined(KNOWHERE_USE_SVE)` so no-SVE builds fall through to the NEON path.

CLiqing: Fixed in 42ded009. The outer `supports_sve()` dispatch is now guarded by `#if defined(KNOWHERE_USE_SVE)`, so builds without SVE compiler support compile the NEON assignments directly instead of entering an empty runtime-SVE branch. I also checked the other `supports_sve()` users; the SINDI paths were already guarded by `defined(__aarch64__) && defined(KNOWHERE_USE_SVE)`, so no similar fallback change was needed there.

foxspy:
/lgtm
wait ci-pass

foxspy (approved):
/lgtm

PR #1694, opened 2026-07-01
enhance: make StatusCategoryOf exhaustive (no silent default)

czs007:
issue: https://github.com/zilliztech/knowhere/issues/1695

## What

Make `StatusCategoryOf` (`include/knowhere/expected.h`) exhaustive: drop the `default:` inside its switch, rely on a post-switch fallback plus a localized `-Werror=switch`.

## Why

`StatusCategoryOf` ended its switch with `default: return inner_error`. A `default` inside the switch suppresses `-Wswitch`, so a newly added `knowhere::Status` silently falls into `inner_error` instead of being deliberately classified — and the retry / ownership decision the host derives from this category drifts with it (this is the follow-up hardening of the category helper added in #1675).

Dropping the default makes adding a `knowhere::Status` without classifying it a build error. This surfaced `cardinal_inner_error`, which was only ever caught by the default and is now classified explicitly as an inner error (behavior preserved). Adds coverage in `test_error_code.cc`.

## Part of

The segcore error-classification effort across producers: milvus-io/milvus#50903. Follow-up to #1675.

### DIFF
--- a/include/knowhere/expected.h
+++ b/include/knowhere/expected.h
@@ -73,6 +73,19 @@ enum class StatusCategory {
     inner_error = 2,
 };
 
+// Classify every knowhere::Status into a closed 3-value category. This is a
+// switch with NO `default:` plus a post-switch fallback on purpose:
+//   * a `default:` inside the switch suppresses -Wswitch, so a newly added
+//     Status would silently fall into `inner_error` instead of being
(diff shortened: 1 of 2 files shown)


foxspy (approved):
/lgtm

PR #1699, opened 2026-07-02
fix: guard iterator APIs from exceptions

foxspy:

## Summary
- Follow-up to #1675: iterator objects returned by `AnnIterator` escape the facade guard, and their deferred heavy computations (e.g. `compute_dist_func` deferred to the first `Next()` call) can throw after creation, bypassing the error-code contract.
- Change `IndexNode::iterator::Next()`/`HasNext()` to `noexcept` APIs returning `expected<>`; exceptions are converted to error codes at the `IndexIterator`/`PrecomputedDistanceIterator` public boundary. Index implementations that only override `next_batch` need no changes.
- Propagate per-query iterator errors in the default iterator-based `RangeSearch` implementation, preserving the original status instead of collapsing it.
- Add unit coverage for iterator error-code behavior (throwing `next_batch`, throwing deferred compute, exhausted iterator).

Note: this changes the public iterator interface; downstream callers need to unwrap the returned `expected<>`.

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -172,12 +172,14 @@ class IndexNode : public Object {
     };
 
     // not thread safe.
+    // Next()/HasNext() must not throw: errors are reported through the returned expected<>,
+    // consistent with the error-code based contract of the facade APIs.
     class iterator {
      public:
-        virtual std::pair<int64_t, float>
-        Next() = 0;
-        [[nodiscard]] virtual bool
-        HasNext() = 0;
+        virtual expected<std::pair<int64_t, float>>
+        Next() noexcept = 0;
+        [[nodiscard]] virtual expected<bool>
+        HasNext() noexcept = 0;
         virtual ~iterator() {
         }
     };
@@ -676,14 +678,36 @@ class IndexIterator : public IndexNode::iterator {
           use_knowhere_search_pool_(use_knowhere_search_pool) {
     }
 
-    std::pair<int64_t, float>
-    Next() override {
+    expected<std::pair<int64_t, float>>
+    Next() noexcept final {
+        return GuardedCall([&]() -> expected<std::pair<int64_t, float>> { return next_impl(); });
+    }
+
+    [[nodiscard]] expected<bool>
+    HasNext() noexcept final {
+        return GuardedCall([&]() -> expected<bool> { return has_next_impl(); });
+    }
+
+    virtual void
+    initialize() {
+        if (initialized_) {
+            throw std::runtime_error("initialize should not be called twice");
+        }
+        UpdateNext();
+        initialized_ = true;
+    }
+
+ protected:
+    // The *_impl methods may throw or return an error result; exceptions are converted to
+    //   error codes by GuardedCall at the public Next()/HasNext() boundary.
+    virtual expected<std::pair<int64_t, float>>
+    next_impl() {
         if (!initialized_) {
             initialize();
         }
         auto& q = !refine_ ? res_ : refined_res_;
         if (q.empty()) {
-            throw std::runtime_error("No more elements");
+            return expected<std::pair<int64_t, float>>::Err(Status::knowhere_inner_error, "No more elements");
         }
         auto ret = q.top();
         q.pop();
@@ -691,7 +715,7 @@ class IndexIterator : public IndexNode::iterator {
         auto update_next_func = [&]() {
             UpdateNext();
             if (retain_iterator_order_) {
-                while (HasNext()) {
+                while (!res_.empty() || !refined_res_.empty()) {
                     auto& q = !refine_ ? res_ : refined_res_;
                     auto next_ret = q.top();
                     // with the help of `sign_`, both `res_` and `refine_res` are min-heap.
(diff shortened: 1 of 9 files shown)


src/index/data_view_dense_index/index_node_with_data_view_refiner.h
```diff
@@ -279,14 +292,26 @@ class IndexNodeWithDataViewRefiner : public IndexNode {
...
      private:
+        // Called from throwing contexts (initialize/next_impl); errors from the base workspace
+        //   are rethrown and converted to error codes at the public Next()/HasNext() boundary.
         void
         UpdateNext() {
-            if (!base_workspace_->HasNext() || refine_ == false) {
+            auto base_has_next = [&]() {
+                auto has_next = base_workspace_->HasNext();
+                if (!has_next.has_value()) {
+                    throw std::runtime_error(has_next.what());
```
liliu-z: `UpdateNext()` wraps base-workspace failures in `throw std::runtime_error(...)` (here and at line 312), so the enclosing `GuardedCall` recodes every base status — e.g. `Status::malloc_error` from a deferred `bad_alloc` — as `knowhere_inner_error` and double-prefixes the message. The sibling paths in this same class (`next_impl` line 227, `has_next_impl` lines 262/267, the retain-order loop at lines 237-239) propagate the original status, so the same underlying failure surfaces as different error codes depending on `refine_ratio`. Have `UpdateNext` return a `Status` (or throw an exception type carrying the `Status` that `GuardedCall` unwraps) so `next_impl`/`has_next_impl` preserve the original code.

include/knowhere/index/index_node.h
```diff
@@ -805,35 +819,42 @@ class PrecomputedDistanceIterator : public IndexNode::iterator {
...
-            initialize();
-        }
-        if (use_knowhere_search_pool_) {
+    expected<std::pair<int64_t, float>>
+    Next() noexcept override {
+        return GuardedCall([&]() -> expected<std::pair<int64_t, float>> {
+            if (!initialized_) {
+                initialize();
+            }
+            if (!has_next_unchecked()) {
```
liliu-z, line 830: The exhaustion check consulted by `Next()`/`HasNext()` reads the unsorted slot (`next_ < results_.size() && results_[next_].id != -1`), so when a query yields more than `sort_size_` (~50k) results and bitset filtering leaves a `-1` placeholder at a `partial_sort` block boundary, the iterator reports exhaustion while valid entries remain in the unsorted tail. This is not a regression from this PR — the removed public `HasNext()` used the byte-identical unsorted predicate, so gated callers terminated at the same element count before — which is why it should be filed as a separate issue rather than patched asymmetrically here. The correct fix is symmetric: make the exhaustion predicate sort-aware or track a valid-entry count used by both `HasNext` and `Next`.

include/knowhere/index/index_node.h
```diff
@@ -172,12 +172,14 @@ class IndexNode : public Object {
...
+    // Next()/HasNext() must not throw: errors are reported through the returned expected<>,
+    // consistent with the error-code based contract of the facade APIs.
     class iterator {
      public:
-        virtual std::pair<int64_t, float>
-        Next() = 0;
-        [[nodiscard]] virtual bool
-        HasNext() = 0;
+        virtual expected<std::pair<int64_t, float>>
+        Next() noexcept = 0;
```
alexanderguzhva: I would not keep here noexcept, this is a way too strong assumption; any memory allocation using `new` may throw an exception

include/knowhere/index/index_node.h
```diff
@@ -172,12 +172,14 @@ class IndexNode : public Object {
...
+        [[nodiscard]] virtual expected<bool>
+        HasNext() noexcept = 0;
```
alexanderguzhva: same

alexanderguzhva (changes requested):

src/index/data_view_dense_index/index_node_with_data_view_refiner.h
foxspy: Good catch — fixed in f349c72c. Added `StatusException` (carries a `knowhere::Status`) with a dedicated catch in `GuardedCall` that restores the carried status and passes the message through unmodified. `UpdateNext` now rethrows base-workspace failures as `StatusException`, so the refine path surfaces the same status as the passthrough paths (e.g. `malloc_error` stays `malloc_error`), with no double prefixing. Added a unit test asserting status/message preservation.

include/knowhere/index/index_node.h
foxspy: Agreed this predates the PR — the predicate is byte-identical to the removed public `HasNext()`, so this PR keeps behavior unchanged here. Filed https://github.com/zilliztech/knowhere/issues/1716 to track the symmetric fix (sort-aware exhaustion predicate / valid-entry count).

include/knowhere/index/index_node.h
foxspy: Fair point — removed `noexcept` from both pure virtual declarations in f349c72c. The error-code contract is carried by the `expected<>` return type and documented on the interface; the built-in implementations (`IndexIterator`/`PrecomputedDistanceIterator`) keep their `noexcept` overrides since their bodies are fully wrapped in `GuardedCall` (allocation failures there are caught and reported as `Status::malloc_error`). Third-party implementations are no longer forced into the stronger guarantee.

include/knowhere/index/index_node.h
foxspy: Same — removed in f349c72c.

Issue #1703, opened 2026-07-03
[Enhancement]: three-way Status classification and knowhere-owned segcore error-code mapping


czs007:
## What

1. `StatusCategory` is currently two-way ({input_error, inner_error}), which cannot carry a retry verdict: a transient failure (OOM pressure, disk IO hiccup — a retry or replica reroute may succeed) is indistinguishable from a permanent one (capability missing, data corrupt, engine bug). Grow it to {input, transient, permanent} with `IsTransientError`.

2. The knowhere::Status → milvus::ErrorCode mapping currently lives on the **milvus** side (`KnowhereStatusToErrorCode` in internal/core/src/common/Utils.h) — a hand-maintained copy that can drift from knowhere's own view of its statuses. Move it into knowhere (producer owns classification, same convention as `milvus_storage::ToSegcoreError`), with a no-default switch + `-Werror=switch` so a newly added Status that is not classified breaks the build.

3. Three statuses are currently misclassified as input_error: `not_implemented` / `invalid_instruction_set` are capability errors (the request itself is fine, this build/CPU cannot serve it) and `invalid_serialized_index_type` is data corruption — all server-side permanent, not the caller's fault.

Part of the error-handling hardening tracked in milvus-io/milvus#50903.

PR #1704, opened 2026-07-03
enhance: three-way status classification; own the segcore code mapping

czs007:
issue: #1703

Two coupled changes that share one classification truth:

## 1. StatusCategory: {input, inner} → {input, transient, permanent}

Two-way classification cannot carry a retry verdict. `transient_error` (currently `malloc_error`, `disk_file_error` — a retry or replica reroute may succeed) is now distinct from `permanent_error`; new `IsTransientError()`. `inner_error` stays as a deprecated alias of `permanent_error`, and `IsInnerError()` matches that alias: true only for permanent errors (it has no in-tree callers, so nothing depended on a wider meaning) — existing code keeps compiling.

Three statuses move out of `input_error` to agree with the reviewed fine-grained mapping: `not_implemented` / `invalid_instruction_set` are **capability** errors (the request is fine, this build/CPU cannot serve it) and `invalid_serialized_index_type` is **data corruption** — all server-side permanent, not the caller's fault. `timeout` deliberately stays permanent: it is Cardinal-only (BuildAsync cancel-or-build-timeout) and conflates cancel with timeout; revisit once separated.

## 2. ToSegcoreErrorCode moves into knowhere

New `knowhere/segcore_error_code.h`: `ToSegcoreErrorCode(Status) -> milvus::ErrorCode`, migrated verbatim from the milvus-side `KnowhereStatusToErrorCode` — producer owns classification, same convention as `milvus_storage::ToSegcoreError`. No-default switch + `-Werror=switch`: adding a Status without classifying it breaks the build. The milvus-side copy becomes a thin delegate in a follow-up once this ships.

## Consistency lock

A test iterates every Status value and pins the two views together:
input ⟺ InvalidParameter · transient ⟺ a retriable code (MemAllocateFailed/FileReadFailed) · permanent ⟺ a non-retriable one · success ⟺ Success.

## Verification

- `[error_code]` suites: 6 cases / 156 assertions, all pass.
- Full knowhere_tests build clean on current main (milvus-common `ae43d5c`).

Part of the error-handling hardening tracked in milvus-io/milvus#50903.

### DIFF
--- a/include/knowhere/expected.h
+++ b/include/knowhere/expected.h
@@ -69,8 +69,17 @@ enum class Status {
 
 enum class StatusCategory {
     success = 0,
+    // the request itself is at fault (caller must fix it; retry is useless)
     input_error = 1,
-    inner_error = 2,
+    // server-side and permanent: retrying cannot help (capability missing,
+    // data corrupt, engine bug)
+    permanent_error = 2,
+    // server-side and transient: a retry / replica-reroute may succeed
+    // (OOM pressure, disk IO hiccup)
+    transient_error = 3,
+    // deprecated alias for the pre-three-way name; == permanent_error so
+    // existing comparisons keep compiling with unchanged semantics
+    inner_error = permanent_error,
 };
 
 // Classify every knowhere::Status into a closed 3-value category. This is a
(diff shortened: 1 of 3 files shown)


include/knowhere/expected.h
```diff
@@ -135,9 +155,17 @@ IsInputError(knowhere::Status status) {
     return StatusCategoryOf(status) == StatusCategory::input_error;
 }
 
+// "inner" keeps its historical meaning: any server-side error, i.e. NOT the
+// caller's fault -- both transient and permanent qualify.
 inline constexpr bool
 IsInnerError(knowhere::Status status) {
-    return StatusCategoryOf(status) == StatusCategory::inner_error;
+    auto category = StatusCategoryOf(status);
+    return category == StatusCategory::permanent_error || category == StatusCategory::transient_error;
```
alexanderguzhva: This function is not used inside `Knowhere` at all. So, please implement this function according to the previous `inner_error = permanent_error,` definition in order to avoid unneeded confusion.

tests/ut/test_error_code.cc
```diff
@@ -179,35 +276,46 @@ TEST_CASE("StatusCategoryOf classifies every status without a silent default", "
...
-        Status::empty_index,
-        Status::not_implemented,
-        Status::index_not_trained,
-        Status::index_already_trained,
-        Status::invalid_value_in_json,
-        Status::arithmetic_overflow,
-        Status::invalid_binary_set,
-        Status::invalid_instruction_set,
-        Status::invalid_index_error,
+        Status::invalid_args,          Status::invalid_param_in_json, Status::out_of_range_in_json,
```
alexanderguzhva: I'd rather keep it as a list, not a table

include/knowhere/expected.h
czs007: Done in b63dd27a — `IsInnerError()` now returns true only for `permanent_error`, matching the `inner_error = permanent_error` alias; updated the consistency test accordingly (transient statuses are no longer "inner").

tests/ut/test_error_code.cc
czs007: Done in b63dd27a — kept one status per line (wrapped with `// clang-format off/on`, since clang-format would otherwise re-pack this array into columns).

PR #1707, opened 2026-07-08
[2.6] fix: default ef in GPU_CAGRA CheckAndAdjust

marcelo-cjl:
## Summary
- Cherrypick #1511 to the 2.6 branch.
- Fix `GPU_CAGRA` with `adapt_for_cpu=true` when search params omit `ef`.
- Default `ef` to `max(k, 16)` for the adapt-for-cpu HNSW search path and keep explicit `ef < k` validation.
- Add regression coverage for searching a deserialized adapt-for-cpu CAGRA index without `ef`.

pr: #1511
issue: #1510

## Test plan
- [x] `git diff --check origin/2.6..HEAD`
- [x] Source-level verification that #1511 applies cleanly on `origin/2.6`


foxspy (approved):
/lgtm

PR #1708, opened 2026-07-09
Update mergify

yhmo:
Mergify reads its configuration from the repository default branch, so PRs targeting release branches are still evaluated using the Mergify rules defined on main. Because of that, release-branch PRs can be blocked by checks that are only present on some branches.

This change makes the Mergify CI rules branch-aware:
  - main and 3.x branches require the full check set, including ut on macos-15
  - 2.x branches do not require ut on macos-15, because that workflow is not present there

The rules for adding ci-passed and removing ci-passed were updated consistently so Mergify applies the correct expectations for each branch family.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,7 +3,7 @@ pull_request_rules:
     conditions:
       - or:
           - base=main
-          - base~=^2(\.\d+){1,2}$
+          - base~=^\d+(\.\d+){1,2}$
       - -status-success=DCO
     actions:
       label:
@@ -16,7 +16,7 @@ pull_request_rules:
     conditions:
       - or:
           - base=main
-          - base~=^2(\.\d+){1,2}$
+          - base~=^\d+(\.\d+){1,2}$
       - status-success=DCO
     actions:
       label:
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -151,11 +151,11 @@ pull_request_rules:
           - do-not-merge/description-tag-conflict
 
 
-  - name: Test passed for code changed for main
+  - name: Test passed for code changed for main and 3.x branches
     conditions:
       - or:
           - base=main
-          - base~=^2(\.\d+){1,2}$
+          - base~=^3(\.\d+){1,2}$
```
yhmo: .github/mergify.yml line:158
Medium ---- The top-level branch matcher was widened to any numeric release branch, but the ci-passed and failure-comment rules below still only enumerate 2.x and 3.x. If a future branch like 4.0 is cut, Mergify will match it for DCO/label gating but never add ci-passed or emit the e2e guidance, which contradicts the PR description's "any numeric release branch" claim.

alexanderguzhva:
issue: #1710 
/kind improvement

PR #1718, opened 2026-07-11
[WIP] ci: refactor 2.6 cache strategy

jamesgao-jpg:
## Question

How can Knowhere `2.6` warm reusable release-branch caches while matching the cache behavior and code style introduced for `main`?

The `2.6` Unit Test and Analyzer workflows currently warm on pushes to `main`, so direct pushes to `2.6` produce neither branch runs nor `refs/heads/2.6` caches. Existing cache keys are also split by job and assume a fixed CCache directory.

## Solution

- Warm Unit Test and Analyzer caches on direct pushes to `2.6`.
- Preserve all existing GitHub-hosted runner assignments.
- Detect CCache's configured directory, keep compiler caches job-specific, and cap uploaded CCache at 1 GB.
- Scope CCache and Conan 1 keys by stable runner class and target branch.
- Share Conan across jobs and remove job fragmentation from new keys.
- Start the new `2.6` cache namespace cold instead of retaining a migration-only OS/job fallback; successful `2.6` runs seed caches reusable by later PRs.
- Update every cache caller, including UT, wheel, Analyzer, and Release.
- Install CCache before Analyzer resolves its cache directory.
- Document the rationale beside branch scoping, cold-start behavior, dynamic path detection, and size limiting.

Knowhere PR https://github.com/zilliztech/knowhere/pull/1717 introduces the equivalent policy on `main` and is the direct behavior/style baseline for this PR. Cardinal PR https://github.com/zilliztech/cardinal/pull/889 applies the independently implemented and verifier-compared `2.6` counterpart; Cardinal PR https://github.com/zilliztech/cardinal/pull/887 provides the related Cardinal `master` change.

## Known tradeoff

The shared Conan cache is immutable after the first producer saves it. On a cold branch cache, a non-UT job may save before UT/Analyzer and omit Catch2. Builds remain correct and common dependencies remain reusable, but Catch2 may be downloaded again until the Conan key changes. This matches the accepted policy in PR #1717; selecting one dependency-superset Conan producer is a possible follow-up.

## Verification

- Independent worker implementation followed by a separate cross-repository verifier agent
- Verifier confirmed semantic and style parity with Cardinal `2.6` and Knowhere `main`
- Restore/save key symmetry checked programmatically
- All composite callers provide required inputs
- No legacy OS/job restore references remain
- All modified YAML parses successfully
- Scoped Actionlint reports no cache-refactor errors
- `git diff --check`
- DCO sign-off verified exactly

### DIFF
--- a/.github/actions/cache-restore/action.yaml
+++ b/.github/actions/cache-restore/action.yaml
@@ -1,10 +1,6 @@
 name: 'Knowhere Cache'
 description: 'Restore branch-aware Conan and hosted-runner CCache snapshots'
 inputs:
-  os:
-    description: 'OS name used only to restore legacy cache keys during migration'
-    required: true
-    default: 'ubuntu22.04'
   runner-class:
     description: 'Stable runner class such as selfhosted-linux, github-linux, or github-macos'
     required: true
@@ -52,7 +48,4 @@ runs:
         # share it. Runner class and branch prevent incompatible environments
         # and maintained branches from contaminating dependency snapshots.
         key: knowhere-${{ inputs.runner-class }}-${{ inputs.cache-branch }}-conan1-${{ hashFiles('conanfile.py') }}
-        # Keep the old OS/job prefix temporarily so existing caches seed the new layout.
-        restore-keys: |
-          knowhere-${{ inputs.runner-class }}-${{ inputs.cache-branch }}-conan1-
-          knowhere-${{ inputs.os }}-${{ github.job }}-conan-
+        restore-keys: knowhere-${{ inputs.runner-class }}-${{ inputs.cache-branch }}-conan1-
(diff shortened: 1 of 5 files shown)


foxspy:
The CCache directory lookup can fail without failing the composite action in both `.github/actions/cache-restore/action.yaml:31` and `.github/actions/cache-save/action.yaml:33`:

```bash
echo "dir=$(ccache --get-config cache_dir)" >> "${GITHUB_OUTPUT}"
```

Because the simple command is `echo`, its successful exit status masks a non-zero status from `ccache`, even with `set -euo pipefail`. This can silently publish an empty path and defer the failure to the cache step.

Please split the lookup from the output write so the query failure propagates:

```bash
cache_dir="$(ccache --get-config cache_dir)"
echo "dir=${cache_dir}" >> "${GITHUB_OUTPUT}"
```

Issue #1722, opened 2026-07-16
Improve Search end-to-end performance


foxspy:
## Description

The Search request path has opportunities to reduce fixed processing overhead and improve end-to-end latency. This issue tracks incremental performance optimizations across that path.

## Expected Behavior

Incremental optimizations should reduce end-to-end Search latency while preserving existing results, validation behavior, API compatibility, and concurrency safety.

## Actual Behavior

The current Search path retains avoidable fixed overhead outside the core search execution, which is more visible for small workloads.

PR #1723, opened 2026-07-16
feat: reuse prepared search configurations

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/1722

## Summary
- add an opt-in Search path for shared immutable prepared configurations
- cache the latest prepared configuration by exact JSON equality with concurrent miss deduplication
- preserve the existing Search path for index implementations that do not opt in
- add reuse, invalidation, and concurrency coverage

### DIFF
--- a/include/knowhere/index/index_node.h
+++ b/include/knowhere/index/index_node.h
@@ -12,7 +12,11 @@
 #ifndef INDEX_NODE_H
 #define INDEX_NODE_H
 
+#define KNOWHERE_SEARCH_CONFIG_CACHE_VERSION 1
+
+#include <atomic>
 #include <functional>
+#include <memory>
 #include <mutex>
 #include <queue>
 #include <utility>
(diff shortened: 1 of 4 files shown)


alexanderguzhva:
@foxspy This problem makes me wonder why it is needed to pass a raw json into `Index<T>::Search()` function. Wouldn't it better to expose it to the external caller in the following way:
```C++
auto version = knowhere::Version::GetCurrentVersion().VersionNumber();
auto index = knowhere::IndexFactory::Instance().Create<T>(index_type, version);

// create a dedicated config object?
auto config = CreateConfigFromJson(json);
// modify the config
...

// the search
auto search_result = index->Search(dataset, config, bitset, op_context);
```
Otherwise, every search operation is going to perform a json comparison operation, which is a recursive comparison of `std::map<std::string, json>`. 
Also, `std::hash<nlohmann::json>` is available. 
Also, according to the introduced code, it caches only 1 parsed json object.

So, maybe something like the following? Maybe, even hash is not needed.
```C++
struct JsonConfig {
    // used for the fast comparison
    uint64_t hash;
    nlohmann::json json;

    static JsonConfig CreateConfigFromJson(const nlohmann::json& input_json) {
        // save json and compute hash using std::hash<nlohmann::json>
    }

    const json& get_json() const { return json; }
};
```

I'm just wondering whether this config caching needs to be built right inside the knowhere.

---

On the other hand, the solution is totally acceptable if you'd like to avoid parsing a single json that it unlikely to change, and it is too complicated to make a complex solution.

alexanderguzhva:
@foxspy waiting for comments / unhold

PR #1727, opened 2026-07-17
update: bump cardinal (v1 ref) to v2.5.109

foxspy:
Bump the cardinal v1 reference (`cmake/libs/cardinal/v1/CMakeLists.txt`) from `v2.5.108` to `v2.5.109`.


foxspy:
issue: #1344

Issue #1730, opened 2026-07-21
Optimizations for full-text / sparse search


chasingegg:
this is the umbrella issue for full-text/sparse search, including search optimization, build optimization, index compression, code cleanup, bug fixes, etc.

PR #1737, opened 2026-07-27
make flat as the default codec for sparse index

chasingegg:
#1730 

To make the index could roll back to 2.6 version since 2.6 version knowhere does not support sparse index compression

/kind improvement


chasingegg:
closing use #1738

PR #1742, opened 2026-07-30
fix: normalize invalid cuVS search result ids

marcelo-cjl:
issue: #1744

## What

- Initialize cuVS search output id and distance buffers in the shared Knowhere cuVS wrapper before calling cuVS.
- Normalize any invalid cuVS id (`id < 0` or `id >= size`) to Knowhere's invalid result contract: id `-1` with max distance.
- Add a GPU regression test that covers filtered searches where `topk` exceeds the number of valid candidates across cuVS-backed GPU indexes.

## Why

This fixes the invalid result id issue reported in https://github.com/milvus-io/milvus/issues/51862. In the failing path, cuVS search with filtering may leave output slots unwritten when there are fewer valid candidates than `topk`; Knowhere then copied those uninitialized ids back to Milvus. Knowhere should normalize those unwritten or otherwise invalid ids before returning search results.

## Validation

- `cmake --build build/Release --target knowhere_tests -- -j16`
- `./build/Release/tests/ut/knowhere_tests "Test GPU cuVS indexes pad filtered results" --durations yes` passed, 3600 assertions
- `./build/Release/tests/ut/knowhere_tests "Test All GPU Index" -c "Test Gpu Index Search Simple Bitset" --durations yes` passed, 350 assertions

### DIFF
--- a/src/common/cuvs/integration/cuvs_knowhere_index.cuh
+++ b/src/common/cuvs/integration/cuvs_knowhere_index.cuh
@@ -33,6 +33,7 @@
 #include <raft/linalg/matrix_vector_op.cuh>
 #include <raft/linalg/normalize.cuh>
 #include <raft/linalg/reduce.cuh>
+#include <raft/matrix/init.cuh>
 #include <tuple>
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


alexanderguzhva:
issue: #1744 
/kind bug
/lgtm

PR #1749, opened 2026-08-03
Add SVS leanvec ood and static vamana support

ethanglaser:
Introduces support for SVS leanvec ood and static vamana, as well as bumping the SVS runtime bindings version to latest release (0.4.0)

### DIFF
--- a/include/knowhere/comp/index_param.h
+++ b/include/knowhere/comp/index_param.h
@@ -133,6 +133,9 @@ constexpr const char* EMB_LIST_STRATEGY_MUVERA = "muvera";
 constexpr const char* EMB_LIST_STRATEGY_LEMUR = "lemur";
 
 // Representative query sample attached to a train DataSet, for OOD-aware training.
+// The tensor is ALWAYS fp32, regardless of the index data type, and is borrowed rather than owned by the DataSet:
+// the caller must keep it alive for the duration of the build. Keeping it fp32 lets it pass unconverted through
+// IndexNodeDataMockWrapper for the fp16/bf16 index variants.
 constexpr const char* TRAIN_QUERY_TENSOR = "train_query_tensor";
 constexpr const char* TRAIN_QUERY_ROWS = "train_query_rows";
 };  // namespace meta
--- a/include/knowhere/comp/index_param.h
+++ b/include/knowhere/comp/index_param.h
@@ -221,6 +221,7 @@ constexpr const char* SVS_SEARCH_WINDOW_SIZE = "svs_search_window_size";
 constexpr const char* SVS_SEARCH_BUFFER_CAPACITY = "svs_search_buffer_capacity";
 constexpr const char* SVS_ALPHA = "svs_alpha";
 constexpr const char* SVS_STORAGE_KIND = "svs_storage_kind";
+constexpr const char* SVS_IS_STATIC = "svs_is_static";
 constexpr const char* SVS_LEANVEC_DIM = "svs_leanvec_dim";
 constexpr const char* SVS_LEANVEC_OOD = "svs_leanvec_ood";
 constexpr const char* SVS_IVF_NLIST = "svs_ivf_nlist";
--- a/include/knowhere/comp/index_param.h
+++ b/include/knowhere/comp/index_param.h
@@ -132,9 +132,9 @@ constexpr const char* EMB_LIST_STRATEGY_TOKENANN = "tokenann";
 constexpr const char* EMB_LIST_STRATEGY_MUVERA = "muvera";
 constexpr const char* EMB_LIST_STRATEGY_LEMUR = "lemur";
 
-// Out-of-distribution query-training vectors, attached to the SVS LeanVec train DataSet.
-constexpr const char* SVS_LEANVEC_QUERY_TENSOR = "svs_leanvec_query_tensor";
-constexpr const char* SVS_LEANVEC_QUERY_ROWS = "svs_leanvec_query_rows";
+// Representative query sample attached to a train DataSet, for OOD-aware training.
+constexpr const char* TRAIN_QUERY_TENSOR = "train_query_tensor";
+constexpr const char* TRAIN_QUERY_ROWS = "train_query_rows";
 };  // namespace meta
 
 namespace indexparam {
--- a/include/knowhere/comp/index_param.h
+++ b/include/knowhere/comp/index_param.h
@@ -67,6 +67,8 @@ constexpr const char* INDEX_SVS_FLAT = "SVS_FLAT";
 constexpr const char* INDEX_SVS_VAMANA = "SVS_VAMANA";
 constexpr const char* INDEX_SVS_VAMANA_LVQ = "SVS_VAMANA_LVQ";
 constexpr const char* INDEX_SVS_VAMANA_LEANVEC = "SVS_VAMANA_LEANVEC";
+constexpr const char* INDEX_SVS_IVF = "SVS_IVF";
+constexpr const char* INDEX_SVS_IVF_LEANVEC = "SVS_IVF_LEANVEC";
 
 constexpr const char* INDEX_SPARSE_INVERTED_INDEX = "SPARSE_INVERTED_INDEX";
 constexpr const char* INDEX_SPARSE_WAND = "SPARSE_WAND";
@@ -129,6 +131,10 @@ constexpr const char* EMB_LIST_RAW_INDEX = "EMB_LIST_RAW_INDEX";
 constexpr const char* EMB_LIST_STRATEGY_TOKENANN = "tokenann";
 constexpr const char* EMB_LIST_STRATEGY_MUVERA = "muvera";
 constexpr const char* EMB_LIST_STRATEGY_LEMUR = "lemur";
+
+// Out-of-distribution query-training vectors, attached to the SVS LeanVec train DataSet.
+constexpr const char* SVS_LEANVEC_QUERY_TENSOR = "svs_leanvec_query_tensor";
+constexpr const char* SVS_LEANVEC_QUERY_ROWS = "svs_leanvec_query_rows";
 };  // namespace meta
 
 namespace indexparam {
@@ -216,6 +222,10 @@ constexpr const char* SVS_SEARCH_BUFFER_CAPACITY = "svs_search_buffer_capacity";
 constexpr const char* SVS_ALPHA = "svs_alpha";
 constexpr const char* SVS_STORAGE_KIND = "svs_storage_kind";
 constexpr const char* SVS_LEANVEC_DIM = "svs_leanvec_dim";
+constexpr const char* SVS_LEANVEC_OOD = "svs_leanvec_ood";
+constexpr const char* SVS_IVF_NLIST = "svs_ivf_nlist";
+constexpr const char* SVS_IVF_NPROBE = "svs_ivf_nprobe";
+constexpr const char* SVS_IVF_K_REORDER = "svs_ivf_k_reorder";
 
 // FAISS additional Params
 constexpr const char* HNSW_REFINE = "refine";
--- a/src/index/svs/svs_config.h
+++ b/src/index/svs/svs_config.h
@@ -113,6 +113,8 @@ class SvsVamanaLeanVecConfig : public SvsVamanaConfig {
  public:
     // Dimensionality for LeanVec compression. Default is d/2 (set to 0 to use default).
     CFG_INT svs_leanvec_dim;
+    // Train the LeanVec projection for out-of-distribution queries (arXiv:2312.16335).
+    CFG_BOOL svs_leanvec_ood;
     KNOWHERE_DECLARE_CONFIG(SvsVamanaLeanVecConfig) {
         KNOWHERE_CONFIG_DECLARE_FIELD(svs_graph_max_degree)
             .description("maximum degree of the Vamana graph.")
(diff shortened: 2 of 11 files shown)


alexanderguzhva:
@ethanglaser please run `clang-format` for `tests/ut/test_index_check.cc` and `tests/ut/test_svs.cc`, there are minor formatting issues

src/index/svs/svs_config.h
```diff
@@ -66,6 +68,10 @@ class SvsVamanaConfig : public BaseConfig {
             .description("data storage format: fp32, fp16, sqi8.")
             .set_default("fp32")
             .for_train();
+        KNOWHERE_CONFIG_DECLARE_FIELD(svs_is_static)
+            .description("build an immutable static Vamana index instead of the dynamic variant.")
+            .set_default(false)
```
mihaic: The index should be static by default.
```suggestion
            .set_default(true)
```

src/index/svs/svs_vamana.cc
```diff
@@ -490,7 +491,29 @@ class SvsVamanaLeanVecIndexNode : public SvsVamanaIndexNode<DataType> {
...
+                const int64_t n_queries = dataset->Get<int64_t>(meta::SVS_LEANVEC_QUERY_ROWS);
+                std::unique_ptr<float[]> normalized_queries;
+                if (queries != nullptr && n_queries > 0) {
+                    if (is_cosine) {
+                        normalized_queries = CopyAndNormalizeVecs(queries, n_queries, dim);
+                        queries = normalized_queries.get();
+                    }
+                    idx->train_with_queries(n, data, n_queries, queries);
+                } else {
+                    LOG_KNOWHERE_WARNING_ << "svs_leanvec_ood set without query vectors; "
```
mihaic: We intend to use OOD without a query vectors sample, so this should not be a warning.
```suggestion
                    LOG_KNOWHERE_DEBUG_ << "svs_leanvec_ood set without query vectors; "
```

include/knowhere/comp/index_param.h
```diff
@@ -129,6 +131,10 @@ constexpr const char* EMB_LIST_RAW_INDEX = "EMB_LIST_RAW_INDEX";
 constexpr const char* EMB_LIST_STRATEGY_TOKENANN = "tokenann";
 constexpr const char* EMB_LIST_STRATEGY_MUVERA = "muvera";
 constexpr const char* EMB_LIST_STRATEGY_LEMUR = "lemur";
+
+// Out-of-distribution query-training vectors, attached to the SVS LeanVec train DataSet.
+constexpr const char* SVS_LEANVEC_QUERY_TENSOR = "svs_leanvec_query_tensor";
+constexpr const char* SVS_LEANVEC_QUERY_ROWS = "svs_leanvec_query_rows";
```
mihaic: In principle, these queries can be used for other hypothetical OOD methods as well, not just LeanVec. I think we should use a neutral name. If you agree with the suggestion, please make the changes everywhere.
```suggestion
constexpr const char* TRAIN_QUERY_TENSOR = "train_query_tensor";
constexpr const char* TRAIN_QUERY_ROWS = "train_query_rows";
```

src/index/svs/svs_vamana.cc
```diff
@@ -490,7 +491,29 @@ class SvsVamanaLeanVecIndexNode : public SvsVamanaIndexNode<DataType> {
             idx->search_buffer_capacity = lv_cfg.svs_search_buffer_capacity.value();
 
             // LeanVec requires training before adding vectors
-            idx->train(dataset->GetRows(), (const float*)dataset->GetTensor());
+            const auto n = dataset->GetRows();
+            const auto dim = dataset->GetDim();
+            const auto* data = (const float*)dataset->GetTensor();
+            if (lv_cfg.svs_leanvec_ood.value_or(false)) {
```
mihaic: We should use OOD by default.
```suggestion
            if (lv_cfg.svs_leanvec_ood.value_or(true)) {
```

src/index/svs/svs_config.h
```diff
@@ -153,6 +161,70 @@ class SvsVamanaLeanVecConfig : public SvsVamanaConfig {
             .set_default(0)
             .set_range(0, 65536)
             .for_train();
+        KNOWHERE_CONFIG_DECLARE_FIELD(svs_leanvec_ood)
+            .description("train the LeanVec projection for out-of-distribution queries.")
+            .set_default(false)
```
mihaic: We should use OOD by default.
```suggestion
            .set_default(true)
```

mihaic (changes requested):
I think we should use static and OOD by default.

alexanderguzhva:
@mihaic please feel free to apply these OOD changes

include/knowhere/comp/index_param.h
ethanglaser: Sounds good, changed

alexanderguzhva:
@ethanglaser fyi
```diff
tests/ut/test_index_check.cc
  ====================
  --- original
  
  +++ formatted
  
  @@ -583,8 +583,8 @@
  
           REQUIRE(IndexFactory::Instance().FeatureCheck(IndexEnum::INDEX_SVS_IVF_LEANVEC, knowhere::feature::BF16));
           REQUIRE_FALSE(
               IndexFactory::Instance().FeatureCheck(IndexEnum::INDEX_SVS_IVF_LEANVEC, knowhere::feature::BINARY));
  -        REQUIRE_FALSE(IndexFactory::Instance().FeatureCheck(IndexEnum::INDEX_SVS_IVF_LEANVEC,
  -                                                            knowhere::feature::SPARSE_U32_F32));
  +        REQUIRE_FALSE(
  +            IndexFactory::Instance().FeatureCheck(IndexEnum::INDEX_SVS_IVF_LEANVEC, knowhere::feature::SPARSE_U32_F32));
   #endif
       }
   
  tests/ut/test_svs.cc
  ====================
  --- original
  
  +++ formatted
  
  @@ -306,8 +306,7 @@
  
       };
   
       // attach the query-training sample to the train dataset for OOD training
  -    train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_TENSOR,
  -                  static_cast<const float*>(train_query_ds->GetTensor()));
  +    train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_TENSOR, static_cast<const float*>(train_query_ds->GetTensor()));
       train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_ROWS, static_cast<int64_t>(n_train_q));
   
       SECTION("Build and KNN Search") {
  @@ -526,13 +525,12 @@
  
       };
   
       // attach the query-training sample to the train dataset for OOD training
  -    train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_TENSOR,
  -                  static_cast<const float*>(train_query_ds->GetTensor()));
  +    train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_TENSOR, static_cast<const float*>(train_query_ds->GetTensor()));
       train_ds->Set(knowhere::meta::SVS_LEANVEC_QUERY_ROWS, static_cast<int64_t>(n_train_q));
   
       SECTION("Build and KNN Search") {
  -        auto idx = knowhere::IndexFactory::Instance().Create<knowhere::fp32>(
  -            knowhere::IndexEnum::INDEX_SVS_IVF_LEANVEC, version);
  +        auto idx = knowhere::IndexFactory::Instance().Create<knowhere::fp32>(knowhere::IndexEnum::INDEX_SVS_IVF_LEANVEC,
  +                                                                             version);
           REQUIRE(idx.has_value());
           auto index = idx.value();

```

mihaic (approved):

alexanderguzhva:
issue: #1752
/kind improvement

src/index/svs/svs_ivf.cc
```diff
@@ -0,0 +1,407 @@
...
+
+    // SVS IVF ingests all rows during train(); Add is a no-op so the Build (Train+Add) path
+    // does not double-insert the same dataset.
+    Status
+    Add(const DataSetPtr dataset, std::shared_ptr<Config> cfg, bool use_knowhere_build_pool) override {
+        if (!index_) {
+            LOG_KNOWHERE_ERROR_ << "Can not add data to empty index.";
+            return Status::empty_index;
+        }
+        return Status::success;
```
yhmo: src/index/svs/svs_ivf.cc line:119
High ---- This makes every public Add call after Train or Build report success without inserting the supplied rows. The underlying IndexSVSIVF is constructed with is_static=false and implements add(), so callers get silent data loss and Count() never increases. Please override Build to avoid the initial double-add while letting later Add calls delegate to index_->add (or return an explicit unsupported status), and add a Train/Build-then-Add regression test.

src/index/svs/svs_ivf.cc
```diff
@@ -0,0 +1,407 @@
...
+#include "knowhere/index/index_node_data_mock_wrapper.h"
+#include "knowhere/log.h"
+#include "knowhere/utils.h"
+
+namespace knowhere {
+
+namespace {
+
+std::optional<faiss::SVSStorageKind>
+str_to_svs_storage_kind(const std::string& s) {
```
yhmo: src/index/svs/svs_ivf.cc line:38
Low ---- This storage-kind parser duplicates the full mapping already present in svs_vamana.cc. Because SVS runtime storage kinds evolve together, the two anonymous helpers can drift and make one index reject a newly supported format. Could we move this mapping into a shared SVS helper used by both implementations?

src/index/svs/svs_ivf.cc
```diff
@@ -0,0 +1,407 @@
...
+        auto* leanvec = static_cast<faiss::IndexSVSIVFLeanVec*>(idx);
+        const auto& lv_cfg = static_cast<const SvsIvfLeanVecConfig&>(cfg);
+
+        const auto n = dataset->GetRows();
+        const auto dim = dataset->GetDim();
+        const auto* data = (const float*)dataset->GetTensor();
+        if (lv_cfg.svs_leanvec_ood.value_or(true)) {
+            // OOD: learn the projection from a representative query sample if one is attached
+            // to the train dataset, otherwise fall back to the database vectors themselves.
+            const auto* queries = dataset->Get<const float*>(meta::TRAIN_QUERY_TENSOR);
```
yhmo, line 365: src/index/svs/svs_ivf.cc line:381
Medium ---- The fp16/bf16 registrations run Build through IndexNodeDataMockWrapper, whose conversion only copies tensor/rows/dim (plus emb-list metadata), so TRAIN_QUERY_TENSOR/ROWS never reach this lookup. Those advertised LeanVec variants therefore silently fall back to database-vector training instead of using the supplied OOD sample. Please propagate and convert these fields across the mock boundary, and cover at least one fp16/bf16 OOD build; the identical Vamana path has the same issue.

src/index/svs/svs_config.h
```diff
@@ -153,6 +161,70 @@ class SvsVamanaLeanVecConfig : public SvsVamanaConfig {
...
+    }
+};
+
+class SvsIvfConfig : public BaseConfig {
+ public:
+    // Number of inverted-list centroids / clusters.
+    CFG_INT svs_ivf_nlist;
+    // Number of clusters probed during search. Larger values improve recall at the cost of latency.
+    CFG_INT svs_ivf_nprobe;
+    // Reranking multiplier for compressed datasets. <=0 keeps the index default.
```
yhmo: src/index/svs/svs_config.h line:177
Low ---- This says non-positive values preserve the index default, but 0 is accepted by the range and Train assigns it to IndexSVSIVF::k_reorder, replacing the provider default of 1.0. SearchParameters ignores non-positive overrides, so 0 remains the effective index value rather than keeping the default. Please either skip assignment for <=0 (and possibly allow a negative sentinel) or update the contract/range to describe zero as disabling reorder.

src/index/svs/svs_ivf.cc
ethanglaser: Valid, but taking a different fix than suggested. Instead of delegating a later Add to index_->add; that doesn't work because IndexSVSIVF::train() throws if impl is already set, so Train(ds) + Add(ds) would still be broken - the bug would just move. Instead svs_ivf.cc now follows the DiskANN precedent for immutable indexes: override Build to do the whole job, and return not_implemented from Train/Add rather than accepting rows it will discard.

src/index/svs/svs_ivf.cc
ethanglaser: Extracted the anonymous-namespace copies from svs_ivf.cc and svs_vamana.cc into new src/index/svs/svs_utils.h::str_to_svs_storage_kind. While consolidating, found both copies were missing lvq8x0, which exists in the faiss enum, so SVS_VAMANA_LVQ with svs_storage_kind=lvq8x0 was rejected as "unknown". Fixed in the shared version.

src/index/svs/svs_config.h
ethanglaser: The field lacked .for_train(), so Config::Load skipped it entirely under PARAM_TYPE::TRAIN and the build-time assignment was dead code. The real defect was that k_reorder could never be set at build time — it was always the faiss default of 1.0. Added .for_train() (also to svs_ivf_nprobe, same gap) plus a > 0.0f guard, since SearchParametersSVSIVF only honors positive overrides and assigning 0 would clobber the default.

src/index/svs/svs_ivf.cc
ethanglaser: Converting wasn't necessary, the sample is always fp32 and is borrowed (stored in the std::any slot, which the DataSet destructor doesn't free), so it forwarding as-is. Documented that contract at the TRAIN_QUERY_TENSOR declaration in index_param.h.

src/index/svs/svs_ivf.cc
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/index/svs/svs_ivf.cc
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/index/svs/svs_ivf.cc
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/index/svs/svs_config.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

alexanderguzhva:
@mihaic please rebase on top of the master branch , which fixes the problem with UTs. Thanks.

foxspy (approved):
/lgtm

PR #1753, opened 2026-08-05
fix: pin OpenBLAS pthread recipe

CLiqing:
issue: https://github.com/zilliztech/knowhere/issues/1754

## What

Pin Linux OpenBLAS to the known-good native pthread recipe revision available from the Conan 2 remote:

`openblas/0.3.30#aca4131c143d4c109923372e052c643c`

Keep `dynamic_arch=True`; remove the temporary `use_openmp=False` override.

## Why

An unpinned `openblas/0.3.30` resolves to the newer OpenMP recipe revision when CI runs `conan install --update`. That configuration was used by the failing CI path which crashed in `alloc_mmap -> blas_memory_alloc -> sgemm_`.

Knowhere CI uses Conan 2 and resolves recipes from `default-conan-local2`; the equivalent no-OpenMP recipe revision in that remote is `aca413…`. Pinning it makes the dependency graph reproducible without changing the global OpenBLAS recipe selection.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -138,7 +138,7 @@ class KnowhereConan(ConanFile):
         self.requires("simde/0.8.2#5e1edfd5cba92f25d79bf6ef4616b972")
         self.requires("xxhash/0.8.3#caa6d0af1b951c247922e38fbcebdbe6")
         if self.settings.os == "Linux":
-            self.requires("openblas/0.3.30")
+            self.requires("openblas/0.3.30#aca4131c143d4c109923372e052c643c")
         if not self.options.with_light:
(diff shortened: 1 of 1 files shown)


chasingegg (approved):
/lgtm

alexanderguzhva:
@CLiqing any performance changes?

CLiqing:
> @CLiqing any performance changes?

https://github.com/milvus-io/milvus/issues/50892

Just to fix a bug. The `openmp:True` configuration previously set for Knowhere was not actually taking effect; the backend was using pthreads. However, a Conan recipe update, caused this configuration to activate during Knowhere's CI build, resulting in a crash. By pinning the version now, we avoid inadvertently using the OpenMP backend; this restores the previous behavior and should not impact performance.

PR #1761, opened 2026-08-10
fix: correct DiskANN distance calculation for fully cached vectors

SpadeA-Tang:
issue: https://github.com/zilliztech/knowhere/issues/1762

Fix incorrect COSINE and INNER_PRODUCT results returned by DiskANN CalcDistByIDs when all requested vectors are cached.


alexanderguzhva:
/kind improvement
issue: #1762

foxspy (approved):
/lgtm

Issue #1687, opened 2026-06-23
Support nullable external id mapping in the index layer


marcelo-cjl:
## Description

Nullable vector fields can store only valid rows in compact raw/index storage, while Knowhere public APIs still need to expose logical/external row ids. The id mapping needs a clear owner and lifetime model in the index layer so Milvus segcore does not have to maintain another full mapping copy for sealed indexed paths.

The current nullable paths make this easy to get wrong because bitset filtering, brute-force search, iterators, serialized/reloaded indexes, and vector retrieval APIs can each cross a different internal-id domain.

## Expected Behavior

- Knowhere owns a reusable id-map snapshot for nullable compact storage and backend relayout mappings.
- Search-like IndexNode APIs return public out ids exactly once, without double final mapping in the public Index wrapper.
- Bitset count/filter logic can use mapped ids while preserving snapshot lifetime during query execution.
- Raw brute-force, iterators, GetVector/GetSparseVector, and GetEmbListByIds use the same mapping contract.
- Cardinal integrations can keep Cardinal-local reorder/source ids internally while exposing stable out ids at Knowhere boundaries.

## Actual Behavior

The nullable mapping contract is not centralized enough. Several call paths have to remember how to map bitsets or result ids themselves, which makes nullable TokenANN/EBL, serialized reload, and raw retrieve paths fragile.

## Related PRs

- zilliztech/knowhere#1673
- zilliztech/cardinal#859
- zilliztech/cardinal#860

PR #1768, opened 2026-08-11
fix: release nullable maps and separate storage-id distance APIs

marcelo-cjl:
## Summary
- Release vector-domain compact/public maps after Cardinal takes ownership of both lookup directions, while keeping EmbList list-domain lookup in Knowhere.
- Compact nullable public ids for retrieve-style APIs before backend storage access, skipping missing nullable ids consistently for vector and EmbList reads.
- Split distance calculation semantics: `CalcDistByIDs` accepts public/out ids and maps them through `IdMap`; backend/refine/rerank paths use `CalcDistByStorageIds` with storage ids.
- Expand nullable id-map coverage for malformed ids, mapped/plain mode contracts, EmbList append/list-id edge cases, handoff state, and public-vs-storage distance APIs.

issue: #1687

## Test plan
- [x] `PATH=/tmp/knowhere-ci12-bin:$PATH make pre-commit`
- [x] `cmake --build build/Release --target knowhere_tests -j8`
- [x] `make WITH_UT=True WITH_ASAN=True`
- [x] `make test`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "Test GetEmbListByIds error cases" -r compact`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "[nullable][id_map][api]" -r compact`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "[nullable][emblist][api]" -r compact`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "[nullable][ivf][emblist][add]" -r compact`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "[nullable][scann_dvr]" -r compact`
- [x] `. build/Release/generators/conanrun.sh && ./build/Release/tests/ut/knowhere_tests "[nullable][hnsw][api]" -r compact`

### DIFF
--- a/include/knowhere/id_map.h
+++ b/include/knowhere/id_map.h
@@ -436,41 +436,32 @@ class IdMap {
         }
     }
 
-    expected<int64_t>
-    CompactOutToIn(int64_t out_id, size_t compact_count) const {
-        const auto compact_id = OutCount() == 0 ? out_id : static_cast<int64_t>(GetOutToInId(out_id));
-        if (compact_id < 0 || static_cast<size_t>(compact_id) >= compact_count) {
-            return expected<int64_t>::Err(Status::invalid_args,
-                                          "selected id maps outside compact storage: " + std::to_string(out_id));
+    int64_t
+    MapOutToIn(int64_t out_id) const {
+        if (out_id < 0) {
+            return kInvalidId;
         }
-        return compact_id;
+        return OutCount() == 0 ? out_id : static_cast<int64_t>(GetOutToInId(out_id));
     }
 
-    expected<const int64_t*>
-    CompactOutToIn(const int64_t* out_ids, size_t count, std::vector<int64_t>& compact_ids,
-                   size_t compact_count) const {
+    Status
+    CompactOutToIn(const int64_t* out_ids, size_t count, std::vector<int64_t>& compact_ids) const {
         if (count != 0 && out_ids == nullptr) {
-            return expected<const int64_t*>::Err(Status::invalid_args, "selected ids are null");
-        }
-        if (OutCount() == 0) {
-            for (size_t i = 0; i < count; ++i) {
-                auto compact_id = CompactOutToIn(out_ids[i], compact_count);
-                if (!compact_id.has_value()) {
-                    return expected<const int64_t*>::Err(compact_id.error(), compact_id.what());
-                }
-            }
-            return out_ids;
+            return Status::invalid_args;
         }
 
-        compact_ids.resize(count);
+        // Retrieve-style APIs may receive arbitrary public ids. Keep only ids
+        // that have a stored vector payload; distance-refine callers already
+        // pass compact/internal ids and must not use this helper.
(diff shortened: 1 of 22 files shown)


foxspy (approved):
/lgtm

foxspy (changes requested):
Blocking on the current head: `CompactOutToIn()` silently drops every unmapped ID and shortens the request, but `GetVectorByIds()` / `GetEmbListByIds()` return only a tensor/offsets without result IDs or input positions. For a request like `[valid A, nullable hole B, valid C]`, the caller receives two rows and cannot determine that row 1 corresponds to C rather than B. Please either preserve one output slot per input, return an explicit input-to-output/ID mapping, or reject mixed partial results instead of silently compacting them. This also needs a regression test that verifies the public result contract, not only the returned tensor values.

foxspy:
Re-checked current head `91519d52`: the earlier OOB concern from `ef03565c` is fixed by filtering negative/out-of-range ids in `GetEmbListByIds()` before offset access, with a second bounds check in `GetEmbListByStorageIds()`. The remaining API-domain split is intentional and documented by the author: Knowhere returns surviving compact payload rows in input order, while Milvus performs final nullable result composition from original ids/metadata. I am withdrawing this as a current-head blocker; please keep the public-vs-storage contract documentation synchronized.

marcelo-cjl:
@foxspy This keeps the same nullable selected-id boundary as before the pushdown. `GetVectorByIds()` / `GetEmbListByIds()` take public ids at the API boundary, then compact them through `CompactOutToIn()` before touching backend storage. Nullable holes or sparse public ids are therefore not used as compact storage ids and cannot cause out-of-range backend access. Milvus owns the final result composition from the original requested ids and nullable metadata; Knowhere retrieves only the surviving vector payload rows, in surviving input order. The API contract here is compacted payload rows, not the final nullable selected-id result.

marcelo-cjl:
@foxspy This split is intentional. `CalcDistByIDs()` stays as the public-id entry point, and `CalcDistByStorageIds()` is the backend/rerank primitive for ids already in the storage domain. The old override pattern mixed public ids and storage ids in one virtual surface, which is exactly what this PR is removing. In-tree backends and wrappers are already migrated to the new boundary, and keeping deprecated virtual shims would preserve the ambiguous mixed-domain contract.

foxspy (approved):
Re-reviewed current head `91519d52`. The previous mixed-result concern is resolved by the documented API boundary: Knowhere returns compacted surviving payload rows in input order, while Milvus owns final nullable selected-id composition from the original requested ids and nullable metadata. The old `GetEmbListByIds` OOB finding applied to an earlier head; current code filters out-of-range ids before offset access and the exact-head Ubuntu/macOS/ASan matrix is green.

PR #1773, opened 2026-08-13
fix: EmbList serialization compatibility across index versions

SpadeA-Tang:
issue: https://github.com/zilliztech/knowhere/issues/1774


foxspy (approved):
/lgtm

PR #1776, opened 2026-08-15
fix: keep current index version at 8

marcelo-cjl:
## What

Keep the default/current index version at 8 while leaving version 11 available as the maximum supported nullable id-map format.

Also update nullable emb-list tests so MUVERA/LEMUR scenarios explicitly use the emb-list meta v2 minimum version instead of inheriting the default current version.

## Why

#1768 correctly introduced the nullable/version-gated format, but making it the default current version would publish that development format by default. Restoring current_version to 8 preserves existing default index behavior.

After restoring current_version, the nullable MUVERA/LEMUR emb-list UTs must opt into version 11 because their serialized metadata is intentionally rejected for versions below the emb-list meta v2 gate.

Related to #1687.

## Validation

- git diff --check
- make BUILD_DIR=/home/ubuntu/skills/issue_review/workspaces/knowhere_pr_1776_main/build_ut_asan WITH_UT=True WITH_ASAN=True
- build_ut_asan/Release/tests/ut/knowhere_tests "Nullable IVF_FLAT MUVERA/LEMUR emb-list APIs rebuild list ids after reload"
- build_ut_asan/Release/tests/ut/knowhere_tests "Nullable native matrix covers every IndexNode operation"

### DIFF
--- a/include/knowhere/version.h
+++ b/include/knowhere/version.h
@@ -21,8 +21,8 @@ namespace knowhere {
 namespace {
 static constexpr int32_t default_version = 0;
 static constexpr int32_t minimal_version = 0;
-static constexpr int32_t current_version = 12;
-static constexpr int32_t maximum_version = 12;
+static constexpr int32_t current_version = 8;
+static constexpr int32_t maximum_version = 11;
 }  // namespace
 
 class Version {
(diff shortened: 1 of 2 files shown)


foxspy (approved):
/lgtm

PR #1781, opened 2026-08-18
[2.6] fix diskann ut

chasingegg:
/kind improvement

similar fix on test_diskann.cc in #1768 , port to 2.6 branch. ut fails with cardinal build after #1764


foxspy (approved):
/lgtm

Issue #1769, opened 2026-08-12
Benchmark IVF_FLAT_CC and SCANN_DVR train performance


marcelo-cjl:
## Summary

Track and improve train performance of `IVF_FLAT_CC` and `SCANN_DVR`.

The train-only path (building the coarse quantizer + index model, excluding
Add) is a significant portion of index build cost for large datasets. This
issue tracks benchmarking the train path and the optimizations applied to it.

## Scope

- Benchmark train-only cost for `IVF_FLAT_CC` and `SCANN_DVR`.
- Provide a reproducible benchmark setup supporting multiple local datasets
  (start with the SIFT fbin dataset under `~/data/sift`).
- Measure train/build-model cost only; do not include Add time.
- Record timing and memory signals for later performance analysis.

## Progress

- Benchmark harness and dataset runner added in `benchmark/`.
- `SCANN_DVR` coarse quantizer training optimized: SuperKMeans
  (spherical inner-product) enabled by default, recall-equivalent to
  Clustering but ~60% faster full build on qwen 4096-dim.
  See PR #1783.

PR #1783, opened 2026-08-20
enhance: enable SuperKMeans by default for CPU IVF training

marcelo-cjl:
issue: #1769

## Summary
- Enable the existing Faiss SuperKMeans coarse-training path by default for CPU numeric IVF variants, including SCANN. Honor explicit use_super_kmeans=false.
- Fall back to Clustering below Faiss's minimum supported dimension (32 with the current defaults), without adding a minimum nlist requirement. Binary IVF and GPU/cuVS training remain unchanged and default the option to false.
- Reuse the Faiss snapshot already upgraded in main by PR #1811. This revised PR contains no vendored Faiss algorithm changes.
- Keep functional coverage in the existing test_cluster.cc and test_search.cc files; benchmark scripts and raw results remain outside the PR.
- Build performance (single run per configuration):

## Build performance

Compared the same Release library with `use_super_kmeans=false` (Clustering) versus `true` (SuperKMeans). **Build below is measured public Train + Add time**, excluding dataset loading and queries; it is not Train alone. SCANN includes PQ training/encoding and raw-vector storage.

| Dataset | Index | Clustering Build (s) | SuperKMeans Build (s) | Speedup | Time reduction |
| --- | --- | ---: | ---: | ---: | ---: |
| SIFT | IVF_FLAT | 11.57 | 8.95 | 1.29x | 22.6% |
| SIFT | SCANN | 12.51 | 9.91 | 1.26x | 20.8% |
| Cohere | IVF_FLAT | 64.00 | 43.70 | 1.46x | 31.7% |
| Cohere | SCANN | 70.01 | 50.29 | 1.39x | 28.2% |
| OpenAI | IVF_FLAT | 104.01 | 62.65 | 1.66x | 39.8% |
| OpenAI | SCANN | 112.26 | 69.56 | 1.61x | 38.0% |

Data: SIFT 1,000,000 x 128 (L2); Cohere 1,000,000 x 768 (COSINE); OpenAI 500,000 x 1536 (COSINE). Full base datasets, float32, nlist=1024, 8 build threads, OMP=8, BLAS=1. SCANN: sub_dim=2, with_raw_data=true. AMD EPYC 7R13 x86_64/AVX2 host, GCC 12, Release, no ASan. The implementation's default seed/settings are otherwise unchanged. Only the first build of each configuration is reported; no averaging or selection by performance. This is a single-run observation on a shared host, not a repeatability guarantee.

## Test Plan
- [x] Native Release build of the current PR source and unit-test binary.
- [x] Targeted functional tests: knowhere_tests '[superkmeans],Test SuperKMeans*' — 3 test cases, 44 assertions passed.
- [x] Full-base IVF_FLAT/SCANN build comparison on SIFT, Cohere and OpenAI; single-run results above.
- [ ] Full unit-test suite and GPU builds were not run (outside the requested validation scope).

### DIFF
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -682,11 +682,11 @@ IvfIndexNode<DataType, IndexType>::TrainInternal(const DataSetPtr dataset, std::
         }
         // apply clustering config
         ApplyClusteringConfig(base_index->cp);
-        // SuperKMeans is designed for d >= 2 * d_prime_min (default 32) and
-        // large nlist (k >= 1024). Outside that range fall back to Clustering
-        // instead of failing the build or degrading recall.
+        // SuperKMeans requires d >= 2 * d_prime_min (default 32). Fall back to
+        // Clustering below that hard limit; use_super_kmeans otherwise remains
+        // the user's choice, including for small nlist values.
         bool use_super_kmeans = scann_cfg.use_super_kmeans.value();
-        if (use_super_kmeans && (dim < 2 * faiss::SuperKMeansParameters{}.d_prime_min || nlist < 1024)) {
+        if (use_super_kmeans && dim < 2 * faiss::SuperKMeansParameters{}.d_prime_min) {
             use_super_kmeans = false;
         }
         base_index->cp.use_super_kmeans = use_super_kmeans;
--- a/src/index/ivf/ivf.cc
+++ b/src/index/ivf/ivf.cc
@@ -19,6 +19,7 @@
 #include "faiss/IndexIVFRaBitQ.h"
 #include "faiss/IndexIVFRaBitQFastScan.h"
 #include "faiss/IndexRefine.h"
+#include "faiss/SuperKMeans.h"
 #include "faiss/VectorTransform.h"
 #include "faiss/cppcontrib/knowhere/IndexBinaryFlat.h"
 #include "faiss/cppcontrib/knowhere/IndexBinaryIVF.h"
@@ -599,6 +600,7 @@ IvfIndexNode<DataType, IndexType>::TrainInternal(const DataSetPtr dataset, std::
         }
         // apply clustering config
         ApplyClusteringConfig(index->cp);
+        index->cp.use_super_kmeans = ivf_flat_cfg.use_super_kmeans.value();
         // train
         index->train(rows, static_cast<const float*>(data));
         // transfer ownership of qzr to index
@@ -621,6 +623,7 @@ IvfIndexNode<DataType, IndexType>::TrainInternal(const DataSetPtr dataset, std::
         }
         // apply clustering config
         ApplyClusteringConfig(index->cp);
+        index->cp.use_super_kmeans = ivf_flat_cc_cfg.use_super_kmeans.value();
         // train
         index->train(rows, static_cast<const float*>(data));
         // transfer ownership of qzr to index
@@ -652,6 +655,7 @@ IvfIndexNode<DataType, IndexType>::TrainInternal(const DataSetPtr dataset, std::
 
         // apply clustering config
         ApplyClusteringConfig(index->get_base_ivf_index()->cp);
+        index->get_base_ivf_index()->cp.use_super_kmeans = ivf_pq_cfg.use_super_kmeans.value();
         // train
         index->train(rows, static_cast<const float*>(data));
         // transfer ownership of qzr to index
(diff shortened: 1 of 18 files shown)


marcelo-cjl:
@foxspy @alexanderguzhva Please review this when you have some free time.

src/index/ivf/ivf.cc
```diff
@@ -678,6 +682,14 @@ IvfIndexNode<DataType, IndexType>::TrainInternal(const DataSetPtr dataset, std::
         }
         // apply clustering config
         ApplyClusteringConfig(base_index->cp);
+        // SuperKMeans is designed for d >= 2 * d_prime_min (default 32) and
+        // large nlist (k >= 1024). Outside that range fall back to Clustering
+        // instead of failing the build or degrading recall.
+        bool use_super_kmeans = scann_cfg.use_super_kmeans.value();
+        if (use_super_kmeans && (dim < 2 * faiss::SuperKMeansParameters{}.d_prime_min || nlist < 1024)) {
```
alexanderguzhva: This is questionable, because I constantly use SuperKMeans for 256 centroids. What is the justification and why not allow a user to decide?

thirdparty/faiss/faiss/utils/simd_impl/distances_arm_sve.cpp
```diff
@@ -520,13 +523,138 @@ void fvec_L2sqr_ny<SIMDLevel::ARM_SVE>(
...
+    svfloat32_t xv = svdup_n_f32(0.0f);
+    svfloat32_t xacc = svld1_f32(pg, x);
+    xv = svsel_f32(pg, xacc, xv);
+    float global_min = HUGE_VALF;
+    size_t global_idx = 0;
+    for (size_t c = 0; c < ny; ++c) {
+        svfloat32_t yv = svld1_f32(pg, y + c * D);
+        svfloat32_t diff = svsub_f32_x(pg, yv, xv);
+        svfloat32_t sq = svmul_f32_x(pg, diff, diff);
+        const float dis = svaddv_f32(pg, sq);
```
alexanderguzhva: a faster way would be to maintain `svfloat32_t global_mins` and `svuint32_t global_ids`. But this one works as well, although it is performance-wise suboptimal

thirdparty/faiss/faiss/SuperKMeans.cpp
```diff
@@ -57,9 +57,9 @@ namespace faiss {
 namespace {
 
 struct TrainState {
-    /// Orthogonal rotation. Train in rotated space (X_tilde = X * R);
```
alexanderguzhva, line 60: please keep the comment

thirdparty/faiss/faiss/SuperKMeans.cpp
```diff
@@ -198,7 +208,8 @@ int update_centroids_and_split(
         int k,
         TrainState& state,
         std::vector<int64_t>& labels64,
-        std::vector<float>& hassign) {
+        std::vector<float>& hassign,
+        bool spherical) {
```
alexanderguzhva: the question from the Faiss team would be the following: why do you need to have `spherical` here at all, instead of pre-normalizing the input data outside of this code?

src/index/ivf/ivf.cc
marcelo-cjl: Good point. The dimension check is a hard SuperKMeans precondition, but k >= 1024 is only a usage recommendation. I removed the nlist-based override so use_super_kmeans=true is honored for 256 centroids as well, kept the low-dimension fallback required by d_prime_min, and updated the tests so the enabled SCANN case actually exercises SuperKMeans.

thirdparty/faiss/faiss/utils/simd_impl/distances_arm_sve.cpp
marcelo-cjl: Agreed. I changed the low-dimensional SVE loop to keep per-lane global minima and ids across centroid batches, with predicated tail handling and first-index tie semantics, so the final horizontal reduction is performed only once. I also added boundary and tie coverage for D2/D4/D8.

thirdparty/faiss/faiss/SuperKMeans.cpp
marcelo-cjl: Done.

thirdparty/faiss/faiss/SuperKMeans.cpp
marcelo-cjl: We still need spherical centroid normalization inside the iteration: pre-normalizing the input is not equivalent for raw inner-product data, and centroid means do not remain unit-normalized. However, I removed spherical from update_centroids_and_split and aligned the flow with Clustering: update and split first, then apply spherical centroid post-processing once. This keeps the unit-centroid invariant without changing raw IP into cosine semantics.

alexanderguzhva:
issue: #1823 
/kind improvement

PR #1792, opened 2026-08-25
fix: preserve SVS all-pass selector semantics

jamesgao-jpg:
## What

- Keep an explicit all-pass bitmap on the SVS Vamana selector path even when its known filtered count is zero.
- Apply the same dispatch rule to kNN and range search.
- Add regression coverage proving that known-zero and unknown-count all-pass bitsets produce identical results.

## Why

PR #1673 changed `BitsetView::empty()` so a full-sized bitmap with a known filtered count of zero is considered empty. vecTool passes unfiltered searches as exactly that representation, so SVS Vamana stopped constructing its all-pass selector and switched to the native unfiltered search path.

The two SVS paths use different search heuristics and produced different recall at the same search window. This change preserves the global `BitsetView::empty()` semantics introduced by #1673 while retaining the historical SVS behavior whenever a caller explicitly supplies a bitmap.

This is related to the all-clear SVS handling previously addressed for SVS Flat in #1603.

## Reproduction

Using one unchanged vecTool-built SVS Vamana index:

- `BitsetView(bits, N, 0)`: recall at windows 32/64/128 was `0.744950 / 0.854140 / 0.927569`.
- `BitsetView(bits, N)`: recall was `0.905918 / 0.905918 / 0.927569`.

Alternating the two views against the same serialized index reproduced the result deterministically without changing the index artifact.

## Verification

- `git diff --check`
- Focused x86 Knowhere unit-test build is in progress; CI exercises the same SVS-enabled test target.

### DIFF
--- a/src/index/svs/svs_vamana.cc
+++ b/src/index/svs/svs_vamana.cc
@@ -158,7 +158,7 @@ class SvsVamanaIndexNode : public IndexNode {
                     sp.search_window_size = v_cfg.svs_search_window_size.value();
                     sp.search_buffer_capacity = v_cfg.svs_search_buffer_capacity.value();
                     std::unique_ptr<BitsetViewIDSelector> bw_idselector;
-                    if (!bitset.empty()) {
+                    if (!bitset.empty() || bitset.num_bits() != 0) {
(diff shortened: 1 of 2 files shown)


foxspy (approved):
/lgtm

PR #1797, opened 2026-08-27
bump milvus-common version

chasingegg:
/kind improvement


chasingegg:
milvus related: https://github.com/milvus-io/milvus/pull/52670

Issue #1747, opened 2026-07-31
[Feature]: Support RaBitQ quantization for HNSW and DiskANN (HNSW_RBQ / DISKANN_RBQ)


foxspy:
## Motivation

Knowhere currently provides RaBitQ support through the IVF-based `IVF_RABITQ` / `IVF_RABITQ_FASTSCAN` paths, but it does not provide RaBitQ-quantized variants for HNSW or DiskANN.

Adding RaBitQ support to these graph-based indexes would allow us to evaluate a higher-accuracy, low-bit quantization option for graph traversal and refinement, especially for high-dimensional or large-scale workloads.

Related Milvus proposal: https://github.com/milvus-io/milvus/issues/51618

## Requested support

Add RaBitQ quantization support for:

- `HNSW_RBQ`
- `DISKANN_RBQ`

The final public index names may be adjusted to remain consistent with Knowhere's existing naming conventions.

## Scope

- Reuse the existing RaBitQ distance/quantization implementation where applicable.
- Define the build, load, search, serialization, and configuration paths for both index types.
- Preserve compatibility with the existing non-RaBitQ HNSW and DiskANN implementations.
- For DiskANN, define how the RaBitQ reference frame and optional full-precision refinement integrate with graph traversal and SSD reads.
- Add deterministic correctness, serialization, recall, and boundary tests.
- Add benchmarks comparing recall, QPS/latency, memory usage, refined-vector count, and SSD reads against the existing HNSW/DiskANN quantization paths.

## Acceptance criteria

- Both index types can be built, saved, loaded, and searched through the public Knowhere APIs.
- Search parameters and validation are documented and reject unsupported combinations clearly.
- Recall and performance results are reported on representative datasets and dimensions.
- Existing HNSW, DiskANN, and IVF RaBitQ tests remain green.

Sallery-X:
Could I take this？ Any progress？

PR #1798, opened 2026-08-28
feat: add high-performance HNSW RaBitQ index

CLiqing:
issue: #1747

## Summary

- introduce `HNSW_RABITQ` with 1-8 bit dense database codes and FP32 queries (`qb=0`), supporting L2, inner product, and cosine similarity
- build the HNSW graph with exact FP32 distances, then attach `RandomRotationMatrix -> IndexRaBitQ` storage for search
- implement cosine semantics without modifying the input dataset: retain per-vector inverse norms and apply query/database norm correction during distance computation
- add optimized scalar, AVX2, AVX-512, and Sapphire Rapids single/batch-4 multi-bit distance kernels

## Compatibility and scope

- `HNSW_RABITQ` V1 is immutable after its initial build and supports L2/IP/COSINE
- database codes use the bundled-Faiss dense RaBitQ layout; query quantization is intentionally not exposed in V1, and nonzero `rbq_bits_query` is rejected explicitly
- the HNSW wrapper exposes RBQ1-RBQ8; the bundled generic `IndexRaBitQ` retains its existing 1-9 bit range
- cosine coverage includes search, range search, iterator, distance-by-ID, FP32 refinement, and serialization/deserialization; Cohere and OpenAI checks agree with pre-normalized IP within 0.0004 recall
- this implementation is not index-format or runtime-behavior compatible with Faiss #5526: it uses an exact FP32 graph followed by random rotation and full multi-bit L2/IP scoring, whereas [#5526](https://github.com/facebookresearch/faiss/pull/5526) builds directly with RaBitQ storage and uses staged L2 search
- the two implementations can coexist in source: the Knowhere-specific type is under `faiss::cppcontrib::knowhere` and uses `IHRK`/`IHRC`, while `IHNr` remains reserved for #5526; shared bundled-Faiss quantizer/SIMD changes will still require normal reconciliation during future Faiss updates

### DIFF
--- a/src/index/hnsw/faiss_hnsw.cc
+++ b/src/index/hnsw/faiss_hnsw.cc
@@ -9,10 +9,13 @@
 // is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 // or implied. See the License for the specific language governing permissions and limitations under the License.
 
+#include <faiss/IndexPreTransform.h>
+#include <faiss/IndexRaBitQ.h>
 #include <faiss/cppcontrib/knowhere/IndexBinaryScalarQuantizer.h>
 #include <faiss/cppcontrib/knowhere/IndexCosine.h>
 #include <faiss/cppcontrib/knowhere/IndexFlat.h>
 #include <faiss/cppcontrib/knowhere/IndexHNSWBinary.h>
+#include <faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.h>
 #include <faiss/cppcontrib/knowhere/IndexSQ4Uniform.h>
 #include <faiss/cppcontrib/knowhere/MetricType.h>
 #include <faiss/cppcontrib/knowhere/impl/CountSizeIOWriter.h>
@@ -672,31 +675,41 @@ convert_ds_to_float(const DataSetPtr& src, DataFormatEnum data_format) {
 }
 
 Status
-add_to_index(faiss::Index* const __restrict index, const DataSetPtr& dataset, const DataFormatEnum data_format) {
+add_to_index(faiss::Index* const __restrict index, const DataSetPtr& dataset, const DataFormatEnum data_format,
+             bool chunk_fp32 = false) {
     const auto* data = dataset->GetTensor();
     const auto rows = dataset->GetRows();
     const auto dim = dataset->GetDim();
 
-    if (data_format == DataFormatEnum::fp32) {
+    if (data_format == DataFormatEnum::fp32 && !chunk_fp32) {
         // add as is
         index->add(rows, reinterpret_cast<const float*>(data));
     } else {
         // convert data into float in pieces and add to the index
         constexpr int64_t n_tmp_rows = 4096;
-        std::unique_ptr<float[]> tmp = std::make_unique<float[]>(n_tmp_rows * dim);
+        std::unique_ptr<float[]> tmp;
+        if (data_format != DataFormatEnum::fp32) {
+            tmp = std::make_unique<float[]>(n_tmp_rows * dim);
+        }
 
         for (int64_t irow = 0; irow < rows; irow += n_tmp_rows) {
             const int64_t start_row = irow;
             const int64_t end_row = std::min(rows, start_row + n_tmp_rows);
             const int64_t count_rows = end_row - start_row;
 
-            if (!convert_rows_to_fp32(data, tmp.get(), data_format, start_row, count_rows, dim)) {
-                LOG_KNOWHERE_ERROR_ << "Unsupported data format";
-                return Status::invalid_args;
+            const float* chunk = nullptr;
+            if (data_format == DataFormatEnum::fp32) {
+                chunk = reinterpret_cast<const float*>(data) + start_row * dim;
+            } else {
+                if (!convert_rows_to_fp32(data, tmp.get(), data_format, start_row, count_rows, dim)) {
+                    LOG_KNOWHERE_ERROR_ << "Unsupported data format";
+                    return Status::invalid_args;
+                }
+                chunk = tmp.get();
             }
 
             // add
-            index->add(count_rows, tmp.get());
+            index->add(count_rows, chunk);
         }
     }
 
@@ -3032,6 +3045,235 @@ class BaseFaissRegularIndexHNSWPQNodeTemplate : public BaseFaissRegularIndexHNSW
     }
 };
 
+// Build an exact HNSW graph with Flat storage, train RaBitQ independently,
+// then replace the Flat storage after both indexes have received the data.
+// RaBitQ does not currently provide the symmetric code-to-code distance that
+// HNSW graph construction requires, so the finalized index is intentionally
+// immutable.
+class BaseFaissRegularIndexHNSWRaBitQNode : public BaseFaissRegularIndexHNSWNode {
+ public:
+    BaseFaissRegularIndexHNSWRaBitQNode(const int32_t& version, const Object& object, DataFormatEnum data_format)
+        : BaseFaissRegularIndexHNSWNode(version, object, data_format) {
+    }
+
+    static std::unique_ptr<BaseConfig>
+    StaticCreateConfig() {
+        return std::make_unique<FaissHnswRaBitQConfig>();
+    }
+
+    std::unique_ptr<BaseConfig>
+    CreateConfig() const override {
+        return StaticCreateConfig();
+    }
+
+    std::string
+    Type() const override {
+        return knowhere::IndexEnum::INDEX_HNSW_RABITQ;
+    }
+
+    bool
+    IsAdditionalScalarSupported(bool) const override {
+        return false;
+    }
+
+    bool
+    IsIndexRefineEnabled() const override {
+        return !indexes.empty() && std::all_of(indexes.begin(), indexes.end(), [](const auto& index) {
+            return index != nullptr &&
+                   dynamic_cast<const faiss::cppcontrib::knowhere::IndexRefine*>(index.get()) != nullptr;
+        });
+    }
+
+ protected:
+    std::vector<std::unique_ptr<faiss::IndexPreTransform>> tmp_index_rabitq;
+
+    Status
+    TrainInternal(const DataSetPtr dataset, const Config& cfg) override {
+        const auto rows = dataset->GetRows();
+        const auto dim = dataset->GetDim();
+        const auto& hnsw_cfg = static_cast<const FaissHnswRaBitQConfig&>(cfg);
+
+        auto metric = Str2FaissMetricType(hnsw_cfg.metric_type.value());
+        if (!metric.has_value() ||
+            (metric.value() != faiss::METRIC_L2 && metric.value() != faiss::METRIC_INNER_PRODUCT)) {
+            LOG_KNOWHERE_ERROR_ << "HNSW_RABITQ only supports L2, IP and COSINE metrics";
+            return Status::invalid_metric_type;
+        }
+        const bool is_cosine = IsMetricType(hnsw_cfg.metric_type.value(), metric::COSINE);
+        const auto& scalar_info_map =
+            dataset->Get<std::unordered_map<int64_t, std::vector<std::vector<uint32_t>>>>(meta::SCALAR_INFO);
+        if (!scalar_info_map.empty()) {
+            LOG_KNOWHERE_ERROR_ << "HNSW_RABITQ does not support building with scalar info";
+            return Status::invalid_args;
+        }
+
+        auto float_ds_ptr = convert_ds_to_float(dataset, data_format);
+        if (float_ds_ptr == nullptr) {
+            LOG_KNOWHERE_ERROR_ << "Unsupported data format";
+            return Status::invalid_args;
+        }
+        const auto* data = static_cast<const float*>(float_ds_ptr->GetTensor());
+
+        try {
+            std::unique_ptr<faiss::cppcontrib::knowhere::IndexHNSW> hnsw_index;
+            if (is_cosine) {
+                hnsw_index =
+                    std::make_unique<faiss::cppcontrib::knowhere::IndexHNSWFlatCosine>(dim, hnsw_cfg.M.value());
+            } else {
+                hnsw_index = std::make_unique<faiss::cppcontrib::knowhere::IndexHNSWFlat>(dim, hnsw_cfg.M.value(),
+                                                                                          metric.value());
+            }
+            hnsw_index->hnsw.efConstruction = hnsw_cfg.efConstruction.value();
+
+            const auto rbq_bits = static_cast<uint8_t>(hnsw_cfg.rbq_bits.value());
+            auto rabitq_index = std::make_unique<faiss::IndexRaBitQ>(dim, metric.value(), rbq_bits, rbq_bits > 1);
+            // HNSW_RABITQ V1 always uses an FP32 query.
+            rabitq_index->qb = 0;
+            rabitq_index->centered = false;
+            auto rotation = std::make_unique<faiss::RandomRotationMatrix>(dim, dim);
+            std::unique_ptr<faiss::IndexPreTransform> transformed_rabitq;
+            if (is_cosine) {
+                transformed_rabitq = std::make_unique<faiss::cppcontrib::knowhere::IndexPreTransformRaBitQCosine>(
+                    rotation.get(), rabitq_index.get());
+            } else {
+                transformed_rabitq = std::make_unique<faiss::IndexPreTransform>(rotation.get(), rabitq_index.get());
+            }
+            transformed_rabitq->own_fields = true;
+            rotation.release();
+            rabitq_index.release();
+
+            std::unique_ptr<faiss::Index> final_index;
+            if (hnsw_cfg.refine.value_or(false) && hnsw_cfg.refine_type.has_value()) {
+                const auto hnsw_d = hnsw_index->storage->d;
+                const auto hnsw_metric_type = hnsw_index->storage->metric_type;
+                auto final_index_cnd = pick_refine_index(data_format, hnsw_cfg.refine_type, std::move(hnsw_index),
+                                                         hnsw_d, hnsw_metric_type);
+                if (!final_index_cnd.has_value()) {
+                    return Status::invalid_args;
+                }
+                final_index = std::move(final_index_cnd.value());
+            } else {
+                final_index = std::move(hnsw_index);
+            }
+
+            LOG_KNOWHERE_INFO_ << "Training exact HNSW graph storage";
+            final_index->train(rows, data);
+            LOG_KNOWHERE_INFO_ << "Training RaBitQ storage";
+            transformed_rabitq->train(rows, data);
+
+            indexes[0] = std::move(final_index);
+            tmp_index_rabitq.clear();
+            tmp_index_rabitq.emplace_back(std::move(transformed_rabitq));
+        } catch (const std::exception& e) {
+            LOG_KNOWHERE_WARNING_ << "faiss inner error: " << e.what();
+            return Status::faiss_inner_error;
+        }
+
+        return Status::success;
+    }
+
+    Status
+    AddInternal(const DataSetPtr dataset, const Config&) override {
+        if (isIndexEmpty()) {
+            LOG_KNOWHERE_ERROR_ << "Can not add data to an empty index.";
+            return Status::empty_index;
+        }
+        if (tmp_index_rabitq.size() != indexes.size() || tmp_index_rabitq.empty() || tmp_index_rabitq[0] == nullptr) {
+            LOG_KNOWHERE_ERROR_ << "HNSW_RABITQ is immutable after its initial Add";
+            return Status::not_implemented;
+        }
+
+        const auto& scalar_info_map =
+            dataset->Get<std::unordered_map<int64_t, std::vector<std::vector<uint32_t>>>>(meta::SCALAR_INFO);
+        if (!scalar_info_map.empty()) {
+            LOG_KNOWHERE_ERROR_ << "HNSW_RABITQ does not support building with scalar info";
+            return Status::invalid_args;
+        }
+
+        try {
+            LOG_KNOWHERE_INFO_ << "Adding " << dataset->GetRows() << " rows to exact HNSW graph";
+            auto status = add_to_index(indexes[0].get(), dataset, data_format);
+            if (status != Status::success) {
+                return status;
+            }
+
+            LOG_KNOWHERE_INFO_ << "Adding " << dataset->GetRows() << " rows to RaBitQ storage";
+            // IndexPreTransform materializes its transformed input. Bound that
+            // temporary allocation instead of rotating the full FP32 dataset
+            // in one call.
+            status = add_to_index(tmp_index_rabitq[0].get(), dataset, data_format, true);
+            if (status != Status::success) {
+                return status;
+            }
+
+            faiss::cppcontrib::knowhere::IndexRefine* index_refine =
+                dynamic_cast<faiss::cppcontrib::knowhere::IndexRefine*>(indexes[0].get());
+            auto* index_hnsw = index_refine != nullptr
+                                   ? dynamic_cast<faiss::cppcontrib::knowhere::IndexHNSW*>(index_refine->base_index)
+                                   : dynamic_cast<faiss::cppcontrib::knowhere::IndexHNSW*>(indexes[0].get());
+            if (index_hnsw == nullptr) {
+                LOG_KNOWHERE_ERROR_ << "HNSW_RABITQ build produced an unexpected base index";
+                return Status::invalid_index_error;
+            }
+
+            const bool is_cosine = faiss::cppcontrib::knowhere::is_cosine_index(index_hnsw->storage);
+            std::unique_ptr<faiss::cppcontrib::knowhere::IndexHNSWRaBitQ> index_hnsw_rabitq;
+            if (is_cosine) {
+                index_hnsw_rabitq = std::make_unique<faiss::cppcontrib::knowhere::IndexHNSWRaBitQCosine>();
+            } else {
+                index_hnsw_rabitq = std::make_unique<faiss::cppcontrib::knowhere::IndexHNSWRaBitQ>();
+            }
+            // C++ slicing is intentional: preserve the exact graph while
+            // changing only the runtime HNSW type and its vector storage.
+            static_cast<faiss::cppcontrib::knowhere::IndexHNSW&>(*index_hnsw_rabitq) =
+                static_cast<faiss::cppcontrib::knowhere::IndexHNSW&>(*index_hnsw);
+
+            // Validate the replacement before relinquishing either owner so a
+            // malformed storage cannot leave the exact graph half-finalized.
+            auto* flat_storage = index_hnsw->storage;
+            index_hnsw_rabitq->storage = tmp_index_rabitq[0].get();
+            index_hnsw_rabitq->own_fields = false;
+            if (is_cosine) {
+                dynamic_cast<faiss::cppcontrib::knowhere::IndexHNSWRaBitQCosine*>(index_hnsw_rabitq.get())
+                    ->validate_cosine_storage();
+            } else {
+                index_hnsw_rabitq->validate_storage();
+            }
+            index_hnsw_rabitq->own_fields = true;
+            tmp_index_rabitq[0].release();
+            index_hnsw->storage = nullptr;
+            delete flat_storage;
+
+            if (index_refine != nullptr) {
+                delete index_refine->base_index;
+                index_refine->base_index = index_hnsw_rabitq.release();
+            } else {
+                indexes[0] = std::move(index_hnsw_rabitq);
+            }
+            tmp_index_rabitq.clear();
+        } catch (const std::exception& e) {
+            LOG_KNOWHERE_WARNING_ << "faiss inner error: " << e.what();
+            return Status::faiss_inner_error;
+        }
+
+        return Status::success;
+    }
+};
+
+template <typename DataType>
+class BaseFaissRegularIndexHNSWRaBitQNodeTemplate : public BaseFaissRegularIndexHNSWRaBitQNode {
+ public:
+    BaseFaissRegularIndexHNSWRaBitQNodeTemplate(const int32_t& version, const Object& object)
+        : BaseFaissRegularIndexHNSWRaBitQNode(version, object, datatype_v<DataType>) {
+    }
+
+    static bool
+    StaticHasRawData(const knowhere::BaseConfig& config, const IndexVersion& version) {
+        const auto& hnsw_cfg = static_cast<const FaissHnswRaBitQConfig&>(config);
+        return has_lossless_refine_index(hnsw_cfg.refine, hnsw_cfg.refine_type, datatype_v<DataType>);
+    }
+};
+
 // this index trains PRQ and HNSW+FLAT separately, then constructs HNSW+PRQ
 class BaseFaissRegularIndexHNSWPRQNode : public BaseFaissRegularIndexHNSWNode {
  public:
@@ -3354,5 +3596,7 @@ KNOWHERE_SIMPLE_REGISTER_DENSE_FLOAT_ALL_GLOBAL(HNSW_PRQ, BaseFaissRegularIndexH
                                                     knowhere::feature::EMB_LIST)
 KNOWHERE_SIMPLE_REGISTER_DENSE_INT_GLOBAL(HNSW_PRQ, BaseFaissRegularIndexHNSWPRQNodeTemplate,
                                           knowhere::feature::MMAP | knowhere::feature::MV | knowhere::feature::EMB_LIST)
+KNOWHERE_SIMPLE_REGISTER_DENSE_FLOAT_ALL_GLOBAL(HNSW_RABITQ, BaseFaissRegularIndexHNSWRaBitQNodeTemplate,
+                                                knowhere::feature::NONE)
 
 }  // namespace knowhere
--- a/src/index/hnsw/faiss_hnsw_config.h
+++ b/src/index/hnsw/faiss_hnsw_config.h
@@ -195,6 +195,60 @@ class FaissHnswPqConfig : public FaissHnswConfig {
     }
 };
 
+class FaissHnswRaBitQConfig : public FaissHnswConfig {
+ public:
+    // Number of bits per database vector dimension.
+    CFG_INT rbq_bits;
+    // Declared to reject an IVF_RABITQ query-bit parameter explicitly instead
+    // of silently ignoring the unknown JSON key. HNSW_RABITQ V1 uses FP32
+    // queries only.
+    CFG_INT rbq_bits_query;
+
+    KNOWHERE_DECLARE_CONFIG(FaissHnswRaBitQConfig) {
+        KNOWHERE_CONFIG_DECLARE_FIELD(rbq_bits)
+            .description("number of RaBitQ bits per database vector dimension")
+            .set_default(1)
+            .set_range(1, 8)
+            .for_train()
+            .for_static();
+        KNOWHERE_CONFIG_DECLARE_FIELD(rbq_bits_query)
+            .description("not supported on HNSW_RABITQ; must be 0 or omitted")
+            .set_default(0)
+            .set_range(0, 8)
+            .for_search()
+            .for_range_search();
+    }
+
+    Status
+    CheckAndAdjust(PARAM_TYPE param_type, std::string* err_msg) override {
+        const auto base_status = FaissHnswConfig::CheckAndAdjust(param_type, err_msg);
+        if (base_status != Status::success) {
+            return base_status;
+        }
+
+        const auto metric = str_to_lower(metric_type.value_or(knowhere::metric::L2));
+        if (metric != "l2" && metric != "ip" && metric != "cosine") {
+            return HandleError(err_msg, "HNSW_RABITQ only supports L2, IP and COSINE metrics",
+                               Status::invalid_metric_type);
+        }
+
+        if ((param_type == PARAM_TYPE::SEARCH || param_type == PARAM_TYPE::RANGE_SEARCH) &&
+            rbq_bits_query.value_or(0) != 0) {
+            return HandleError(err_msg, "HNSW_RABITQ only supports FP32 queries (rbq_bits_query must be 0 or omitted)",
+                               Status::invalid_args);
+        }
+
+        if (param_type == PARAM_TYPE::TRAIN && refine_type.has_value() &&
+            !WhetherAcceptableRefineType(refine_type.value())) {
+            return HandleError(err_msg,
+                               "invalid refine type : " + refine_type.value() +
+                                   ", optional types are [sq4u, sq6, sq8, fp16, bf16, fp32, flat]",
+                               Status::invalid_args);
+        }
+        return Status::success;
+    }
+};
+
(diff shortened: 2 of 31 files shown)


thirdparty/faiss/tests/test_rabitq_simd.cpp
```diff
@@ -377,3 +379,158 @@ TEST(RaBitQBitwiseAndDotProductWithPopcount, Avx2MatchesScalar) {
...
+    std::mt19937 rng(91827);
+    std::uniform_real_distribution<float> value_dist(-2.0f, 2.0f);
+    for (size_t d : kDims) {
+        const auto sign_bits = random_bytes((d + 7) / 8, 71237 + d);
+        std::vector<float> query(d);
+        for (float& value : query) {
+            value = value_dist(rng);
+        }
+        for (size_t extra_bits = 1; extra_bits <= 7; extra_bits++) {
+            const auto extra_code = random_bytes(
```
liliu-z, line 414: The new test sizes its ex_code buffer as exactly (d*ex_bits+7)/8 bytes, but the RaBitQ kernels load extended codes via 8-byte `memcpy` into a uint64_t, so every read of the final group touches up to 8 bytes past the allocation. All 84 combinations of the 12 `kDims` values x `extra_bits` 1..7 overflow: 78 combinations overflow by 7 bytes and 6 by 6 bytes. At d=1/ex_bits=1 the test allocates 1 byte and the kernel reads [0,8); at d=2048/ex_bits=7 it allocates 1792 bytes and the last read touches [1791,1799). This is not limited to the scalar `SIMDLevel::NONE` reference oracle - the vectorized path at rabitq_avx2.cpp:798 uses the same construct, so the AVX2/AVX512 test paths read out of bounds as well. Production code is unaffected: RaBitQuantizer.cpp lays out the code as [ex_code][ex_factors: 8 bytes], so the 8-byte tail always exists, and the existing bench explicitly over-allocates by 8. The kernel therefore carries an unwritten "trailing 8 bytes must be readable" contract that only the new test violates, so the fix belongs in the test, not the kernel. Under ASAN/valgrind this surfaces as a heap-buffer-overflow; the trigger surface is limited to standalone faiss builds because tests/CMakeLists.txt only adds this test under FAISS_OPT_LEVEL=dd on x86_64.

thirdparty/faiss/benchs/bench_rabitq_simd.cpp
```diff
@@ -149,7 +192,27 @@ BENCHMARK(bench_rabitq_rearrange_scalar)
...
+        ->ArgsProduct({scorer_dims})
+        ->ArgNames({"d"});
+BENCHMARK_TEMPLATE(bench_rabitq_selected_float_sum, SIMDLevel::AVX512)
+        ->ArgsProduct({scorer_dims})
+        ->ArgNames({"d"});
+BENCHMARK_TEMPLATE(bench_rabitq_multibit_inner_product, SIMDLevel::NONE)
+        ->ArgsProduct({{1, 3, 7}, scorer_dims})
+        ->ArgNames({"extra_bits", "d"});
+BENCHMARK_TEMPLATE(bench_rabitq_multibit_inner_product, SIMDLevel::AVX2)
+        ->ArgsProduct({{1, 3, 7}, scorer_dims})
```
liliu-z, line 231: `BENCHMARK_MAIN();` is placed before the closing `} // namespace faiss`, so the macro expands to `faiss::main` rather than `::main`. The resulting executable has no entry point and fails at link time with an undefined reference to `main`. This is not a regression: `git show HEAD~1` confirms the macro was already inside the namespace, and `git grep bench_rabitq_simd HEAD~1` over all CMakeLists.txt / .cmake / BUILD / Makefile files returns zero hits, meaning the file was dead source never compiled by any build system before. This PR wires it into the build for the first time, so the newly introduced target is misconfigured from the start. Blast radius is standalone-faiss only: knowhere's own `add_subdirectory` calls cover only cmake/libs/cardinal/v1|v2, tests/ut, benchmark and tests/faiss, so none of the five CMakeLists under thirdparty/faiss are processed by the knowhere build.

src/index/hnsw/impl/IndexBruteForceWrapper.cc
```diff
@@ -56,7 +57,7 @@ IndexBruteForceWrapper::search(faiss::idx_t n, const float* __restrict x, faiss:
                                const faiss::SearchParameters* __restrict params) const {
     FAISS_THROW_IF_NOT(k > 0);
 
-    std::unique_ptr<faiss::DistanceComputer> dis(index->get_distance_computer());
+    std::unique_ptr<faiss::DistanceComputer> dis(create_hnsw_native_distance_computer(index, params));
```
alexanderguzhva: I'm not sure I follow

src/index/hnsw/impl/IndexBruteForceWrapper.cc
```diff
@@ -116,7 +117,7 @@ IndexBruteForceWrapper::range_search(faiss::idx_t n, const float* x, float radiu
     RH_min bres_min(result, radius);
     RH_max bres_max(result, radius);
 
-    std::unique_ptr<faiss::DistanceComputer> dis(index->get_distance_computer());
+    std::unique_ptr<faiss::DistanceComputer> dis(create_hnsw_native_distance_computer(index, params));
```
alexanderguzhva: same

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -2851,12 +2851,27 @@ std::unique_ptr<Index> read_index_up(IOReader* f, int io_flags) {
         // rabitq.nb_bits is already set to 1 by read_RaBitQuantizer
         idxq->code_size = idxq->rabitq.code_size;
         idx = std::move(idxq);
-    } else if (h == fourcc("Ixrr")) {
-        // Ixrr = multi-bit format (new)
+    } else if (
```
alexanderguzhva: this code is expected to modify the baseline Faiss? In this case, the change needs to go into the baseline Faiss first

thirdparty/faiss/faiss/impl/RaBitQuantizer.cpp
```diff
@@ -288,13 +391,33 @@ struct RaBitQDistanceComputerNotQ final : RaBitQDistanceComputer {
...
+                dot_qo += rotated_q[i];
+            }
+        }
+        const float final_dot = query_fac.c1 * dot_qo - query_fac.c34;
+        const float pre_dist = base_fac->or_minus_c_l2sqr +
+                query_fac.qr_to_c_L2sqr -
+                2 * base_fac->dp_multiplier * final_dot;
+        return metric_type == MetricType::METRIC_L2
+                ? std::max(0.0f, pre_dist)
+                : -0.5f * (pre_dist - query_fac.qr_norm_L2sqr);
```
alexanderguzhva, line 420: as far as I remember, the Faiss uses a bit different approach for computing IP distances. Please study how they do it

thirdparty/faiss/faiss/utils/simd_impl/rabitq_avx2.cpp
```diff
@@ -471,6 +498,221 @@ namespace faiss::rabitq::multibit {
...
+                lanes[5] + lanes[6] + lanes[7] +
+                ip_byte_scalar(codes[j], query, i, d, cb);
+    }
+}
+
+namespace {
+
+#if (defined(__GNUC__) || defined(__clang__)) && \
+        (defined(__x86_64__) || defined(__i386__))
+#define FAISS_RABITQ_HAS_BMI2_TARGET 1
```
alexanderguzhva, line 704: no. 
please check how BMI2 is handled by the baseline faiss.
The reason is that intel CPUs have excellent support of BMI2 instructions, while AMD CPUs prior to Zen4 have emulated-crazy-slow BMI2 instructions execution.

thirdparty/faiss/faiss/utils/simd_impl/rabitq_avx512.cpp
```diff
@@ -721,7 +1124,488 @@ inline float ip_bitplane_avx2(
...
+        size_t ex_bits,
+        float cb) {
+    __m512 acc = _mm512_setzero_ps();
+    const __m512 v_one = _mm512_set1_ps(1.0f);
+    const __m512 v_cb = _mm512_set1_ps(cb);
+    const __m512 v_sign_weight =
+            _mm512_set1_ps(static_cast<float>(1u << ex_bits));
+
+    uint64_t pext_masks[4];
+    __m512 v_weights[4];
```
alexanderguzhva, line 1493: was https://github.com/facebookresearch/faiss/pull/5522/changes applied?

thirdparty/faiss/faiss/CMakeLists.txt
```diff
@@ -437,23 +437,14 @@ endif()
 # Export FAISS_HEADERS variable to parent scope.
 set(FAISS_HEADERS ${FAISS_HEADERS} PARENT_SCOPE)
 
-# Detect BMI2 compiler support.
```
alexanderguzhva, line 440: No.
the changes of this kind should go into `cmake/libs/libfaiss.cmake`

alexanderguzhva:
@CLiqing just in case, is this PR compatible with https://github.com/facebookresearch/faiss/pull/5526 ?

thirdparty/faiss/tests/test_rabitq_simd.cpp
CLiqing: Fixed. The standalone test now allocates `(d * extra_bits + 7) / 8 + 8`, matching the documented trailing-readable production layout. The targeted RaBitQ SIMD suites pass.

thirdparty/faiss/faiss/utils/simd_impl/rabitq_avx512.cpp
CLiqing: Yes. The AVX-512 bitplane implementation is aligned with Faiss PR #5522: 16-dimensional mask-register reconstruction, two PEXT operations per 16 dimensions, and the 8-dimensional masked tail. This PR additionally supplies batch-4 and dense-layout paths plus the runtime fast-BMI2 gate.

src/index/hnsw/impl/IndexBruteForceWrapper.cc
CLiqing: The request-local query-quantization plumbing that caused this change has been removed. V1 fixes `qb=0`.

thirdparty/faiss/faiss/utils/simd_impl/rabitq_avx2.cpp
CLiqing: I checked the current baseline Faiss implementation. It uses `FAISS_BMI2_FLAGS` / `__BMI2__` to provide a compile-time PEXT path and a non-BMI2 fallback, but I could not find an existing runtime gate for slow PEXT on AMD pre-Zen4. This patch keeps that fallback structure and follows the existing `SIMDConfig` microarchitecture-routing pattern: the PEXT functions are isolated with a BMI2 target, Intel with BMI2 uses them, AMD uses them conservatively only when AVX-512F indicates Zen4+, and older AMD CPUs take the non-BMI2 path. A forced-fallback test covers AVX2/AVX-512 single and batch-4.

thirdparty/faiss/faiss/CMakeLists.txt
CLiqing: Fixed. The baseline `thirdparty/faiss/faiss/CMakeLists.txt` is unchanged from main. Knowhere-specific source/ISA wiring is handled in `cmake/libs/libfaiss.cmake`.

thirdparty/faiss/benchs/bench_rabitq_simd.cpp
CLiqing: Fixed and verified. `BENCHMARK_MAIN()` is outside `namespace faiss`; the new `bench_rabitq_simd` target compiles and links successfully in the standalone Faiss build.

thirdparty/faiss/faiss/impl/RaBitQuantizer.cpp
CLiqing: Additional clarification after checking the current baseline Faiss implementation: the special 1-bit IP path derives the positive similarity as `-0.5 * (pre_dist - ||q||^2)`. This code is the full multi-bit path, for which Faiss `compute_full_multibit_distance` uses `q_dot_c + f_add_ex + f_rescale_ex * residual_inner_product`. The dense layout changes only how that residual inner product is decoded and scored; it does not change the estimator. Since Faiss #5226, HNSW consumes the positive similarity directly through `C_similarity`, so no extra sign conversion is required. The dense-vs-split tests cover IP for bits 2..8, single and batch-4, across aligned and unaligned dimensions.

thirdparty/faiss/faiss/impl/index_read.cpp
CLiqing: Addressed by moving the complete Knowhere-specific HNSW runtime type and its IO to `faiss/cppcontrib/knowhere`. The baseline `faiss/impl/index_read.cpp` and `index_write.cpp` no longer have changes from this PR. The reusable dense RaBitQ quantizer/SIMD enhancements remain in bundled Faiss.

CLiqing:
@alexanderguzhva It is not index-format or runtime-behavior compatible with [#5526](https://github.com/facebookresearch/faiss/pull/5526) because the construction and search contracts differ.

This PR builds the graph with exact FP32 distances and then attaches `RandomRotationMatrix -> IndexRaBitQ` storage with full multi-bit L2/IP scoring; [#5526](https://github.com/facebookresearch/faiss/pull/5526) builds directly with RaBitQ storage and uses staged L2 search.

The implementations can coexist in source: the Knowhere type is under `faiss::cppcontrib::knowhere` and uses `IHRK`, while `IHNr` remains reserved for #5526. Shared RaBitQuantizer/SIMD changes would still need normal reconciliation during a future bundled-Faiss update.

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -3032,6 +3035,204 @@ class BaseFaissRegularIndexHNSWPQNodeTemplate : public BaseFaissRegularIndexHNSW
     }
 };
 
+// Build an exact HNSW graph with Flat storage, train RaBitQ independently,
+// then replace the Flat storage after both indexes have received the data.
+// RaBitQ does not currently provide the symmetric code-to-code distance that
+// HNSW graph construction requires, so the finalized index is intentionally
+// immutable.
+class BaseFaissRegularIndexHNSWRaBitQNode : public BaseFaissRegularIndexHNSWNode {
```
liliu-z, line 3053: `BaseFaissRegularIndexHNSWRaBitQNode` (faiss_hnsw.cc:3043) inherits `BaseFaissIndexNode::IsAdditionalScalarSupported` (faiss_hnsw.cc:82-88), which returns `true` for `is_mv_only`, while `TrainInternal` (:3088) and `AddInternal` (:3159) both return `Status::invalid_args` on non-empty `SCALAR_INFO` and the index registers with `knowhere::feature::NONE` (:3558). Every other node under `BaseFaissIndexNode` registers `feature::MV`, so HNSW_RABITQ is the first type where the probe lies: a caller that queries the exported `Index::IsAdditionalScalarSupported(true)`, gets `true`, and supplies `SCALAR_INFO` hits a build failure that no retry can clear.

thirdparty/faiss/faiss/IndexRaBitQ.cpp
```diff
@@ -21,8 +21,13 @@ using rabitq_utils::SignBitFactorsWithError;
 
 IndexRaBitQ::IndexRaBitQ() = default;
 
-IndexRaBitQ::IndexRaBitQ(idx_t d_in, MetricType metric, uint8_t nb_bits_in)
-        : IndexFlatCodes(0, d_in, metric), rabitq(d_in, metric, nb_bits_in) {
+IndexRaBitQ::IndexRaBitQ(
+        idx_t d_in,
+        MetricType metric,
+        uint8_t nb_bits_in,
```
liliu-z, line 27: `IndexRaBitQ.cpp:145` hardcodes `code_size_base = (index->d + 7) / 8` to locate `SignBitFactorsWithError`, but dense-layout codes put the factors at `(d * nb_bits + 7) / 8` (`RaBitQuantizer.cpp:126-128`). An index built with the new `dense_layout=true` constructor and `qb=0` therefore reads `base_fac->f_error` out of the middle of the code bitstream, and `should_refine_candidate` prunes candidates against a garbage error bound — true neighbours are dropped with no error and no crash. The same offset is computed correctly in `RaBitQuantizer.cpp:394-396` and `:565-567`; this is the one copy that was missed.

src/index/hnsw/faiss_hnsw_config.h
```diff
@@ -195,6 +195,43 @@ class FaissHnswPqConfig : public FaissHnswConfig {
     }
 };
 
+class FaissHnswRaBitQConfig : public FaissHnswConfig {
```
liliu-z, line 198: `FaissHnswRaBitQConfig` (faiss_hnsw_config.h:198-233) declares only `rbq_bits`, while `faiss_hnsw.cc:3110` hardcodes `rabitq_index->qb = 0`. The unknown-JSON-key check in `Config::FormatAndCheck` is commented out (src/common/config.cc:66-81), so a user migrating from IVF_RABITQ who passes the documented `rbq_bits_query` key (index_param.h:250) has it dropped with no error and no log line. `IvfRaBitQFastScanConfig` faces the identical situation and declares the field specifically to reject it (ivf_config.h:447-449, 481-488).

thirdparty/faiss/faiss/IndexRaBitQ.h
```diff
@@ -37,7 +37,8 @@ struct IndexRaBitQ : IndexFlatCodes {
     explicit IndexRaBitQ(
             idx_t d,
             MetricType metric = METRIC_L2,
-            uint8_t nb_bits = 1);
+            uint8_t nb_bits = 1,
+            bool dense_layout = false);
```
liliu-z, line 41: `qb` defaults to 4 (IndexRaBitQ.h:30) and the new dense constructor (IndexRaBitQ.cpp:24-34) leaves it alone, but `RaBitQuantizer::get_distance_computer` throws `"dense-layout RaBitQ currently requires qb=0"` for any `qb != 0` (RaBitQuantizer.cpp:887-890). `IndexRaBitQ(d, metric, bits, /*dense_layout=*/true)` trains and adds successfully and then throws on the first `search()` / `get_FlatCodesDistanceComputer()`; both callers in this PR work only because they assign `qb = 0` by hand (faiss_hnsw.cc:3110, tests/ut/test_hnsw_rabitq.cc:280-281).

src/index/hnsw/faiss_hnsw.cc
```diff
@@ -3032,6 +3035,204 @@ class BaseFaissRegularIndexHNSWPQNodeTemplate : public BaseFaissRegularIndexHNSW
...
+
+        try {
+            LOG_KNOWHERE_INFO_ << "Adding " << dataset->GetRows() << " rows to exact HNSW graph";
+            auto status = add_to_index(indexes[0].get(), dataset, data_format);
+            if (status != Status::success) {
+                return status;
+            }
+
+            LOG_KNOWHERE_INFO_ << "Adding " << dataset->GetRows() << " rows to RaBitQ storage";
+            status = add_to_index(tmp_index_rabitq[0].get(), dataset, data_format);
```
liliu-z: `AddInternal` hands the whole dataset to `add_to_index(tmp_index_rabitq[0].get(), ...)` (faiss_hnsw.cc:3174); for `DataFormatEnum::fp32` that is a single `index->add(rows, ...)` (faiss_hnsw.cc:683-685) into `IndexPreTransform::add`, which allocates `new float[n * d_out]` via `apply_chain` → `VectorTransform::apply` (IndexPreTransform.cpp:121-157, VectorTransform.cpp:151-155). Build peak RSS gains a transient copy of the entire fp32 dataset that HNSW_PQ/HNSW_PRQ never pay, since `IndexPQ` encodes in place through `IndexFlatCodes::add`. fp16/bf16 are unaffected because they already take the 4096-row chunked branch at faiss_hnsw.cc:688-703.

thirdparty/faiss/perf_tests/CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ include(../cmake/link_to_faiss_lib.cmake)
 
 link_to_faiss_lib(faiss_perf_tests_utils)
 
+add_executable(bench_rabitq_simd ../benchs/bench_rabitq_simd.cpp)
```
liliu-z: `perf_tests/CMakeLists.txt:34-36` adds `bench_rabitq_simd` to the default build target with no `FAISS_OPT_LEVEL` guard and no `EXCLUDE_FROM_ALL`, and `benchs/bench_rabitq_simd.cpp:191-214` instantiates `SIMDLevel::AVX2` and `SIMDLevel::AVX512` benchmarks under only `#if defined(__x86_64__)`. `perf_tests` is added whenever `BUILD_TESTING` is on (thirdparty/faiss/CMakeLists.txt:156-161; `include(CTest)` defaults it ON), but `FAISS_SIMD_SRC` reaches the plain `faiss` target only under `FAISS_OPT_LEVEL=dd` (faiss/CMakeLists.txt:534-536) and `faiss_avx2` carries only `FAISS_SIMD_AVX2_SRC` (:466), so the default `generic` and the `avx2` configuration both fail to link. In `dd` builds it links but has no `is_simd_level_available` check, so it executes AVX-512 kernels on non-AVX-512 x86_64 hosts. Knowhere's own build never compiles this file, so CI does not catch it.

src/index/hnsw/faiss_hnsw_config.h
```diff
@@ -195,6 +195,43 @@ class FaissHnswPqConfig : public FaissHnswConfig {
...
+ public:
+    // Number of bits per database vector dimension.
+    CFG_INT rbq_bits;
+
+    KNOWHERE_DECLARE_CONFIG(FaissHnswRaBitQConfig) {
+        KNOWHERE_CONFIG_DECLARE_FIELD(rbq_bits)
+            .description("number of RaBitQ bits per database vector dimension")
+            .set_default(1)
+            .set_range(1, 9)
```
liliu-z: `rbq_bits` accepts 1..9 (faiss_hnsw_config.h:207) and anything above 1 selects the dense layout (faiss_hnsw.cc:3108), but the dense dispatch switches in `rabitq_avx2.cpp:597-608` and `rabitq_avx512.cpp:916-927` — and both `_batch_4` variants (avx2:612-638, avx512:931-957) — only cover nbits 2..8 and fall through to the per-dimension scalar `compute_inner_product_dense<SIMDLevel::NONE>`. Every graph-search distance at `rbq_bits=9` runs unvectorized on every SIMD level, and 9 is the single width excluded from the dense-vs-packed equivalence test (tests/ut/test_hnsw_rabitq.cc:276); the `{IP, 13, 9}` e2e scenario at :189 asserts only finiteness and ordering.

src/index/hnsw/faiss_hnsw.cc
CLiqing: Fixed. HNSW_RABITQ now overrides IsAdditionalScalarSupported() to return false for both probe modes, matching its feature registration and build behavior. The HNSW_RABITQ UT covers both values.

thirdparty/faiss/faiss/IndexRaBitQ.cpp
CLiqing: Fixed. The staged scan now locates SignBitFactorsWithError after the dense stream at (d * nb_bits + 7) / 8. The regression test compares packed and dense IndexRaBitQ search labels and distances across L2/IP, bits 2-8, and aligned/unaligned dimensions.

thirdparty/faiss/faiss/IndexRaBitQ.h
CLiqing: Fixed. The dense-layout IndexRaBitQ constructor now sets qb=0 itself, while the non-dense legacy default remains unchanged. The dense regression test verifies the constructor default before search.

thirdparty/faiss/perf_tests/CMakeLists.txt
CLiqing: Fixed. bench_rabitq_simd is now an EXCLUDE_FROM_ALL target only for x86 FAISS_OPT_LEVEL=dd, and fixed-ISA benchmarks check runtime SIMD availability before execution. Fresh generic and AVX2 default builds pass and do not expose this target; the DD target compiles, links, and runs.

src/index/hnsw/faiss_hnsw_config.h
CLiqing: Fixed without exposing query quantization. HNSW_RABITQ now declares rbq_bits_query only to reject nonzero values explicitly during search/range search; omitted or zero keeps the V1 FP32-query behavior.

src/index/hnsw/faiss_hnsw.cc
CLiqing: Fixed. The RaBitQ IndexPreTransform storage is now added in 4096-row chunks, bounding the rotated FP32 temporary allocation. The exact HNSW graph add path and non-FP32 conversion paths retain their existing behavior.

src/index/hnsw/faiss_hnsw_config.h
CLiqing: Fixed by narrowing the HNSW_RABITQ public range to 1-8 bits, which is the range covered by the dense SIMD single/batch-4 implementation and equivalence tests. The generic bundled IndexRaBitQ keeps its existing 1-9 bit support.

thirdparty/faiss/faiss/impl/RaBitQuantizer.h
```diff
@@ -47,10 +47,16 @@ struct RaBitQuantizer : Quantizer {
     // - nb_bits = 2-9: multi-bit RaBitQ (1 sign bit + ex_bits extra bits)
     size_t nb_bits = 1;
 
+    // Store one complete scalar code per dimension in a dense n-bit stream.
+    // This combines the sign bit and extra bits at build time while retaining
+    // the exact nbits/dimension budget. For nbits=8 this is one byte/dimension.
+    bool dense_layout = false;
```
liliu-z, line 53: `RaBitQuantizer::dense_layout` (RaBitQuantizer.h:53) changes the on-disk code layout but is not part of the standard wire format: `index_write.cpp:1027` tags every multi-bit `IndexRaBitQ` as `Ixrr` and `index_read.cpp:2856` default-constructs the index, so a `faiss::write_index`/`read_index` round-trip restores `dense_layout=false` and reinterprets dense codes with packed offsets. `read_RaBitQuantizer` does not recompute `code_size`, and for any `d % 8 == 0` (64, 128, 768, 1536) the dense and packed sizes are byte-identical, so nothing rejects the file — every distance is silently wrong. `IndexRaBitQFastScan::IndexRaBitQFastScan(const IndexRaBitQ&, int)` (IndexRaBitQFastScan.cpp:106) copies the quantizer and then reads sign bits at `orig_code[j/8] >> (j%8)`, which is garbage for dense codes, also without a guard.

tests/ut/test_hnsw_rabitq.cc
```diff
@@ -0,0 +1,385 @@
...
+        const auto query = GenDataSet(kNq, scenario.dim, 401 + scenario.dim + scenario.rbq_bits);
+        auto json = MakeHnswRaBitQConfig(scenario.dim, scenario.metric, scenario.rbq_bits);
+
+        auto index = knowhere::IndexFactory::Instance()
+                         .Create<knowhere::fp32>(knowhere::IndexEnum::INDEX_HNSW_RABITQ, version)
+                         .value();
+        REQUIRE(index.Build(train, json) == knowhere::Status::success);
+        REQUIRE(index.Count() == kNb);
+
+        const auto before = index.Search(query, json, nullptr);
```
liliu-z, line 269: Every assertion in `test_hnsw_rabitq.cc` is self-referential: `CheckValidKnnResult` (:65) only checks id range and `isfinite`, `CheckKnnOrder` (:78) only checks monotonicity, the serialization round-trip (:274) compares the index against itself, `CalcDistByIDs` (:253) compares two paths that resolve to the same distance computer, and the dense-vs-packed case (:282) compares two layouts of the same algorithm. If the query were rotated with a different matrix than the base vectors, or the shared RaBitQ distance formula had a sign error, all five `TEST_CASE`s still pass and the index ships returning wrong neighbors. `HNSW_RABITQ` appears in no other test file, so this suite is the only coverage.

thirdparty/faiss/tests/test_rabitq_simd.cpp
```diff
@@ -377,3 +379,175 @@ TEST(RaBitQBitwiseAndDotProductWithPopcount, Avx2MatchesScalar) {
         }
     }
 }
+
+template <SIMDLevel SL>
```
liliu-z, line 383: `tests/faiss/CMakeLists.txt` builds `faiss_tests` from a hardcoded `FAISS_TEST_SRCS` list that does not contain `test_rabitq_simd.cpp`, and this PR does not add it, so the 174 lines of new AVX2/AVX-512 comparison tests never run in knowhere CI. The only remaining coverage for the ~1400 new kernel lines is the dense-vs-packed case in `test_hnsw_rabitq.cc:282`, which exercises exactly one runtime-detected SIMD level: on an AVX-512 CI host the AVX2 dense kernels are never executed, yet they are the ones that run on AVX2-only production hosts, where a kernel bug produces wrong distances undetected.

thirdparty/faiss/faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.h>
+
+#include <faiss/VectorTransform.h>
+#include <faiss/impl/FaissAssert.h>
+
+namespace faiss::cppcontrib::knowhere {
+
+IndexHNSWRaBitQ::IndexHNSWRaBitQ() = default;
+
+IndexHNSWRaBitQ::IndexHNSWRaBitQ(faiss::IndexPreTransform* storage_in, int M)
```
liliu-z: `IndexHNSWRaBitQ(IndexPreTransform*, int)` delegates to `IndexHNSW(storage, M)` (IndexHNSW.cpp:241), which only builds an empty `hnsw(M)`; `entry_point` stays `-1` so `HNSW.cpp:1006` returns immediately and every search yields no results, while the constructor still sets `ntotal` from the storage and presents the index as populated — and `add()` is overridden to throw, so the graph can never be filled in. It also sets `own_fields = true` at line 19 before calling `validate_storage()` at line 22: if validation throws, `~IndexHNSW` runs and deletes the storage the caller still owns, giving a double free. Neither this constructor nor the non-const `pretransform_index()`/`rabitq_index()` overloads have any caller.

thirdparty/faiss/faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <faiss/IndexPreTransform.h>
+#include <faiss/IndexRaBitQ.h>
+#include <faiss/cppcontrib/knowhere/IndexHNSW.h>
+
+namespace faiss::cppcontrib::knowhere {
+
+// Private Knowhere serialization tag. Upstream Faiss reserves "IHNr" for
+// its incompatible direct-build/staged-search IndexHNSWRaBitQ format.
+inline constexpr char kHnswRaBitQFourcc[] = "IHRK";
```
alexanderguzhva, line 21: this should not be here. It needs to be in `index_read.cpp` and `index_write.cpp`. Yes, let it be duplicate. Also, please make sure that our four-CC codes do not intersect with Faiss ones. So, a given four-CC should not be used by Faiss.

alexanderguzhva:
@CLiqing Please create a corresponding PR in the baseline faiss code and we'll ask Faiss guys to take a look. Thanks.

thirdparty/faiss/faiss/impl/RaBitQuantizer.h
CLiqing: Fixed. Dense multi-bit `IndexRaBitQ` now uses the distinct `Ixrd` tag; the standard reader restores `dense_layout`, recomputes and validates `code_size`, and validates the code buffer length. Converting a dense index to `IndexRaBitQFastScan` now throws explicitly. Cross-reader round-trip tests cover the standard and cppcontrib readers.

tests/ut/test_hnsw_rabitq.cc
CLiqing: Fixed. The test now computes an independent FP32 brute-force oracle and checks recall for L2/IP, RBQ1/RBQ4/RBQ8, and aligned/unaligned dimensions. It also verifies COSINE storage/refine distances against original vectors, so rotation, sign, or metric mistakes are no longer self-validating.

thirdparty/faiss/tests/test_rabitq_simd.cpp
CLiqing: Fixed. `test_rabitq_simd.cpp` is now listed in `tests/faiss/CMakeLists.txt`. I built `faiss_tests` locally and ran 10 RaBitQ tests, including explicit AVX2/AVX-512 single and batch-4 equivalence and the non-BMI2 fallback.

thirdparty/faiss/faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.cpp
CLiqing: Fixed. The unused pointer constructor and the unused non-const accessors were removed, eliminating both the empty-graph behavior and the exception-path ownership hazard.

thirdparty/faiss/faiss/cppcontrib/knowhere/IndexHNSWRaBitQ.h
CLiqing: Fixed. Knowhere HNSW RaBitQ serialization is now implemented in the standard Faiss `index_read.cpp`/`index_write.cpp`, while the cppcontrib implementation remains duplicated as requested. The tags are `IHRK`, `IHRC`, and `IRKC`; I checked them against the bundled/upstream Faiss tag set. Cross-reader tests cover standard Faiss and cppcontrib IO.

CLiqing:
Created the requested baseline Faiss PR: https://github.com/facebookresearch/faiss/pull/5552

It contains the reusable dense multi-bit RaBitQ layout, serialization, SIMD single/batch-4 scorers, BMI2 runtime fallback, and focused tests. The Knowhere-specific HNSW integration remains in this PR.

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -1305,6 +1306,19 @@ static void read_HNSW(HNSW& hnsw, IOReader* f) {
...
+    READVECTOR(hnsw.cum_nneighbor_per_level);
+    READVECTOR(hnsw.levels);
+    READVECTOR(hnsw.offsets);
+    read_vector(hnsw.neighbors, f);
+    READ1(hnsw.entry_point);
+    READ1(hnsw.max_level);
+    READ1(hnsw.efConstruction);
+    READ1(hnsw.efSearch);
+    READ1(hnsw.upper_beam);
+}
```
liliu-z, line 1320: `read_knowhere_HNSW` (index_read.cpp:1309-1320) omits the `validate_HNSW(hnsw)` call that `read_HNSW` performs at :1306, and the IHRK/IHRC branch at :2552 only calls `validate_storage()`/`validate_cosine_storage()`, neither of which inspects `hnsw.levels`, `hnsw.offsets`, `hnsw.neighbors` or `hnsw.entry_point`. This is the only HNSW-graph-carrying fourcc in `read_index_up` that bypasses `validate_HNSW` — the branches at :2611, :3464 and :3488 all go through `read_HNSW`. Take a valid HNSW_RABITQ index serialized by knowhere (fourcc IHRK), set one entry of `neighbors[]` to `ntotal`, and load it through `faiss::read_index` — reachable from `src/index/faiss/faiss.cc:345` (BinarySet bytes) and :365/:368 (file), and exercised directly by `tests/ut/test_hnsw_rabitq.cc:323`. Deserialization succeeds, then the first search reads out of bounds, because `HNSW::neighbor_range` (cppcontrib/knowhere/impl/HNSW.cpp:53-58) indexes `offsets[]`/`neighbors[]` with no bounds check.

tests/faiss/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(FAISS_TEST_SRCS
   ../../thirdparty/faiss/tests/test_pq_code_distance.cpp
   ../../thirdparty/faiss/tests/test_cppcontrib_uintreader.cpp
   ../../thirdparty/faiss/tests/test_distances_simd.cpp
+  ../../thirdparty/faiss/tests/test_rabitq_simd.cpp
```
liliu-z, line 10: `tests/faiss/CMakeLists.txt:10` adds `test_rabitq_simd.cpp` to `FAISS_TEST_SRCS`, which feeds both the `if(__X86_64)` target at :38 and the `if(__AARCH64)` target at :61. The file unconditionally instantiates AVX2/AVX512 explicit specializations (:502, :509, :516, :524-529, :539, :546, :552 are new in this PR) whose definitions live only in `rabitq_avx2.cpp`/`rabitq_avx512.cpp`, which `cmake/libs/libfaiss.cmake:75` and :124 strip from `FAISS_SRCS` on aarch64. Building on aarch64 with `-DWITH_FAISS_TESTS=ON` fails to link `faiss_tests` with `undefined reference to selected_float_sum<(SIMDLevel)1>`, `multibit::compute_inner_product<(SIMDLevel)2>`, `multibit::compute_inner_product_batch_4<(SIMDLevel)1>`, `rearrange_bit_planes<(SIMDLevel)1>` and `bitwise_and_dot_product_with_popcount<(SIMDLevel)1>`. Reproduced on aarch64 with g++ 11.4.0. `WITH_FAISS_TESTS` defaults to OFF (CMakeLists.txt:42), so CI does not catch it.

thirdparty/faiss/faiss/impl/index_read.cpp
```diff
@@ -2351,6 +2375,24 @@ std::unique_ptr<Index> read_index_up(IOReader* f, int io_flags) {
...
+    } else if (
+            h ==
+            fourcc(cppcontrib::knowhere::kRaBitQPreTransformCosineFourcc)) {
+        auto ixpt = std::make_unique<
+                cppcontrib::knowhere::IndexPreTransformRaBitQCosine>();
+        ixpt->own_fields = true;
+        read_index_header(*ixpt, f);
+        int nt;
+        READ1(nt);
+        FAISS_THROW_IF_NOT_MSG(
```
liliu-z, line 2387: The new IRKC branch at index_read.cpp:2385-2392 reads the transform count with only an `nt >= 0` check, while the `IxPT` branch it was copied from adds `FAISS_CHECK_DESERIALIZATION_LOOP_LIMIT` at :2410 and validates `chain[0]->d_in == d`, `chain[i]->d_in == chain[i-1]->d_out` and `chain[nt-1]->d_out == index->d` at :2421-2443. `validate_norms()` at :2394 only checks inverse-norm count and positivity, so a top-level IRKC index gets no chain validation at all. Load a crafted IRKC index whose single chain transform has `d_in != d` through `faiss::read_index` (reachable from src/index/faiss/faiss.cc:345). Deserialization succeeds, then `IndexPreTransform::apply_chain` transforms with mismatched dimensions during search. A large `nt` additionally drives unbounded `chain` growth that the loop-limit macro — used 13 times elsewhere in this same file — exists to bound.

Issue #1794, opened 2026-08-26
[Bug]: EmbListOffset(lims, rows) drops trailing empty lists, producing a short offset meta


foxspy:
### Summary

`EmbListOffset(const size_t* lims, size_t rows)` infers the end of `lims` by scanning for the first element equal to `rows`. Empty lists do not advance the cumulative offset, so **a run of empty lists at the tail of a segment appears in `lims` as a run of repeated values equal to `rows`** — and the loop swallows all of them, pushing back only one.

The resulting `EmbListOffset` is short by exactly the number of trailing empty lists, which makes `num_el()` (and therefore `ExternalCount()`, added in #1793) smaller than the segment's real row count.

https://github.com/zilliztech/knowhere/blob/branch_v305/include/knowhere/emb_list_utils.h#L30-L41

```cpp
EmbListOffset(const size_t* lims, size_t rows) {
    size_t idx = 0;
    assert(lims[idx] == 0);
    assert(rows > 0);
    while (lims[idx] < rows) {        // <-- infers array length from a sentinel value
        assert(idx == 0 || lims[idx] >= lims[idx - 1]);
        offset.push_back(lims[idx]);
        idx++;
    }
    assert(lims[idx] == rows);        // always true: the loop stops at the first match
    offset.push_back(lims[idx]);      // <-- only one entry is pushed back
}
```

The build path reaches it here:

- `include/knowhere/index/index_node.h:424` — `BuildEmbList(dataset, cfg, lims, dataset->GetRows(), ...)`; the 4th argument is the **flattened vector count**, used as the sentinel.
- `src/index/index_node.cc:402` — `EmbListOffset doc_offset(lims, num_rows);`

### Worked example

8 rows; rows 0–2 hold 2/3/2 vectors, rows 3–7 are empty lists. Total flattened vectors `rows = 7`.

```
lims (written by the caller, n+1 = 9 entries):
  [0, 2, 5, 7, 7, 7, 7, 7, 7]
        ^-- correct num_el = 8

loop: 0<7 push, 2<7 push, 5<7 push, 7<7 false -> stop at idx 3
      then push lims[3] = 7

offset = [0, 2, 5, 7]   ->  num_el() = 3     (expected 8, lost 5)
```

[...]

PR #1801, opened 2026-09-01
fix: preserve trailing empty embedding lists

marcelo-cjl:
issue: #1794

## Summary
- Add explicit `EMB_LIST_COUNT` metadata so embedding-list offsets are consumed with their real list count instead of inferring array length from the flattened vector count.
- Remove the unsafe two-argument `EmbListOffset` constructor, thread the explicit count through build, add, search, brute-force, conversion, SWIG, and retrieval paths, and validate offset boundaries in release builds.
- Preserve trailing empty embedding lists while keeping `NQ` semantics separate, and add focused regression coverage for trailing-empty offsets and retrieved metadata.

## Test Plan
- [x] `make WITH_UT=True BUILD_DIR="$PWD/build/main/ut"`
- [x] `source build/main/ut/Release/generators/conanrun.sh && build/main/ut/Release/tests/ut/knowhere_tests "[trailing_empty]"` (3 assertions passed)
- [x] `source build/main/ut/Release/generators/conanrun.sh && build/main/ut/Release/tests/ut/knowhere_tests "[GetEmbListByIds],[empty_emb_list],[emb_list]"` (2171 assertions passed; one capability-based skip)
- [x] `pre-commit run --files <touched files>`
- [ ] Full Knowhere unit-test suite (not run; validation was scoped to the affected EmbList paths)

### DIFF
--- a/include/knowhere/comp/index_param.h
+++ b/include/knowhere/comp/index_param.h
@@ -125,6 +125,7 @@ constexpr const char* DIM_MAX_SCORE_RATIO = "dim_max_score_ratio";
 // emb list meta
 constexpr const char* EMB_LIST_META = "EMB_LIST_META";
 constexpr const char* EMB_LIST_OFFSET = "EMB_LIST_OFFSET";
+constexpr const char* EMB_LIST_COUNT = "EMB_LIST_COUNT";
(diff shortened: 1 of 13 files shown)


foxspy (approved):
/lgtm

Issue #1802, opened 2026-09-01
[Bug]: PrepareBitset conflates vec and list ID domains


marcelo-cjl:
### Summary

Knowhere has two filter ID domains for embedding-list data:

- `VEC`: one bit per searchable vector
- `LIST`: one bit per embedding list

`BitsetView` currently does not identify which domain its bits belong to. `IdMapData` also does not carry a domain, while `IdMap` exposes one shared external count and `PrepareBitsetMap` selects a mapping from the maps that happen to be present. As a result, a vector-domain bitset can be validated or projected with list-domain metadata.

### Current behavior

`Index<T>::Search` calls `PrepareBitset` before the EmbList search dispatch. At that point:

- `PrepareBitset` validates `bitset.num_bits()` against the single `IdMap::OutCount()`.
- `PrepareBitsetMap` prefers the EmbList map when present and otherwise falls back to the default map.
- Neither decision is driven by the bitset's intended `VEC` or `LIST` domain.

A minimal conflicting state is:

```text
backend vector count: 1024
embedding-list count: 2
bitset domain: VEC
bitset size: 1024
IdMap external count selected by PrepareBitset: 2
```

The vector-domain bitset is then rejected as larger than the list-domain external count, even though it is correctly sized for the searchable vectors.

### Expected behavior

The bitset and ID mapping should use the same explicit domain:

```cpp
enum class IdDomain {
    VEC,
    LIST,
};
```

- `BitsetView` carries its `IdDomain`, including when the bitmap is empty.
- `IdMap` exposes the count and internal-to-external mapping for a requested domain.
- `PrepareBitset` validates and projects using only the mapping matching `bitset.id_domain()`.
- A `VEC` identity mapping may be represented without allocating a dense identity array, but it must still retain its domain and count.
- Missing or incompatible domain mappings return an explicit error instead of falling back to another domain.

Conceptually:

```cpp
PrepareBitset(BitsetView bitset) {
    const auto domain = bitset.id_domain();
    const auto count = id_map.OutCount(domain);
    const auto& out_ids = id_map.InToOutIds(domain);

    ValidateBitsetSize(bitset, count);
    ProjectBitset(bitset, out_ids);
}
```

### Affected areas

- `include/knowhere/bitsetview.h`
- `include/knowhere/dataset.h` (`IdMapData`)
- `include/knowhere/id_map.h`
- `include/knowhere/index/index_node.h` (`PrepareBitset` / `PrepareBitsetMap`)
[...]

PR #1803, opened 2026-09-01
fix: normalize EmbList element id maps

marcelo-cjl:
issue: #1802

## Summary
- Normalize outer/list nullable metadata to a contiguous vector IdMap when ordinary metrics build an element index over EmbList input.
- Preserve the normal IdMap handoff so Cardinal can compose its internal relayout with the public vector ID space instead of dropping the mapping.
- Keep EBL metric/list-domain mapping unchanged, and cover sealed, filtered, all-null, and growing element-index paths.

## Test Plan
- [x] `PATH=/usr/lib/llvm-15/bin:$PATH /home/ubuntu/.local/bin/pre-commit run --files src/index/index.cc tests/ut/test_nullable_id_map.cc`
- [x] `cmake --build artifacts/build-ut/Release --target knowhere_tests`
- [x] `source artifacts/build-ut/Release/generators/conanrun.sh && artifacts/build-ut/Release/tests/ut/knowhere_tests '[nullable][id_map]'` (559 assertions in 7 cases)
- [x] `cmake --build artifacts/build-cardinal-ut/Release --target knowhere_tests`
- [x] `source artifacts/build-cardinal-ut/Release/generators/conanrun.sh && artifacts/build-cardinal-ut/Release/tests/ut/knowhere_tests '[nullable][id_map]'` (1984 assertions in 9 cases)
- [x] Native and Cardinal-enabled `Nullable IVF_FLAT_CC emb-list Add preserves append public list ids` (129 assertions each)
- [ ] Milvus end-to-end validation (not run; this PR is validated at the Knowhere component and Cardinal integration tiers)

### DIFF
--- a/src/index/index.cc
+++ b/src/index/index.cc
@@ -11,7 +11,11 @@
 
 #include "knowhere/index/index.h"
 
+#include <cstdint>
+#include <limits>
+#include <numeric>
 #include <string>
+#include <vector>
 
 #include "folly/futures/Future.h"
 #include "knowhere/comp/time_recorder.h"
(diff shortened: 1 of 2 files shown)


foxspy (approved):
/lgtm

PR #1817, opened 2026-09-12
enhance: Do not store raw vectors in SVS Vamana

mihaic:
Knowhere does not use the stored vectors in SVS Vamana. Those vectors are currently just overhead.


foxspy (approved):
/lgtm

alexanderguzhva:
issue: #1822
/kind improvement

Issue #1816, opened 2026-09-10
[Bug] HNSW top-k search skips brute-force fallback with an empty bitset


zhuwenxing:
## Summary

HNSW top-k search can return fewer than `k` results with an empty filter bitset even when sufficient indexed vectors are available and `disable_fallback_brute_force=false`.

The insufficient-results fallback in `BaseFaissRegularIndexHNSWNode::Search` uses `bitset.size() - bitset.count()` as the available-vector count. For a default-constructed bitset, that expression is zero, although an empty bitset means no filtering. The fallback is therefore never invoked.

Reproduced with HNSW_PQ through Milvus. Adding an all-pass scalar filter to the same query and index restores the missing results; explicitly disabling fallback then reproduces the shortage again.

## Environment

- Knowhere: `2868882b02fd31fb658e276b20ad4309446e8173`.
- Milvus server: `1759fb0961d`, verified using `get_server_version(detail=True)`; standalone, Linux/amd64, Cardinal-enabled build. Server logs confirm that the affected index is actually built and loaded as Knowhere `HNSW_PQ`.
- PyMilvus: `3.1.0rc83`; Python 3.12; NumPy 2.5.3.
- `autoIndex.enable=false`; vector index version 11; scalar index version 5; storage version 3.
- 3000 indexed float32 vectors, dimension 128, COSINE, no deletes or nullable vectors.
- Build parameters: `M=2, efConstruction=1, m=1, nbits=1, refine=true, refine_type=SQ8`.
- The faulty condition is also present in current main at `67fe228222cdad2eaf0cb0c9cf67c18620b24db3` (source inspection only; runtime results below are from the affected version above).

## Observed behavior

The original Milvus test, `TestHnswPQBuildParams::test_hnsw_pq_build_params[params63]`, searches with `nq=2, k=10, ef=64, refine_k=1` and no filter. Repeating the unchanged test 10 times on the affected server produced **1 failure and 9 passes**. The failure returned **8 results instead of 10**. The original [Jenkins build 35](http://jenkins.milvus-cell.tc-op.zilliz.cn/job/cardinal-e2e-master/35/allure/#) returned 7 and 8 results across the initial attempt and retry.

To compare fallback behavior without rebuilding the graph between requests, a separate experiment used 100 queries on one fixed 3000-vector index. Every inserted primary key satisfies `id >= 0`.

| k | Filter | disable_fallback_brute_force | Results per query |
|---|---|---|---|
[...]

PR #1824, opened 2026-09-14
Fixes 1816 (HNSW top-k search skips brute-force fallback with an empty bitset)

alexanderguzhva:
issue: #1816 
/kind bug

Generated using Codex.

---

The patch fixes HNSW top-k searches returning fewer than k results because brute-force fallback is incorrectly skipped when no filter is supplied.

For example, a query requests 50 neighbors from 3,000 indexed vectors. Graph traversal finds only one candidate. With fallback enabled, Knowhere should scan the stored vectors to fill the results.

Previously, the fallback condition included:
```C++
  real_topk < bitset.size() - bitset.count()
```

That expression estimates how many eligible vectors exist. But an absent filter is represented by a default-constructed bitset whose size and count are both zero. The comparison therefore becomes:
```C++
  1 < 0  // false: fallback is skipped
```

An empty bitset means all indexed vectors are eligible, so zero is the wrong population count.

The patch computes the eligible population correctly:
```C++
  const size_t eligible_count =
      bitset.empty()
          ? static_cast<size_t>(indexes[index_id]->ntotal)
          : bitset.size() - bitset.count();
```
The fallback condition then uses `real_topk < eligible_count`. It still requires all four conditions:
  - Graph search returned fewer than k valid results.
  - More eligible vectors exist than were returned.
  - A brute-force wrapper is available.
  - `disable_fallback_brute_force` is false.

For filtered searches, the prepared filter count remains in use, preserving visibility boundaries and ID mappings. For unfiltered searches, the count comes from the selected index, which matters when there are multiple sub-indexes.

The regression demonstrates the behavior through Knowhere’s public search API: a graph with disconnected links returns one hit before the fix; afterward, enabled fallback returns the requested 10 or 50 hits. The graph itself remains disconnected - the fix makes the configured fallback handle its insufficient results.


foxspy (approved):
/lgtm

Issue #1820, opened 2026-09-14
Extend clustering APIs for compaction planning and shared search hints


foxspy:
## Description

Clustering consumers need to pass assignment options, import previously computed centroids, and request a compaction plan through the common Cluster API. Query producers and search implementations also need a shared type for per-query centroid row ranges and distances carried in DataSet metadata.

## Expected Behavior

Expose optional clustering operations with configuration validation, retain the existing assignment API for implementations that do not override the new methods, and provide a shared SearchHint definition. Unsupported operations should return not_implemented, and facade methods should contain implementation exceptions.

## Actual Behavior

The Cluster facade exposes training, assignment without options, and centroid retrieval. It does not expose centroid import or compaction planning, and there is no shared header defining the per-query search hint payload.

PR #1826, opened 2026-09-15
feat: add typed compaction results and owned search hints

foxspy:
issue: https://github.com/zilliztech/knowhere/issues/1820

## Summary

- Replace the DataSet-based compaction-planning interface with centroid counts and an owned `CompactionResult`: `row_count`, `centroid_count`, `centroid_counts`, and typed `centroid_groups`. Planner diagnostics belong in logs, not in the execution result.
- Add explicit `AssignWithDistance` while preserving legacy assignment behavior. Validate options and contain exceptions at the facade; unsupported extensions return `not_implemented`.
- Consolidate request-owned DataSet search hints in `include/knowhere/comp/search_hint.h` under the single `search_hints` metadata key. Each range carries signed 64-bit `offset`/`count` and `float id_distance`, with distinct absent/empty semantics. Searchers decide how to consume hints; there is no public mode parameter.
- Add an optional 64-bit METIS dependency and API regression coverage.

## API contract

This updates the interface introduced by PR #1821. `BuildCompactionPlan` callers and overrides must use the typed counts/result contract and rebuild together. Existing assignment calls remain supported.

There is one search-hint contract, with no compatibility adapter or alternate key. The element type is `SearchHint { int64_t offset; int64_t count; float id_distance; }`. `BatchSearchHints` is a vector of optional per-query hint lists; DataSet stores a `std::shared_ptr<const BatchSearchHints>` under `kSearchHintsField`. Producers use `SetSearchHints`, consumers use `GetSearchHints`, and `ClearSearchHints` removes that same payload. The duplicate top-level header is removed; there is no alternate element type or legacy alias.

`id_distance` is a finite query-to-centroid distance (default `0.0f`), not an individual row's query distance. Smaller values prioritize graph entry points, with stable ordering for ties; actual entry/result distances are computed by the index. BF uses only the range union and ignores this priority. No external mode selector is introduced.

```cpp
dataset->SetSearchHints({
    knowhere::QuerySearchHints{{{100, 20, 0.5f}, {200, 10, 1.5f}}},
    std::nullopt,                 // ordinary search for this query
    knowhere::QuerySearchHints{}  // explicit empty candidate hint
});
```

## Validation

- Changed-file pre-commit checks passed.
- Standalone C++17 compile check for the public hint type, including `float id_distance` and its default, passed.
- Local `[cluster_api],[search_hint]` API tests passed: 6 test cases, 69 assertions, including `id_distance` round trips/defaults/immutable ownership, one-key metadata, 64-bit offsets, and per-query absent/empty states.
- Full CI: pending.

## Hold

Draft / HOLD. Do not merge or mark ready until the required tests pass and the author explicitly confirms review approval. No auto-merge is requested.


alexanderguzhva:
@foxspy is it still WIP?

foxspy:
> @foxspy is it still WIP?

Ready for review now—thanks for taking a look!

PR #1829, opened 2026-09-15
feat: add C API and Java JNI bindings

LawrenceTL92:
issue: #1834

## What changes

Knowhere currently has C++ and Python entry points but no standalone C ABI or Java binding. This adds opt-in C and JNI libraries plus a Java 11 API for dense index creation, build, search, serialization, deserialization, BinarySet chunk transfer and brute-force search.

The C ABI uses checked typed handles, explicit shapes and capacities, owned initialization data, and status/thread-local errors. Java accepts direct native-order buffers, exposes AutoCloseable resources, and translates native failures without allowing C++ exceptions across either boundary. DiskANN uses the existing local FileManager and a file-based build/load contract.

CMake/Conan options remain off by default. Maven produces an API JAR and a separate Linux platform JAR; its loader verifies and extracts the bundled shared-library closure. No Spark, Scala, Arrow Java, Hadoop, or artifact-publication service is required. Aquila clustering, deduplication, Isolation Forest and file conversion are deferred.

Thread pools. Search and brute force run one single-threaded task per query row on Knowhere's global search pool, which is created lazily with the hardware thread count. The C API exposes `knowhere_search_thread_pool_resize/_size` and the build-pool pair (mapped to `KnowhereConfig::Set/Get*ThreadPoolSize`; sizes 1..INT32_MAX; size accessors return 0 until the pool exists), and Java exposes `Knowhere.resizeSearchThreadPool`, `searchThreadPoolSize`, `resizeBuildThreadPool` and `buildThreadPoolSize`. Hosts that schedule their own parallelism size the pools before the first search.

## Validation

- macOS arm64: C ABI/concurrency tests passed (2/2), AddressSanitizer C tests passed (2/2), Java 11 JNI tests passed (10/10, `-Xcheck:jni`, JRE signal chaining enabled).
- Linux aarch64 (Ubuntu 24.04, GCC 13): C tests passed (3/3 including DiskANN); JDK 17 JNI tests passed (11/11 including DiskANN). ELF packager and Conan license-graph tests passed (13/13).
- Linux aarch64 platform JAR: real FLAT build/serialize/deserialize/search and brute-force smoke passed on clean Temurin 11.0.32, 17.0.20 and 21.0.12 JREs with networking disabled and no build or Conan paths. Final runs have no HotSpot JNI/signal-handler diagnostics.
- Linux x86-64 (Ubuntu 22.04, GCC 12.3, Xeon Gold 6338): native build passed with 64-way compilation; C tests passed 3/3, JDK 17 JNI/DiskANN tests passed 11/11, and packager tests passed 13/13. The final test fixture handles dpkg ownership on merged-/usr systems; its changed case also passed on Linux aarch64.
- Linux x86-64 platform JAR: real index round-trip and brute-force smoke passed on clean Temurin 11.0.32, 17.0.20 and 21.0.12 JREs, without networking or build/cache mounts. Final logs have no HotSpot JNI/signal-handler diagnostics. All 1,700 tracked server source files were checked against final commit `9dc2b8ad537502d408bc33af05727453295d6622`.

- Linux x86-64 (Ubuntu 22.04, GCC 12.3, 128 threads) with a Temurin 11.0.32 toolchain, after switching CI to JDK 11: C tests passed 3/3, JNI tests passed 11/11 under `-Xcheck:jni`, packager tests passed 13/13, and the API JAR is class-file major version 55 (Java 11). The packaged JARs passed the clean-JRE smoke on Temurin 11.0.32, 17.0.20, 21.0.12 and 25.0.4 with networking disabled and no JNI or signal-handler diagnostics.
- Linux x86-64 (Ubuntu 22.04, GCC 12.3, JDK 11) at 1fff20db: thread pool sizing covered by the C tests (3/3) and the JNI tests (12/12 with `-Xcheck:jni`), with no JNI diagnostics; macOS arm64 C 2/2 and JNI 11/11.

## CI corrections and verification

The workflow now explicitly uses Bash with pipefail, so Maven failures cannot be hidden by tee. The DiskANN binding tests use uniformly distributed samples and require at least 95% recall against brute force, while checking every returned ID, scalar L2 distance, ordering and exclusion. This follows the approximate-search contract; the previous isolated-point fixture could produce graph nodes with no incoming edges. The retained failing graph returned identical results through JNI and the C ABI.

After these changes, Linux x86-64 C tests passed 3/3 and JDK 17 JNI tests passed 11/11, with no HotSpot JNI misuse or overwritten-handler diagnostics. Twenty independently rebuilt DiskANN indexes passed; filtered and unfiltered recall was 100/100 on every run. Native library and Java API implementations are unchanged. The earlier macOS/aarch64 and packaged-JRE records above describe the pre-fix test revision.

## Runtime contract

Shared Folly installs a process-wide SIGPIPE handler during library loading. HotSpot consumers must preload the running JRE's `libjsig` before JVM startup to preserve its handlers through supported signal chaining. The README gives Linux and macOS commands; CI also rejects JNI diagnostics that may not fail Java assertions. The verified packages reference up to GLIBC_2.35 on x86-64 and GLIBC_2.38 on aarch64; these runs do not establish compatibility with older runtime images.

## Java version support

The API JAR is Java 11 bytecode (`javac --release 11`) and is consumed unchanged by later runtimes. CI builds and runs the JUnit suite on JDK 11, the minimum supported version, and exercises the packaged JARs on every currently supported LTS runtime: Temurin 11, 17, 21 and 25. On JDK 24 and later, loading a JNI library is a restricted operation ([JEP 472](https://openjdk.org/jeps/472)); consumers start the JVM with `--enable-native-access=ALL-UNNAMED` (or name the module that contains `io.knowhere`). Without it the JVM prints a warning at load time today and will refuse the call in a future release. The README documents this requirement.

## Known distribution limitation

The pinned milvus-common source omits LICENSE/NOTICE material referenced by its recipe and headers, including a separately licensed flat_hash_map. Packaging retains a missing-license report instead of inventing those documents. The JNI workflow validates runtime behavior and does not publish artifacts. The added publication-license audit step and notification have been removed from CI; publication requirements are outside this binding-validation PR. The PR remains draft for review.


.github/workflows/jni.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+jobs:
+  linux-x86_64:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 240
+    env:
+      CC: gcc-12
+      CXX: g++-12
+      FC: gfortran-12
+      OMP_NUM_THREADS: "2"
```
alexanderguzhva: I don't see these parameters in other workflows, so please omit `OMP_NUM_THREADS` and `OPENBLAS_NUM_THREADS` to allow the machine to use all available threads, unless there is a reason for not doing so

.github/workflows/jni.yaml
```diff
@@ -0,0 +1,141 @@
...
+      OPENBLAS_NUM_THREADS: "2"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "17"
```
alexanderguzhva: I'm confused. The description of the commit mentions Java 11, here I see Java 17, the most recent Java version is 27 (LTS is Java 25). What's the sufficient one and whether there are any differences at all, given a possible backward compatibility?

.github/workflows/jni.yaml
```diff
@@ -0,0 +1,141 @@
...
+          python java/scripts/bundle_native.py --library "$PWD/build/Release/java/libknowhere_jni.so" \
+            --output "$PWD/java/target/native-resources" --platform linux-x86_64
+          mvn -B -f java/pom.xml package -Dnative.platform=linux-x86_64 -DskipTests
+      - name: Test packaged libraries on clean JREs
+        run: |
+          smoke=$(mktemp -d)
+          mkdir -p "$smoke/io/knowhere"
+          cp java/target/knowhere-jni-*.jar "$smoke/"
+          cp java/target/test-classes/io/knowhere/PackageSmoke.class "$smoke/io/knowhere/"
+          for version in 11 17 21; do
```
alexanderguzhva: three versions?

src/c_api/c_api.cc
```diff
@@ -0,0 +1,587 @@
...
+#include "knowhere/version.h"
+
+namespace {
+thread_local char last_error[4096] = {};
+
+class Failure : public std::runtime_error {
+ public:
+    Failure(int32_t code, const std::string& message) : std::runtime_error(message), code(code) {
+    }
+    int32_t code;
```
alexanderguzhva: please initialize fields with default values to avoid compiler warnings
```C++
int32_t code = 0;
```

This applies to all following cases of this kind.

src/c_api/c_api.cc
```diff
@@ -0,0 +1,587 @@
...
+        if (it != registry.end()) {
+            released = std::move(it->second);
+            registry.erase(it);
+        }
+    }
+    // Destruction is outside the registry lock. An in-flight call owns another reference.
+}
+
+uint64_t
+Multiply(uint64_t left, uint64_t right) {
```
alexanderguzhva: Please explain why this complicated function is needed.
Basically, why is it needed to check the overflow of multiplying two u64 values and throw if out of bounds?

conanfile.py
```diff
@@ -31,6 +36,9 @@ class KnowhereConan(ConanFile):
         "cardinal_version_force_checkout": [True, False],
         "with_profiler": [True, False],
         "with_ut": [True, False],
+        "with_c_api": [True, False],
+        "with_jni": [True, False],
+        "with_c_api_tests": [True, False],
```
alexanderguzhva, line 41: please also update the `Makefile` correspondingly, if suitable

java/README.md
```diff
@@ -0,0 +1,192 @@
...
+Sparse layouts and the Aquila clustering, deduplication and Isolation Forest
+operations are outside this interface.
+
+## Build and test
+
+Use the repository's normal C++20/Conan 2 prerequisites, plus a JDK 11 or later
+and Maven. On Linux the platform packaging step also requires Python 3,
+`readelf`, `ldd` and `patchelf`. Configure `JAVA_HOME` when CMake cannot find the
+selected JDK. The example uses Conan 2.28.1 and CMake 3.28.1. Older dependency
+recipes may fail with CMake 4.
```
alexanderguzhva: `may fail with CMake 4` ?
CMake 4+ is a default on recent builds (say, ubuntu 26.04, if I'm not wrong), please test it and ensure that it works

alexanderguzhva:
@LawrenceTL92 please also check the `C and Java bindings / linux-x86_64` error. I can help with this, if needed

.github/workflows/jni.yaml
LawrenceTL92: Thanks for catching this, Alexander. You're right, there was no real reason behind the cap; it was just a conservative default I carried over from early local runs. I've removed both variables in a04627a so the job now uses all available threads, and I also switched the build step to `--parallel "$(nproc)"` to match how the other workflows behave. None of the binding tests depend on a fixed thread count, so this should only make the run faster.

.github/workflows/jni.yaml
LawrenceTL92: Thanks, Alexander, that's a fair point and the mismatch was on our side. The actual requirement is Java 11: the Java API is compiled with `--release 11` (class-file major version 55), which is the hard contract for the Spark 3.5 consumers, and it runs unchanged on newer runtimes. JDK 17 here was only the build toolchain and had no functional purpose, so it was just confusing.

In 906b84b I switched the workflow to build and run the JUnit suite on JDK 11, the minimum we support, and the packaged JARs are now smoke-tested on every currently supported LTS: Temurin 11, 17, 21 and 25 (the loop below).

Regarding backward compatibility on newer JDKs: I ran the packaged JARs on Temurin 25.0.4 and everything passes, but there is one real difference worth knowing. Starting with JDK 24, loading a JNI library is a restricted operation ([JEP 472](https://openjdk.org/jeps/472)); without `--enable-native-access=ALL-UNNAMED` the JVM prints a warning at load time and a future release will refuse the call. I documented this in a new "Supported Java versions" section of java/README.md and in the PR description, and the CI smoke passes the flag for the JDK 25 run. Java 27 is not an LTS, so I left it out of the matrix, but I'm happy to add it if you think it's valuable.

Verified on Linux x86-64 with the JDK 11 toolchain: C tests 3/3, JNI tests 11/11 with `-Xcheck:jni`, packager tests 13/13, and clean Temurin 11/17/21/25 JRE smoke with no JNI or signal-handler diagnostics.

.github/workflows/jni.yaml
LawrenceTL92: Good question, sorry it wasn't explained. These are the LTS runtimes we support: the JARs are Java 11 bytecode, and this step loads the packaged native libraries on each clean LTS JRE to catch runtime differences that the unit tests on a single JDK wouldn't. As of 906b84b the loop covers 11, 17, 21 and 25 (25 being the current LTS), and the JDK 25 run passes `--enable-native-access=ALL-UNNAMED` because of JEP 472. More detail is in my reply on the JDK version comment above. I added a short comment in the workflow so the intent is visible next time.

src/c_api/c_api.cc
LawrenceTL92: Done in 715db496, thank you. I went through both native files and gave every member an explicit default: `Failure::code`, `IndexResource::dtype` and `file_build` in c_api.cc, and the `Release` and `Region` helper members in knowhere_jni.cc. All of them are still assigned by their constructors, so this is purely declarative and behavior is unchanged. Rebuilt on Linux x86-64 with GCC 12: no new warnings, C tests 3/3 and JNI tests 11/11 still pass.

src/c_api/c_api.cc
LawrenceTL92: Thanks for pushing on this, Alexander, the code did not explain itself and the naming made it worse. Here is the reason it exists.

`rows`, `dimensions`, `top_k` and the buffer capacities all arrive from the caller (from Java they are a `long` and an `int`) and cross the ABI unchecked. The library computes `rows * dimensions * element_bytes` and then verifies `bytes >= needed` in `CheckBuffer`. If that product wraps around, `needed` becomes small, the capacity check passes, and the copy that follows runs past the caller's DirectByteBuffer. That is memory corruption inside the JVM rather than an exception, so the overflow check is what makes the capacity check meaningful at all. The header already promises this ("all dimension/size multiplication is checked for overflow" in c_api.h), and the existing tests rely on it: the Java test that builds with `Long.MAX_VALUE` rows must throw, and the C tests that call allocate/write with `UINT64_MAX` must return `INVALID_ARGUMENT`. The second condition (`<= INT64_MAX` and `<= SIZE_MAX`) is there because the product is later handed to Knowhere as an `int64_t` and used as a `size_t` allocation size.

What made it look odd was that the same function was also being called as `Multiply(x, 1)` in two places purely to get the range check. In bb6ce3c8 I split it into `CheckSize()` for the range check and `CheckedBytes()` for the product, which now uses `__builtin_mul_overflow` instead of the hand-written division, and I added a comment describing the failure it prevents. The two `Multiply(x, 1)` sites now call `CheckSize()` directly. I also added two C test cases that drive `rows * row_bytes` and `count * sizeof(int64_t)` past uint64 and assert they are rejected with the overflow error.

Behavior is unchanged. Rebuilt on Linux x86-64 with GCC 12: C tests 3/3 and JNI tests 11/11 pass.

conanfile.py
LawrenceTL92: Good call, thank you. In 9e3f476b the Makefile now accepts `WITH_C_API=True`, `WITH_JNI=True` and `WITH_C_API_TESTS=True`, has a `test-c-api` target that runs the C tests through ctest, and lists all of this in `make help`. Both READMEs mention the shortcut.

Doing this also caught a real bug, so I'm glad you asked. Running `make WITH_JNI=True` (without `WITH_C_API=True`) failed immediately: Conan 2 refuses the `self.options.with_c_api = True` assignment in `configure()` with "Incorrect attempt to modify option". CI had always passed both options together, so this path was never exercised. I removed the mutation and the recipe now derives the C API state from `with_c_api or with_jni` wherever it is used, which is exactly what CMakeLists.txt already does.

Verified on Linux x86-64: `make WITH_JNI=True WITH_C_API_TESTS=True` completes from a fresh build directory and produces libknowhere_c, libknowhere_jni and the three C test binaries; `make test-c-api` passes 3/3.

java/README.md
LawrenceTL92: You're right that this needed a real test rather than a hedge, thank you. I ran the README build path on Linux x86-64 with CMake 4.4.3 (Conan 2.28.1, GCC 12) from an empty Conan cache so every dependency was actually built under CMake 4.

Result: without any extra setting, `conan install` fails at `double-conversion/3.3.0`, whose CMakeLists declares a `cmake_minimum_required` below 3.5, with CMake's "Compatibility with CMake < 3.5 has been removed" error. With `CMAKE_POLICY_VERSION_MINIMUM=3.5` exported, all 30 dependency packages build, Knowhere itself configures under CMake 4.4.3 with no warnings or deprecation notices, and the C tests (3/3), JNI tests (11/11, JDK 11, `-Xcheck:jni`) and diagnostics check all pass. A few recipes tool-require `cmake/[<4]`, and Conan supplies 3.31 for those builds on its own, so users don't need to do anything there.

This is the same variable the Makefile already exports and the macOS unit-test job already sets, so the code needed no change. In 8c6017be I replaced the vague sentence in java/README.md with the concrete statement (works with 3.28 and 4, export the variable before `conan install` when CMake 4 is on the PATH) and added the export line to the example. The `make WITH_JNI=True` shortcut from the other thread sets it automatically.

CI currently pins CMake 3.28.1 via pip. If you'd like, I can add a CMake 4 leg to the bindings workflow so this stays covered; I left it out for now to keep the job time down.

LawrenceTL92:
@alexanderguzhva thanks a lot for the thorough review. The `C and Java bindings / linux-x86_64` failure was the distribution-license audit step tripping on the pinned milvus-common source, which lacks the LICENSE/NOTICE files its recipe references. That step was never part of the runtime validation, so I removed it from this workflow; publication readiness is out of scope for this PR. The workflow has since passed on GitHub Actions (run for 715db49), and the latest push is running now.

I've also addressed each of your inline comments with a reply and a commit: thread caps removed (a04627a), JDK 11 build with a Java 11/17/21/25 smoke matrix and the JEP 472 note (906b84b), default member initializers (715db49), the size-overflow helper split and explained with new tests (bb6ce3c), Makefile flags plus a Conan option fix they uncovered (9e3f476), and the verified CMake 4 note (8c6017b). Everything was rebuilt and tested on Linux x86-64 before pushing. Happy to iterate further on any of them.

LawrenceTL92:
@alexanderguzhva I pushed one more commit, 1fff20db, adding thread pool sizing to the C and Java APIs: `knowhere_search_thread_pool_resize` / `_size` and the build-pool pair, mapped to `KnowhereConfig::Set/Get*ThreadPoolSize`, plus `Knowhere.resizeSearchThreadPool` and friends.

Search and brute force schedule one single-threaded task per query on the global search pool, which defaults to the hardware thread count; a host such as a Spark executor that already runs one search per task had no way to bound that pool. The C layer validates 1..INT32_MAX because the core only logs and ignores 0, and the size accessors return 0 until a pool exists.

Covered by a new C test and a JUnit test; verified on Linux x86-64 (C 3/3, JNI 12/12 with -Xcheck:jni) and macOS arm64. It belongs in this PR because the bindings are meant to ship as the complete C API and JNI surface, including the runtime knobs an embedding host needs.

alexanderguzhva:
@LawrenceTL92 all right, looks good. Please let me know when the PR is ready, bcz currently it is marked as a draft.

PR #1835, opened 2026-09-17
enhance: report a cancelled search as cancelled, not as an engine error

### LINKED ISSUE #1831: Search cancellation is reported as an engine inner error instead of a dedicated status
### Problem

`knowhere::checkCancellation` (`include/knowhere/context.h`, added in #1399) throws `folly::FutureCancellation` when the caller's `milvus::OpContext` requests cancellation. Every search path then catches it with the generic `std::exception` handler and reports it as an engine failure:

| path | what the caller gets |
|---|---|
| `faiss_hnsw.cc`, `ivf.cc`, `flat.cc`, `svs_*.cc`, `index_node.cc` (RangeSearch / CalcDistByRawIndex) | `Status::faiss_inner_error`, logged as `faiss inner error` at WARNING |
| `diskann.cc` / `diskann_aisaq.cc` via `TryDiskANNCall` | `Status::diskann_inner_error`, logged at ERROR |
| nodes without their own catch (`faiss.cc`, `data_view_dense_index.h`) | `Status::knowhere_inner_error` from `GuardedCall` |
| Cardinal | `Status::cardinal_inner_error` |

The existing unit test even locks this in: `tests/ut/test_search.cc` ("pre-cancelled context should return error") asserts `faiss_inner_error || cardinal_inner_error`.

### Why it matters

A cancellation the caller asked for is indistinguishable from a genuine engine failure at every level:

- same `Status` value and the same `Status2String` text;
- same `StatusCategory` (`permanent_error`), so `IsInnerError()` is true for a request that did nothing wrong;
- same segcore code from `ToSegcoreErrorCode` (`milvus::ErrorCode::KnowhereError`), while the shared registry already has a dedicated `FollyCancel` code for exactly this case;
- WARNING / ERROR log lines for expected control flow.

`Status::timeout` does not fit: it is only produced by the Cardinal `BuildAsync` path, and its own comment in `expected.h` / `segcore_error_code.h` says it conflates cancel with timeout.

### Proposal

1. Add `Status::cancelled` (+ `Status2String`).
2. Add a `StatusCategory::cancelled` and classify the new status there. It is neither the caller's fault, nor a permanent or transient engine error, so none of the existing three categories is right.
3. `ToSegcoreErrorCode(Status::cancelled)` → `milvus::ErrorCode::FollyCancel`.
4. In each search path, catch `folly::FutureCancellation` before the generic `std::exception` handler and return `Status::cancelled`; add the same mapping to `GuardedCall` as a backstop for nodes without their own handler.
5. Update the cancellation tests to assert `Status::cancelled`.

No change to when cancellation is checked or how `OpContext` is threaded through; only the reported status changes.

czs007:
Closes #1831

### What this changes

A search that is cancelled by its caller currently comes back as an engine failure. `folly::FutureCancellation` propagates out of the index and lands in the generic `catch (const std::exception&)` of whichever index was running, so the caller sees `faiss inner error: Future was cancelled`, `hnsw_inner_error`, `diskann_inner_error` and so on depending only on which index happened to be searching. Everything above knowhere then has to treat a deliberate cancellation as a server-side error: it is logged as a failure, counted as a failure, and is a candidate for retry.

This adds a first-class status for it.

- `Status::cancelled = 34`, with its own `StatusCategory::cancelled`. It is deliberately neither `client_error` nor `inner_error`: a cancellation is not the caller's fault and not an engine fault, so code that classifies by category does not have to special-case it.
- `IsCancelled()`, `Status2String` and `StatusCategoryOf` updated. `StatusCategoryOf` has no `default:` label, so `-Werror=switch` makes any future status that forgets to pick a category a build error rather than a silent misclassification.
- `ToSegcoreErrorCode` maps it to `FollyCancel`, the code segcore and the Go layer already use for this.
- Every search path catches `folly::FutureCancellation` before its generic handler: flat, ivf, faiss hnsw, the three svs indices, `index_node.cc`, and both DiskANN paths.
- The two DiskANN paths previously discarded the status their `TryDiskANNCall` helper returned and substituted a fixed `diskann_inner_error` / `aisaq_error`. They now keep a `cancelled` status.
- The brute-force search paths in `brute_force.cc` read the cancellation token inside their task-pool bodies. Before this they accepted an `OpContext` and never looked at it, so a cancelled brute-force search ran to completion.
- `GuardedCall` recognises a cancellation that no index handled, so a path added later cannot silently regress to `inner_error`.

Recognising folly's exception lives in a separate translation unit (`src/common/cancellation.cc`) so `expected.h` does not pull in folly.

### Tests

`test_error_code.cc` covers the new status, its category and its segcore mapping. `test_search.cc` and `test_diskann.cc` cover cancelling a live search on each path and assert the status is `cancelled` rather than the index's own error.

Full suite on this branch: 241 cases, 240 passed, 1 skipped, 137,507,899 assertions.


foxspy:
**[P2] Preserve cancellation through the multi-vector search wrappers.**

At `302f2fe`, the underlying `Search` / `CalcDistByStorageIds` paths can return `Status::cancelled`, but the multi-vector strategies still overwrite it:

- ANN stage: [TokenANN](https://github.com/zilliztech/knowhere/blob/302f2fe67426080c67914e913e07793cf5f6811c/src/index/emb_list/emb_list_strategy_token_ann.cc#L94-L97), [MUVERA](https://github.com/zilliztech/knowhere/blob/302f2fe67426080c67914e913e07793cf5f6811c/src/index/emb_list/emb_list_strategy_muvera.cc#L206-L209), and [LEMUR](https://github.com/zilliztech/knowhere/blob/302f2fe67426080c67914e913e07793cf5f6811c/src/index/emb_list/emb_list_strategy_lemur.cc#L364-L367) convert every failed result to `emb_list_inner_error` and log it at ERROR.
- Rerank stage: the [shared helper](https://github.com/zilliztech/knowhere/blob/302f2fe67426080c67914e913e07793cf5f6811c/src/index/emb_list/emb_list_strategy.cc#L66-L70) overwrites the status returned by `CalcDistByStorageIds`; each strategy also converts a failed rerank status to `emb_list_inner_error`.

Consequently, searching a successfully built multi-vector index with a pre-cancelled `OpContext` still returns an internal error through the public `Index::Search()` API when the underlying ANN search detects cancellation. `GuardedCall` cannot recover the cancellation once it has been converted into an `expected` error value. This is an existing wrapper behavior that leaves this fix incomplete.

Please propagate `cancelled` and its error message through the ANN wrappers, shared rerank helper, and strategy-level rerank wrappers, without logging cancellation at ERROR. Please also add regression coverage for:

1. TokenANN / MUVERA / LEMUR searches with a pre-cancelled context, asserting that the public Search API returns `Status::cancelled`.
2. Cancellation observed at a rerank checkpoint after ANN succeeds, asserting that the final status remains `cancelled` rather than being rewritten to `emb_list_inner_error`.

czs007:
Thanks, you're right. It's the same shape as the DiskANN wrapper this PR already fixes, and I missed these. Addressed in 7bf0e7d.

Cancellation is now passed through unchanged, and no longer logged at ERROR, at all seven places it was being rewritten:

- the ANN stage of TokenANN, MUVERA and LEMUR;
- the shared `RerankByCalcDistByStorageIds` helper;
- each strategy's handling of the status that helper returns.

The other `emb_list_inner_error` returns are genuine inner errors (uninitialised strategy, invalid id, size limits) and are unchanged.

Regression coverage, in `test_emb_list.cc`. Both cases fail on the previous code with `emb_list_inner_error`:

1. A search cancelled before it starts, for each of TokenANN, MUVERA and LEMUR, through the public `Index::Search`.
2. Cancellation first observed at the rerank checkpoint after ANN succeeds. A shared token can't make the ANN stage succeed while the rerank sees the cancellation, so the test drives a TokenANN strategy over a real HNSW index wrapped to report cancellation from `CalcDistByStorageIds` only, and asserts that the ANN stage did run first.

One limit on the second case: it exercises TokenANN. MUVERA and LEMUR reach the rerank through the same helper and the same pass-through, but they need a trained strategy to drive directly, so I haven't added the equivalent for them. Happy to if you'd like it.

Full suite: 242 cases, 241 passed, 1 skipped (the same skip as before).

foxspy (approved):
/lgtm

PR #1836, opened 2026-09-17
fix: select Faiss backend in HNSW fallback regression test

marcelo-cjl:
The regression test added in #1824 creates `HNSW` but then reads and edits its serialized data as a Faiss index. In Cardinal builds at index version 11, `HNSW` resolves to Cardinal and the test fails with `Index type 0x05e639ae not recognized` before reaching the fallback assertions.

Select `HNSW_DEPRECATED` when Cardinal is enabled, keeping the serialization key as `HNSW`. Preserve all existing fallback assertions in `test_faiss_hnsw.cc`.

Observed in [Cardinal #949 CI](https://github.com/zilliztech/cardinal/actions/runs/35113823282/job/104854050855).

Validation (index version 11, matching Cardinal CI):
- Before: reproduced the exact CI failure with Cardinal + ASan; the original test passed without Cardinal.
- After: `[faiss_hnsw]` passed in both configurations (2 tests, 51 assertions each), including all three fallback sections.
- Pre-commit checks passed.

The CI-only `current_version=11` setup is not part of this change.


foxspy (approved):
/lgtm

alexanderguzhva:
sry

PR #1837, opened 2026-09-17
fix: preserve explicit ID boundaries in zero-filter bitset views

marcelo-cjl:
A growing-index search can prepare a zero-filter bitset before an append and then scan the expanded index without a selector. Add `BitsetView::set_require_id_boundary()` so callers can retain ID-boundary checks even when the exact filter count is zero. Existing callers retain the current zero-filter optimization.

issue: milvus-io/milvus#51941

This is the Knowhere-side API change. Milvus still needs to enable the flag for growing searches; this PR does not complete that integration or establish the historical incident root cause.

Validation: Release + ASan build passed. The deterministic `PrepareBitset -> Add -> Search` regression passed all eight combinations of IVF_FLAT_CC/SCANN_DVR, nullable/non-nullable and L2/COSINE (196 assertions). The unbounded controls return the appended vector; bounded searches return only IDs in the original prefix. Related bitset, IdMap, IVF, SCANN_DVR and data-view regressions passed, as did scoped pre-commit checks. No Milvus integration test was run in this phase.


foxspy (approved):
/lgtm

PR #1838, opened 2026-09-18
restore sparse quantization from serialized indexes

chasingegg:
Sparse indexes built with non-default posting types, such as IP `fp32`, can return incorrect results after reload when deserialization uses the default type. The Milvus regression returned ID `462` instead of `0` as its first result. https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pool-dispatcher/7583/artifact/failure-logs/pool-7583-distpuls-w4/pytest.log

Persist the actual posting type in the reserved index header for flat, block, and SINDI encodings, including the concrete result of BM25 `auto`. 

Before this, we actually reply on the config passed in deserialization stage and it should be the same with build param, the better way is to infer the quant type from index itself. In precious pr #1819 removes quant_type from deserialization param, but only persist sindi quant type, we should extend to all quant types. 

Both ordinary and mmap deserialization restore this type before creating the index. Files without type metadata fall back to the supplied `quant_type` and existing defaults; unknown type markers and metric mismatches are rejected.

Related to #1730

/kind improvement


alexanderguzhva:
@chasingegg to which degree this affects backward compatibility? Was the backward compatibility validated (maybe, in a separate experiment) ?

chasingegg:
> @chasingegg to which degree this affects backward compatibility? Was the backward compatibility validated (maybe, in a separate experiment) ?

compatibility is guaranteed, we use reserved header bytes to save quant type, and default value 0 is unspecified quant type which is legacy file format.

